PR #21, opened 2014-03-14
[NO_MERGE] [JENKINS-22193] - Added an option to disable the Jar caching

oleg-nenashev:
The change adds a "-jar-cache-disabled", which disables the jar caching
It may be helpful to workaround issues like https://issues.jenkins-ci.org/browse/JENKINS-20913


oleg-nenashev:
It's no longer useful, because the issue has been fixed via checksum generation and JVM restart

PR #27, opened 2014-10-18
[JENKINS-6167] Support JNLP slave connection via HTTP proxy

shirosaki:
We get proxy setting from http_proxy environment variable.
https://issues.jenkins-ci.org/browse/JENKINS-6167


tangicolin:
What appends if we want to have http_proxy set to be able to do some step action using the proxy but we want to have a direct connexion between master and slave. 
If a correctly understand this patch, the value of no_proxy is currently not take.

PR #28, opened 2015-01-24
[JENKINS-26580] Refactor slave JNLP engine to make it easier to add more...

akshayabd:
... protocols in the future. The engine will now call a factory that returns a list of protocols to try in order. Developers can implement new protocols in separate classes and add them to the factory to be used. Added tests for existing protocols. I noticed the Engine class doesn't have a corresponding Test class. It would be nice to add one, but I think it needs more refactoring before it would be realistic to add tests for it.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -24,8 +24,8 @@
 package hudson.remoting;
 
 import hudson.remoting.Channel.Mode;
-import hudson.remoting.engine.JnlpProtocol;
-import hudson.remoting.engine.JnlpProtocolFactory;
+import org.jenkinsci.remoting.engine.JnlpProtocol;
+import org.jenkinsci.remoting.engine.JnlpProtocolFactory;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
@@ -179,18 +179,16 @@ public class Engine extends Thread {
 
                     // find out the TCP port
                     HttpURLConnection con = (HttpURLConnection)salURL.openConnection();
-                    if (con != null) {
-                    	if (credentials != null) {
-                    		// TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
-                    		String encoding = Base64.encode(credentials.getBytes("UTF-8"));
-                    		con.setRequestProperty("Authorization", "Basic " + encoding);
-                    	}
+                  	if (credentials != null) {
+                   		// TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
+                   		String encoding = Base64.encode(credentials.getBytes("UTF-8"));
+                   		con.setRequestProperty("Authorization", "Basic " + encoding);
+                   	}
 
-                    	if (proxyCredentials != null) {
-    	                    String encoding = Base64.encode(proxyCredentials.getBytes("UTF-8"));
-    	                    con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
-                    	}
-                    }
+                   	if (proxyCredentials != null) {
+   	                    String encoding = Base64.encode(proxyCredentials.getBytes("UTF-8"));
+   	                    con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
+                   	}
                     try {
                         try {
                             con.setConnectTimeout(30000);
@@ -233,29 +231,29 @@ public class Engine extends Thread {
                 Socket jnlpSocket = connect(port);
                 DataOutputStream outputStream = new DataOutputStream(jnlpSocket.getOutputStream());
                 BufferedInputStream inputStream = new BufferedInputStream(jnlpSocket.getInputStream());
-                String response = "";
+                boolean connected = false;
 
                 // Try available protocols.
                 for (JnlpProtocol protocol : protocols) {
                     events.status("Trying protocol: " + protocol.getName());
-                    response = protocol.performHandshake(outputStream, inputStream);
+                    String response = protocol.performHandshake(outputStream, inputStream);
 
                     // On success do not try other protocols.
                     if (response.equals(GREETING_SUCCESS)) {
+                        connected = true;
                         break;
                     }
 
                     // On failure log the response and form a new connection.
                     events.status("Server didn't understand the protocol: " + response);
                     jnlpSocket.close();
-                    Thread.sleep(500);
                     jnlpSocket = connect(port);
                     outputStream = new DataOutputStream(jnlpSocket.getOutputStream());
                     inputStream = new BufferedInputStream(jnlpSocket.getInputStream());
                 }
 
                 // If no protocol worked.
-                if (!response.equals(GREETING_SUCCESS)) {
+                if (!connected) {
                     onConnectionRejected("None of the protocols were accepted");
                     continue;
                 }
@@ -374,5 +372,9 @@ public class Engine extends Thread {
 
     private static final Logger LOGGER = Logger.getLogger(Engine.class.getName());
 
+    /**
+     * @deprecated Use {@link JnlpProtocol#GREETING_SUCCESS}.
+     */
+    @Deprecated
     public static final String GREETING_SUCCESS = JnlpProtocol.GREETING_SUCCESS;
 }
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -179,16 +179,16 @@ public class Engine extends Thread {
 
                     // find out the TCP port
                     HttpURLConnection con = (HttpURLConnection)salURL.openConnection();
-                  	if (credentials != null) {
-                   		// TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
-                   		String encoding = Base64.encode(credentials.getBytes("UTF-8"));
-                   		con.setRequestProperty("Authorization", "Basic " + encoding);
-                   	}
+                    if (credentials != null) {
+                        // TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
+                        String encoding = Base64.encode(credentials.getBytes("UTF-8"));
+                        con.setRequestProperty("Authorization", "Basic " + encoding);
+                    }
 
-                   	if (proxyCredentials != null) {
-   	                    String encoding = Base64.encode(proxyCredentials.getBytes("UTF-8"));
-   	                    con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
-                   	}
+                    if (proxyCredentials != null) {
+   	                String encoding = Base64.encode(proxyCredentials.getBytes("UTF-8"));
+   	                con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
+                    }
                     try {
                         try {
                             con.setConnectTimeout(30000);
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -24,23 +24,22 @@
 package hudson.remoting;
 
 import hudson.remoting.Channel.Mode;
+import hudson.remoting.engine.JnlpProtocol;
+import hudson.remoting.engine.JnlpProtocolFactory;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.ByteArrayOutputStream;
 import java.net.HttpURLConnection;
 import java.net.Socket;
 import java.net.URL;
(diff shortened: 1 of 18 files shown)


src/main/java/hudson/remoting/engine/JnlpProtocol2.java
```diff
@@ -0,0 +1,95 @@
...
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.Properties;
+
+/**
+ * Implementation of the JNLP2-connect protocol.
+ *
+ * @author Akshay Dayal
+ */
+class JnlpProtocol2 extends JnlpProtocol {
```
oleg-nenashev: It would be useful to explicitly document differences

src/main/java/hudson/remoting/engine/JnlpProtocolFactory.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * The protocols should be tried in the order they are given.
+     */
+    public static List<JnlpProtocol> createProtocols(String secretKey, String slaveName) {
+        List<JnlpProtocol> protocols = new ArrayList<JnlpProtocol>();
+        // New protocols should be tried first.
+        protocols.add(new JnlpProtocol2(secretKey, slaveName));
+        protocols.add(new JnlpProtocol1(secretKey, slaveName));
+
+        return protocols;
```
oleg-nenashev: IMO protocols should be managed from System properties

akshayabd: Could you clarify? Are you suggesting giving the user some ability to specify which protocol or what order to use them in?

I think support for that can be added. However this is a refactor-only pull request, the current implementation effectively tries JNLP2 then JNLP1 and doesn't give users any control, so thats what I kept. Support for anything else can be added later.

src/main/java/hudson/remoting/engine/JnlpProtocol2.java
akshayabd: Acknowledged - I will add more documentation for JNLP2.

src/main/java/hudson/remoting/engine/JnlpProtocolFactory.java
jglick: I see no need for user customization. Newer protocols supersede older ones, which are maintained solely for compatibility with old peers.

src/main/java/hudson/remoting/engine/JnlpProtocol2.java
akshayabd: I've added more documentation for the protocols.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -184,13 +179,13 @@ public void run() {
 
                     // find out the TCP port
                     HttpURLConnection con = (HttpURLConnection)salURL.openConnection();
-                    if (con instanceof HttpURLConnection) {
+                    if (con != null) {
```
jglick: Obviously the previous code was wrong (doing a `checkcast` before `instanceof`), but this is probably wrong too; AFAIK `openConnection` cannot return null.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -234,54 +229,41 @@ public void run() {
...
+
+                    // On success do not try other protocols.
+                    if (response.equals(GREETING_SUCCESS)) {
+                        break;
                     }
+
+                    // On failure log the response and form a new connection.
+                    events.status("Server didn't understand the protocol: " + response);
+                    jnlpSocket.close();
+                    Thread.sleep(500);
```
jglick: Why? This does not seem to have been in the original code.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -234,54 +229,41 @@ public void run() {
...
-                    if (greeting.equals(GREETING_SUCCESS)) {
-                        Properties responses = readResponseHeaders(in);
-                        cookie = responses.getProperty("Cookie");
-                    } else {
-                        onConnectionRejected(greeting);
-                        continue;
+                Socket jnlpSocket = connect(port);
+                DataOutputStream outputStream = new DataOutputStream(jnlpSocket.getOutputStream());
+                BufferedInputStream inputStream = new BufferedInputStream(jnlpSocket.getInputStream());
+                String response = "";
```
jglick: Possibly more readable to move this variable inside the loop, and have a `boolean` flag for whether the handshake succeeded.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -416,5 +374,5 @@ public static Engine current() {
 
     private static final Logger LOGGER = Logger.getLogger(Engine.class.getName());
 
-    public static final String GREETING_SUCCESS = "Welcome";
+    public static final String GREETING_SUCCESS = JnlpProtocol.GREETING_SUCCESS;
```
jglick: Could I think be `@Deprecated` in favor of the new constant location.

src/main/java/hudson/remoting/engine/EngineUtil.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * Engine utility methods.
+ *
+ * @author Akshay Dayal
+ */
+public class EngineUtil {
```
jglick: I think these methods could be `protected static` additions to `JnlpProtocol`, to avoid introducing a new top-level type into the API. (They could even be package-access, since code outside of Remoting is not expected to use them.)

src/main/java/hudson/remoting/engine/JnlpProtocol.java
```diff
@@ -0,0 +1,68 @@
...
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package hudson.remoting.engine;
```
jglick: New top-level types in this module should go into `org.jenkinsci.remoting` or a subpackage.

src/main/java/hudson/remoting/engine/JnlpProtocol.java
```diff
@@ -0,0 +1,68 @@
...
+ * master.
+ *
+ * @author Akshay Dayal
+ */
+public abstract class JnlpProtocol {
+
+    final String secretKey;
+    final String slaveName;
+
+    JnlpProtocol(String secretKey, String slaveName) {
```
jglick: To make their purpose clearer, this constructor and the two fields should be `protected`.

src/main/java/hudson/remoting/engine/JnlpProtocol2.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * This cookie identifies the current connection, allowing us to force the
+     * server to drop the client if we initiate a reconnection from our end
+     * (even when the server still thinks the connection is alive.)
+     */
+    private String cookie;
```
jglick: It is worth clarifying in comments that a `JnlpProtocol` may be stateful: that `JnlpProtocolFactory.createProtocols` is called once per `Engine`, not once per connection. Otherwise a casual reader of the `JnlpProtocol` signature could be forgiven for thinking that this object is a stateless singleton, with any state being only local variables inside one call to `performHandshake`.

jglick:
Can I assume that you have manually verified that the refactored `jnlpJars/remoting.jar` is in fact able to connect both to current Jenkins masters, using protocol version 2, and also to older masters (1.466-) supporting only version 1?

The actual compatibility situation might be complicated by the fix of [SECURITY-49](https://www.cloudbees.com/jenkins-security-advisory-2013-01-04) (in 1.498 and backports to 1.480.2 and some earlier Jenkins Enterprise releases). This seems to have changed not just the secret values, but the format of the handshake: from a global `secretKey` to a pair of `secret` and `nodeName`. So it is possible that connecting to version 1 masters does not work even before this refactoring—i.e., that this security fix should have introduced `JnlpSlaveAgentProtocol3`, allowing the older ones only in case Jenkins was run unsecured—in which case retaining `JnlpProtocol1` now is senseless.

src/main/java/hudson/remoting/Engine.java
akshayabd: Removed the check

src/main/java/hudson/remoting/Engine.java
akshayabd: Removed

src/main/java/hudson/remoting/Engine.java
akshayabd: Done

src/main/java/hudson/remoting/Engine.java
akshayabd: Done

src/main/java/hudson/remoting/engine/EngineUtil.java
akshayabd: Made the methods protected

src/main/java/hudson/remoting/engine/JnlpProtocol.java
akshayabd: Done

src/main/java/hudson/remoting/engine/JnlpProtocol.java
akshayabd: Done

src/main/java/hudson/remoting/engine/JnlpProtocol2.java
akshayabd: Added comments

akshayabd:
PTAL. Made updates based on feedback.

I've tested the changes with 1.580.2 and 1.599-SNAPSHOT. I'm downloading older versions now to test - the downloads are taking several hours for some reason:

http://mirrors.jenkins-ci.org/war-stable/1.466.1/
http://mirrors.jenkins-ci.org/war/1.460/

akshayabd:
Its probably clear but I'll mention it just in case - for JNLP3-connect I'll be making some more changes, ie. will probably change the JnlpProtocolFactory's createProtocols method to take at least one additional parameter, and may also change the performHandshake method signature a little. The reason is the JNLP3-connect protocol would have to either return a Channel, or a some arguments to ChannelBuilder, or a partially filled ChannelBuilder.

I decided not to include that level of change in this pull request - for the existing protocols the refactoring made was sufficient so I left it there.

daniel-beck:
> the downloads are taking several hours for some reason

The archive server is incredibly slow ([INFRA-102](https://issues.jenkins-ci.org/browse/INFRA-102)). Download the maven artifacts from repo.jenkins-ci.org instead:

http://repo.jenkins-ci.org/simple/releases/org/jenkins-ci/main/jenkins-war/

akshayabd:
Update: I verified the new slave.jar was able to connect to a Jenkins master 1.465 and ran a simple job.

akshayabd:
FYI - even 1.465 supported JNLP2-connect protocol:
https://github.com/jenkinsci/jenkins/blob/jenkins-1.465/core/src/main/java/hudson/TcpSlaveAgentListener.java#L184

Looks like support for JNLP2-connect was added in hudson 1.396, links below are for 1.395 vs 1.396
https://github.com/jenkinsci/jenkins/blob/hudson-1_395/core/src/main/java/hudson/TcpSlaveAgentListener.java#L167
https://github.com/jenkinsci/jenkins/blob/jenkins-1_396/core/src/main/java/hudson/TcpSlaveAgentListener.java#L184

Just to be safe I tested the new slave against hudson 1.392 and it worked.

akshayabd:
ping

Hey folks,

I know you are busy, I'd like to get either closure or progress on this. If Kohsuke is unavailable perhaps Jesse or someone else can review?

  - If you prefer a different approach to making it easy to add more JNLP protocols please let me know and I'll look into it, and I'll perhaps close this pull request.
  - If the changes are good please LGTM so I can merge the pull request in. There will be more pull requests that depend on the changes here.

Best,
Akshay

jglick:
:+1: from me, trying to get @kohsuke to look it over.

kohsuke:
My apologies for the delay.

Since you are going to make more changes, I haven't released a newer version.

PR #29, opened 2015-02-18
[JENKINS-21251] Slave channel timeout seemingly caused by usage of System.currentTimeMillis()

jtnord:
[JENKINS-21251](https://issues.jenkins-ci.org/browse/JENKINS-21251)

Initial calculation of duration to wait is susceptible to the clock changing (e.g. NTP).

Changed the code so that the initial calculation is performed with a single call to System.currentTimeMillis().
At this point I do not believe the overhead of using System.nanoTime() is worth the overhead.

### DIFF
--- a/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
+++ b/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
@@ -74,7 +74,7 @@ public class AtmostOneThreadExecutor extends AbstractExecutorService {
             long now = System.nanoTime();
             long end = now + unit.toNanos(timeout);
             while (isAlive() && (end - now > 0L)) {
-                q.wait((end - now) / 1000L);
+                q.wait(TimeUnit.NANOSECONDS.toMillis(end - now));
                 now = System.nanoTime();
             }
         }
--- a/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
+++ b/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
@@ -71,10 +71,12 @@ public class AtmostOneThreadExecutor extends AbstractExecutorService {
 
     public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
         synchronized (q) {
-            long start = System.currentTimeMillis();
-            long end = start+unit.toMillis(timeout);
-            while (isAlive() && System.currentTimeMillis()<end) {
-                q.wait(end-System.currentTimeMillis());
+            long now = System.currentTimeMillis();
+            long end = now + unit.toMillis(timeout);
+            while (isAlive() && now < end) {
+                q.wait(end - now);
+                // XXX this is not safe against Clock skew - but System.nanoTime() has performance implications.
+                now = System.currentTimeMillis();
             }
         }
         return isTerminated();
--- a/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
+++ b/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
@@ -71,12 +71,11 @@ public class AtmostOneThreadExecutor extends AbstractExecutorService {
 
     public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
         synchronized (q) {
-            long now = System.currentTimeMillis();
-            long end = now + unit.toMillis(timeout);
-            while (isAlive() && now < end) {
-                q.wait(end - now);
-                // XXX this is not safe against Clock skew - but System.nanoTime() has performance implications.
-                now = System.currentTimeMillis();
+            long now = System.nanoTime();
+            long end = now + unit.toNanos(timeout);
+            while (isAlive() && (end - now > 0)) {
+                q.wait((end - now) / 1000L);
+                now = System.nanoTime();
             }
         }
         return isTerminated();
--- a/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
+++ b/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
@@ -73,7 +73,7 @@ public class AtmostOneThreadExecutor extends AbstractExecutorService {
         synchronized (q) {
             long now = System.nanoTime();
             long end = now + unit.toNanos(timeout);
-            while (isAlive() && (end - now > 0)) {
+            while (isAlive() && (end - now > 0L)) {
                 q.wait((end - now) / 1000L);
                 now = System.nanoTime();
             }
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -55,6 +55,7 @@ import java.util.Vector;
 import java.util.WeakHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -1015,9 +1016,9 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     public synchronized void join(long timeout) throws InterruptedException {
         long now = System.nanoTime();
-        long end = now + (timeout * 1000L);
+        long end = now + TimeUnit.MILLISECONDS.toNanos(timeout);
         while ((end - now > 0L) && (inClosed == null || outClosed == null)) {
-            wait((end - now) / 1000L);
+            wait(TimeUnit.NANOSECONDS.toMillis(end - now));
             now = System.nanoTime();
         }
     }
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1016,7 +1016,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     public synchronized void join(long timeout) throws InterruptedException {
         long now = System.nanoTime();
         long end = now + (timeout * 1000L);
-        while ((end - now > 0) && (inClosed == null || outClosed == null)) {
+        while ((end - now > 0L) && (inClosed == null || outClosed == null)) {
             wait((end - now) / 1000L);
             now = System.nanoTime();
         }
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -57,6 +57,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -276,7 +278,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     @GuardedBy("lastHeardLock")
     private long lastHeard;
-    private Object lastHeardLock = new String("lastHeardLock");
+    private final ReadWriteLock lastHeardLock = new ReentrantReadWriteLock(true);
 
     /**
      * Single-thread executor for running pipe I/O operations.
@@ -1400,14 +1402,22 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * @see #lastHeard
      */
     public long getLastHeard() {
-        synchronized (lastHeardLock) {
+        try {
+            lastHeardLock.readLock().lock();
             return lastHeard;
+        }
+        finally {
+            lastHeardLock.readLock().unlock();
         }
     }
 
     private void updateLastHeard() {
-        synchronized (lastHeardLock) {
+        try {
+            lastHeardLock.writeLock().lock();
             lastHeard = System.currentTimeMillis();
+        }
+        finally {
+            lastHeardLock.writeLock().unlock();
         }
     }
 
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -24,6 +24,7 @@
 package hudson.remoting;
 
 import org.jenkinsci.remoting.CallableDecorator;
+
 import edu.umd.cs.findbugs.annotations.SuppressWarnings;
 import hudson.remoting.CommandTransport.CommandReceiver;
 import hudson.remoting.PipeWindow.Key;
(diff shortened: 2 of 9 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -1006,9 +1012,14 @@ public void checkRoles(RoleChecker checker) throws SecurityException {
      * @since 1.299
      */
     public synchronized void join(long timeout) throws InterruptedException {
-        long start = System.currentTimeMillis();
-        while(System.currentTimeMillis()-start<timeout && (inClosed==null || outClosed==null))
-            wait(timeout+start-System.currentTimeMillis());
+        long now = System.currentTimeMillis();
+        long end = now + timeout;
+        while (now < end && (inClosed == null || outClosed == null)) {
+            wait(now - end);
```
jtnord: end - now

stephenc: What about if the clock moves backwards and you pass a negative or `0` wait?

jtnord: same as currently happens - instead of waiting for 4 minutes you will wait longer.

This will never be a negative or zero wait due to the "now < end" condition (once I fix the bug).

What will be prevented though is that you will at least wait once (unless inClosed or OutClouse !- null) - whereas before you where never guaranteed to wait.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -270,7 +274,9 @@
      * without telling us anything, the {@link SocketOutputStream#write(int)} will
      * return right away, and the socket only really times out after 10s of minutes.
      */
-    private volatile long lastHeard;
+    @GuardedBy("lastHeardLock")
+    private long lastHeard;
+    private Object lastHeardLock = new String("lastHeardLock");
```
stephenc: I think an actual `Lock` and `Condition` would be better. In either case locking on a String is a bad idea so even if you want to lock on an object's built in syncronizer it should be `private final Object lastHeardLock = new Object()`

jtnord:
@stephenc I had intentionally not used System.nanoTime() which is the only true defense against all clock changes - mainly as this can cause a performance hit on Windows machines < Win8 or Server2012.

Given HotSpot will **not** have spurious wakes from Object.wait() (not sure about IBM, JROckit etc) I'm not sure it is worth the hit?

src/main/java/hudson/remoting/Channel.java
jtnord: How is locking on a String a bad idea?
Its not like I am locking on an interned String which is a very bad idea :) 
infact locking on a String helps when debugging as rather than seeing a random reference you have a good visual clue as to what Locks are held.

in any case will change to a Lock as should at least be less overhead.

src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
```diff
@@ -71,10 +71,12 @@ public boolean isTerminated() {
...
         synchronized (q) {
-            long start = System.currentTimeMillis();
-            long end = start+unit.toMillis(timeout);
-            while (isAlive() && System.currentTimeMillis()<end) {
-                q.wait(end-System.currentTimeMillis());
+            long now = System.currentTimeMillis();
+            long end = now + unit.toMillis(timeout);
+            while (isAlive() && now < end) {
+                q.wait(end - now);
+                // XXX this is not safe against Clock skew - but System.nanoTime() has performance implications.
```
jglick: BTW at @stephenc’s insistence we use `TODO` for all such comments.

jtnord: Hmm...
TODO should never be seen in production code IMHO - its to tag WIP as a
reminder for the contributors(s)
XXX is this is a potential area of concern so when looking for issues
around this code area pay attention to this
FIXME: it works - is hacky but was needed very quickly so needs to be
rewritten in the future - should only be present in production code for a
very short while.

I can change it - but having just one TAG is a step backwards :)  (intelliJ
supports multiple tags IIRC as does eclipse)

stephenc: Correction. I object to `XXX` I am fine with `FIXME` or `TODO` as
appropriate

On Wednesday, February 18, 2015, Jesse Glick <notifications@github.com>
wrote:

> In src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
> <https://github.com/jenkinsci/remoting/pull/29#discussion_r24929965>:
>
> > @@ -71,10 +71,12 @@ public boolean isTerminated() {
> >
> >      public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
> >          synchronized (q) {
> > -            long start = System.currentTimeMillis();
> > -            long end = start+unit.toMillis(timeout);
> > -            while (isAlive() && System.currentTimeMillis()<end) {
> > -                q.wait(end-System.currentTimeMillis());
> > +            long now = System.currentTimeMillis();
> > +            long end = now + unit.toMillis(timeout);
> > +            while (isAlive() && now < end) {
> > +                q.wait(end - now);
> > +                // XXX this is not safe against Clock skew - but System.nanoTime() has performance implications.
>
> BTW at @stephenc <https://github.com/stephenc>’s insistence we use TODO
> for all such comments.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/remoting/pull/29/files#r24929965>.
>


-- 
Sent from my phone

jtnord:
RemoteClassLoader already uses nanoTime()  (there are no other references in remoting or core) so I will update appropriately and hence no need to have a long drawn out discussion on the merits of XXX

src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
jtnord: RemoteClassLoader already uses nanoTime()  (there are no other references in remoting or core) so I will update appropriately and hence no need to have a long drawn out discussion on the merits of XXX

jtnord:
***Do not merge** - contains 2 new test failures.
  laneIndependence(hudson.remoting.SingleLaneExecutorServiceTest): expected:<[zxy]> but was:<[]>
  fifo(hudson.remoting.SingleLaneExecutorServiceTest): expected:<[abcdefghijklmnopqrstuvwxyz]> but was:<[]>

src/main/java/hudson/remoting/SingleLaneExecutorService.java
```diff
@@ -86,10 +86,11 @@ public synchronized boolean isTerminated() {
     }
 
     public synchronized boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
-        long start = System.currentTimeMillis();
-        long end = start+unit.toMillis(timeout);
-        while (!isTerminated() && System.currentTimeMillis()<end) {
-            wait(end - System.currentTimeMillis());
+        long now = System.nanoTime();
+        long end = now + unit.toNanos(timeout);
+        while (!isTerminated() && end - now < 0L) {
```
jtnord: typo end - now > 0L

rsandell:
:+1:

src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
```diff
@@ -71,10 +71,11 @@ public boolean isTerminated() {
...
     public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
         synchronized (q) {
-            long start = System.currentTimeMillis();
-            long end = start+unit.toMillis(timeout);
-            while (isAlive() && System.currentTimeMillis()<end) {
-                q.wait(end-System.currentTimeMillis());
+            long now = System.nanoTime();
+            long end = now + unit.toNanos(timeout);
+            while (isAlive() && (end - now > 0L)) {
+                q.wait((end - now) / 1000L);
```
stephenc: I think you will find that there are `1000000L` nanoseconds in a millisecond and not `1000L` as you suspect

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1006,9 +1014,12 @@ public void checkRoles(RoleChecker checker) throws SecurityException {
      * @since 1.299
      */
     public synchronized void join(long timeout) throws InterruptedException {
-        long start = System.currentTimeMillis();
-        while(System.currentTimeMillis()-start<timeout && (inClosed==null || outClosed==null))
-            wait(timeout+start-System.currentTimeMillis());
+        long now = System.nanoTime();
+        long end = now + (timeout * 1000L);
+        while ((end - now > 0L) && (inClosed == null || outClosed == null)) {
+            wait((end - now) / 1000L);
```
stephenc: units again converting nanoseconds to milliseconds... also since you have an actual Lock it would make more sense to add a Condition and use Condition.awaitNanos() to reduce the running total rather than relying on a wall clock no matter how monotonically increasing it claims to be

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1389,7 +1400,25 @@ public void dumpExportTable(PrintWriter w) throws IOException {
...
+        finally {
+            lastHeardLock.readLock().unlock();
+        }
+    }
+
+    private void updateLastHeard() {
+        try {
+            lastHeardLock.writeLock().lock();
+            // TODO - this is not safe against clock skew and is called from jenkins core (and potentially plugins)
+            lastHeard = System.currentTimeMillis();
```
stephenc: It would be good to Condition.signalAll() on this one

src/main/java/hudson/remoting/ClassicCommandTransport.java
```diff
@@ -89,7 +89,6 @@ private StreamCorruptedException diagnoseStreamCorruption(Exception e) throws St
             else
                 return (StreamCorruptedException)new StreamCorruptedException().initCause(e);
         }
-
```
stephenc: whitespace

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -81,14 +80,15 @@ public PingThread(Channel channel) {
     public void run() {
         try {
             while(true) {
-                long nextCheck = System.currentTimeMillis()+interval;
+                long nextCheck = System.nanoTime() + (interval * 1000L);
```
stephenc: 1ms=1e6ns

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -81,14 +80,15 @@ public PingThread(Channel channel) {
...
 
                 ping();
 
                 // wait until the next check
                 long diff;
-                while((diff=nextCheck-System.currentTimeMillis())>0)
-                    Thread.sleep(diff);
+                while((diff = nextCheck - System.nanoTime()) > 0) {
+                    Thread.sleep(diff / 1000L);
```
stephenc: 1ms=1e6ns

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -103,13 +103,13 @@ public void run() {
     private void ping() throws IOException, InterruptedException {
         Future<?> f = channel.callAsync(new Ping());
         long start = System.currentTimeMillis();
-        long end = start +timeout;
 
-        long remaining;
+        long end = System.nanoTime() + (timeout * 1000L);
```
stephenc: 1ms=1e6ns

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -103,13 +103,13 @@ public void run() {
...
+        long remaining = end - System.nanoTime();
+
         do {
-            remaining = end-System.currentTimeMillis();
             try {
-                f.get(Math.max(0,remaining),MILLISECONDS);
+                f.get(Math.max(10000,remaining),TimeUnit.NANOSECONDS);
```
stephenc: where has the magic constant of 10µs come from

src/main/java/hudson/remoting/Request.java
```diff
@@ -256,21 +256,22 @@ public RSP get() throws InterruptedException, ExecutionException {
...
-                    long now;
-                    while(response==null && (now=System.currentTimeMillis())<end) {
+                    long now = System.nanoTime();
+                    long end = now + unit.toNanos(timeout);
+                    while (response == null && (end - now > 0L)) {
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1,end-now));
+                        Request.this.wait(Math.max(1, (end - now) / 1000L));
```
stephenc: 1ms=10^6ns

src/main/java/hudson/remoting/SingleLaneExecutorService.java
```diff
@@ -86,10 +86,11 @@ public synchronized boolean isTerminated() {
...
 
     public synchronized boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
-        long start = System.currentTimeMillis();
-        long end = start+unit.toMillis(timeout);
-        while (!isTerminated() && System.currentTimeMillis()<end) {
-            wait(end - System.currentTimeMillis());
+        long now = System.nanoTime();
+        long end = now + unit.toNanos(timeout);
+        while (!isTerminated() && (end - now) > 0L) {
+            wait((end - now) /  1000L);
```
stephenc: 1ms=10^6ns

src/main/java/hudson/remoting/SynchronousExecutorService.java
```diff
@@ -33,12 +33,16 @@ public synchronized boolean isTerminated() {
...
 
         while (count!=0) {
-            long d = end - System.currentTimeMillis();
-            if (d<0)    return false;
-            wait(d);
+            long d = end - now;
+            if (d<0) {
+                return false;
+            }
+            wait(d / 1000L);
```
stephenc: 1ms=10^6ns

stephenc:
Lots of unit conversion issues :-1: until addressed. I'd also like to see the Condition used where possible

jtnord:
The new test failures have been addressed.
There are still test failures, and one test that sometimes doesn't finish - but these already exist - so will follow up with a new PR.

see jobs #1,2 &3 on https://jenkins.ci.cloudbees.com/job/libraries/job/remoting/

@reviewbybees

jtnord:
WIP - still not ready for review

src/main/java/hudson/remoting/Channel.java
jtnord: I don't get your point here - no one is (or should be at this point in time) waiting for the lastHeard to be updated and the RW lock is private.
All this is protecting against is one thread seeing a partial 32bit write to a 64bit number when calling getLastHeard() in a different Thread.

jtnord: (Which actually __is__ protected against by using a volatile long - Every day is a school day).

oleg-nenashev:
@jtnord 
Probably, you could decouple all time/lock management functionality to a <code>Timer</code> utility class. It would greatly simplify the code

src/main/java/hudson/remoting/Channel.java
stephenc: Then perhaps switch back to a volatile and remove the lock ;-)

jtnord:
**Note**: supersceded by Pull #33

PR #30, opened 2015-02-20
Fix bad pom and ignore eclipse file

jtnord:
Findbugs was declared twice with different configurations so removed the one that checked for the least amount of issues.

Add eclipse findbugs config file to .gitignore

@reviewbybees

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ target
 /.classpath
 /.project
 /.settings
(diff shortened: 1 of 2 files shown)


stephenc:
:+1:

jtnord:
**Note**: unit test failures from PRB are existing random failures.

jglick:
:+1:

PR #31, opened 2015-02-20
Fix unit test issue on windows.

jtnord:
do not compare a URI path with a File Path - it is doomed to fail on Windows.

The unit tests in question still have some issues on windows as they fail to clean up resources that are locked - however at least you see the asserts passing.

@reviewbybees

### DIFF
--- a/src/test/java/hudson/remoting/PrefetchingTest.java
+++ b/src/test/java/hudson/remoting/PrefetchingTest.java
@@ -2,9 +2,15 @@
 
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
+
 import junit.framework.AssertionFailedError;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.tools.ant.AntClassLoader;
+import org.hamcrest.core.StringContains;
+import org.hamcrest.core.StringEndsWith;
+import org.hamcrest.core.StringStartsWith;
+import org.junit.Assert;
 
 import java.io.File;
 import java.io.IOException;
@@ -131,9 +137,10 @@ public class PrefetchingTest extends RmiTestBase implements Serializable {
      * Validates that the resource is coming from a jar.
      */
     private void verifyResource(String v) throws IOException, InterruptedException {
-        assertTrue(v, v.startsWith("jar:file:"));
-        assertTrue(v, v.contains(dir.getPath()));
-        assertTrue(v, v.endsWith("::hello"));
+        Assert.assertThat(v, StringStartsWith.startsWith("jar:file:"));
+        // need to convert to a URI otherwise this fails on windows or if you have special chars in the path
+        Assert.assertThat(v, StringContains.containsString(dir.toURI().getPath()));
+        Assert.assertThat(v, StringEndsWith.endsWith("::hello"));
     }
 
     /**
@@ -160,7 +167,8 @@ public class PrefetchingTest extends RmiTestBase implements Serializable {
         verifyResource(lines[0]);
 
         assertTrue(lines[1].startsWith("jar:file:"));
-        assertTrue(lines[1].contains(dir.getPath()));
+        // need to convert to a URI otherwise this fails on windows or if you have special chars in the path
+        Assert.assertThat(lines[1], StringContains.containsString(dir.toURI().getPath()));
         assertTrue(lines[1].endsWith("::hello2"));
     }
 
--- a/src/test/java/hudson/remoting/PrefetchingTest.java
+++ b/src/test/java/hudson/remoting/PrefetchingTest.java
@@ -3,13 +3,8 @@
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 
-import junit.framework.AssertionFailedError;
-
 import org.apache.commons.io.FileUtils;
 import org.apache.tools.ant.AntClassLoader;
-import org.hamcrest.core.StringContains;
-import org.hamcrest.core.StringEndsWith;
-import org.hamcrest.core.StringStartsWith;
 import org.junit.Assert;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/test/java/hudson/remoting/PrefetchingTest.java
```diff
@@ -131,9 +137,10 @@ public void testGetResourceAsStream() throws Exception {
      * Validates that the resource is coming from a jar.
      */
     private void verifyResource(String v) throws IOException, InterruptedException {
-        assertTrue(v, v.startsWith("jar:file:"));
-        assertTrue(v, v.contains(dir.getPath()));
-        assertTrue(v, v.endsWith("::hello"));
+        Assert.assertThat(v, StringStartsWith.startsWith("jar:file:"));
+        // need to convert to a URI otherwise this fails on windows or if you have special chars in the path
+        Assert.assertThat(v, StringContains.containsString(dir.toURI().getPath()));
+        Assert.assertThat(v, StringEndsWith.endsWith("::hello"));
```
stephenc: Nit-pick

wouldn't

`Assert.assertThat(v, allOf(StringStartsWith.startsWith("jar:file:"),StringContains.containsString(dir.toURI().getPath()),StringEndsWith.endsWith("::hello")));`

be nicer... that way you have one assert and if failing it will report exactly how it has failed... otherwise you have to run the test three times to ensure you have fixed all the potential issues:

* oh its not got the right start... fix and run
* oh its not got the right contains... fix and run
* oh its not got the right end... fix and run

vs

* oh its not got the right start, contains and end... fix and run

src/test/java/hudson/remoting/PrefetchingTest.java
```diff
@@ -131,9 +136,9 @@ public void testGetResourceAsStream() throws Exception {
      * Validates that the resource is coming from a jar.
      */
     private void verifyResource(String v) throws IOException, InterruptedException {
-        assertTrue(v, v.startsWith("jar:file:"));
-        assertTrue(v, v.contains(dir.getPath()));
-        assertTrue(v, v.endsWith("::hello"));
+        Assert.assertThat(v, allOf(startsWith("jar:file:"), 
+                                   containsString(dir.toURI().getPath()),
```
jglick: Could probably even be `containsString(dir.toURI())`, no?

jglick:
:+1: there is a special place in hell for people who confuse `File.path` with `URI.path`/`URL.path`.

jglick:
BTW if on Linux I run `testGetResources_precache` from a `master` checkout with a space in its path I get an error

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at hudson.remoting.AbstractNioChannelRunner.stop(AbstractNioChannelRunner.java:31)
	at hudson.remoting.RmiTestBase.tearDown(RmiTestBase.java:55)
	at hudson.remoting.PrefetchingTest.tearDown(PrefetchingTest.java:64)
```

Not clear what the relationship to your patch is (seems like an original assertion error is getting swallowed somewhere and causing the test to fail later?), but merging your branch does fix the failure.

src/test/java/hudson/remoting/PrefetchingTest.java
jtnord: @jglick Probably - but on windows this could be file://///C:/  or file:/C:/  (due to UNC funnies) - I have seen different parts of the JDK use different bits at times on windows - so this is tolerant to that.

jtnord:
@jglick - that looks like it could be a different issue  - the South side of the channel has not been closed.
On Windows this test still fails in tearDown as the remote jars can be removed as they are locked (happens earlier in tearDownl).
Will attempt to look at this issue in a later PR in a later PR.

src/test/java/hudson/remoting/PrefetchingTest.java
jglick: Specifically `file.toUri()` is wrong; the only correct call (in Java 7+) is `file.toPath().toUri()`.

jglick:
> that looks like it could be a different issue

It does *look* like it, but again I see this failure only when running from a checkout with a space in the path, and only when your patch is not applied. Thus my suspicion that an `AssertionError` is getting caught, cached, and somehow resurfaced in a channel error. 

What failure do you get in `master` on Windows?

src/test/java/hudson/remoting/PrefetchingTest.java
jtnord: Is this a general comment for future or something you want me to change?

jtnord:
> What failure do you get in `master` on Windows?

I don't get any issues with testGetResource_precache on master.

what I do have is
```
testGetResources(hudson.remoting.PrefetchingTest)  Time elapsed: 3.213 sec  <<< FAILURE!
junit.framework.AssertionFailedError: jar:file:/C:/Users/jnord/AppData/Local/Temp/remoting634181805214733485cache/7F/CE79BB8C55572F4D2C34ABECD36F76.jar!/test/hello.txt::hello
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at hudson.remoting.PrefetchingTest.verifyResource(PrefetchingTest.java:135)
...
```
and the same from `testGetResource`.

What I do have is also different if I run `mvn -Dtest=PrefetchingTest test` vs `mvn test`  (there is one more Error in the one where only the single test is run.

After this fix the issue changes to 
```testJarLoadingTest(hudson.remoting.PrefetchingTest)  Time elapsed: 3.181 sec  <<< ERROR!
java.io.IOException: Unable to delete file: C:\Users\jnord\AppData\Local\Temp\remoting3335497741923094824cache\7F\CE79BB8C55572F4D2C34ABECD36F76.jar
	at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:2106)
...
```
 which happens in the cleanup code.

src/test/java/hudson/remoting/PrefetchingTest.java
jglick: General comment for the future.

jtnord:
master branch failures at https://gist.github.com/jtnord/c99dfd79b871d2da92c5

jglick:
So there are probably multiple problems here. Your error is the one I would expect, based on the system temp directory, not the `target/` path of Remoting binaries. If behavior differs according to whether `-Dtest` is passed or not, then consider passing `forkMode=always` in Surefire configuration.

jtnord:
I will hold off merging this then to see if I can uncover the issue you are seeing on Linux.

jglick:
No need, I think. This clearly solves a bug in the test.

PR #32, opened 2015-02-20
A shutdown Executor should not be able to resurect itself

jtnord:
This looked suspect to me.
It may have unintended consequences - but I believe that if it does reveal any issue with this change then the caller is actually broken.

@reviewbybees


stephenc:
I would tend to concur. :+1: from me but this critically needs a review by @kohsuke before merging

jglick:
:+1: (+ @kohsuke review advisable) though I doubt anyone is actually asking to shut down the executor.

kohsuke:
I talked to @jtnord and we reordered `q.add` and `isShutdown()` check.

We discussed a bit about the need to synchronize `isShutdown()` and `isTerminated()` methods but we didn't agree if it's necessary/useful/etc so we let that go for now.

PR #33, opened 2015-02-20
[JENKINS-21251] Slave channel timeout seemingly caused by usage of System.currentTimeMillis()

jtnord:
Initial calculation of wait time was susceptible to clock drift.

The calculation on how long to wait in various places on the code was
susceptible to the clock changing as multiple calls where made to
System.currentTimeMillis() for the initial calculation.

Changed this so that we use a single call to System.nanoTime and made just
a single call for the initial calculation.

There is still a potential issue for any callers of Channel.getLastHeard()
this will be addressed in a future change and has only been noted in this
commit.

Replaces Pull #29 which was getting multiple different fixes intertwined.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
+++ b/src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
@@ -71,10 +71,11 @@ public class AtmostOneThreadExecutor extends AbstractExecutorService {
 
     public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
         synchronized (q) {
-            long start = System.currentTimeMillis();
-            long end = start+unit.toMillis(timeout);
-            while (isAlive() && System.currentTimeMillis()<end) {
-                q.wait(end-System.currentTimeMillis());
+            long now = System.nanoTime();
+            long end = now + unit.toNanos(timeout);
+            while (isAlive() && (end - now > 0L)) {
+                q.wait(TimeUnit.NANOSECONDS.toMillis(end - now));
+                now = System.nanoTime();
             }
         }
         return isTerminated();
--- a/src/main/java/hudson/remoting/PingThread.java
+++ b/src/main/java/hudson/remoting/PingThread.java
@@ -23,16 +23,15 @@
  */
 package hudson.remoting;
 
-import org.jenkinsci.remoting.Role;
 import org.jenkinsci.remoting.RoleChecker;
 
 import java.io.IOException;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.concurrent.ExecutionException;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Logger;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
  * Periodically perform a ping.
@@ -42,7 +41,7 @@ import java.util.logging.Logger;
  * or when the disconnection is not properly detected.
  *
  * <p>
- * {@link #onDead()} method needs to be overrided to define
+ * {@link #onDead()} method needs to be overridden to define
  * what to do when a connection appears to be dead.
  *
  * @author Kohsuke Kawaguchi
@@ -81,14 +80,15 @@ public abstract class PingThread extends Thread {
     public void run() {
         try {
             while(true) {
-                long nextCheck = System.currentTimeMillis()+interval;
+                long nextCheck = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(interval);
 
                 ping();
 
                 // wait until the next check
                 long diff;
-                while((diff=nextCheck-System.currentTimeMillis())>0)
-                    Thread.sleep(diff);
+                while((diff = nextCheck - System.nanoTime()) > 0) {
+                    Thread.sleep(TimeUnit.NANOSECONDS.toMillis(diff));
+                }
             }
         } catch (ChannelClosedException e) {
             LOGGER.fine(getName()+" is closed. Terminating");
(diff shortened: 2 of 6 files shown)


src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
```diff
@@ -71,10 +71,11 @@ public boolean isTerminated() {
 
     public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
         synchronized (q) {
-            long start = System.currentTimeMillis();
-            long end = start+unit.toMillis(timeout);
-            while (isAlive() && System.currentTimeMillis()<end) {
-                q.wait(end-System.currentTimeMillis());
+            long now = System.nanoTime();
+            long end = now + unit.toNanos(timeout);
+            while (isAlive() && (end - now > 0L)) {
```
jglick: Note that for `timeout == 0`, this will now `wait(0)`, whereas the original code would have skipped the loop altogether. Probably does not matter, just FYI.

src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
```diff
@@ -71,10 +71,11 @@ public boolean isTerminated() {
...
+                q.wait(TimeUnit.NANOSECONDS.toMillis(end - now));
```
jglick: Could use the two-arg overload of `Object.wait` to specify a precise nanosecond time, though at best you could save a millisecond of waiting time that way, so maybe not worth the effort.

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -103,13 +103,13 @@ public void run() {
...
 
-        long remaining;
+        long end = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeout);
+        long remaining = end - System.nanoTime();
+
         do {
-            remaining = end-System.currentTimeMillis();
             try {
-                f.get(Math.max(0,remaining),MILLISECONDS);
+                f.get(Math.max(TimeUnit.MILLISECONDS.toNanos(10),remaining),TimeUnit.NANOSECONDS);
```
jglick: Why the change from 0?

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -120,9 +120,10 @@ private void ping() throws IOException, InterruptedException {
                 // get method waits "at most the amount specified in the timeout",
                 // so let's make sure that it really waited enough
             }
+            remaining = end - System.nanoTime();
         } while(remaining>0);
 
-        onDead(new TimeoutException("Ping started on "+start+" hasn't completed at "+System.currentTimeMillis()));//.initCause(e)
+        onDead(new TimeoutException("Ping started at "+start+" hasn't completed by "+System.currentTimeMillis()));//.initCause(e)
```
jglick: Not related to this PR, but probably this should use time formats, not raw `long`s. (Most easily: `new Date(…)`)

src/main/java/hudson/remoting/Request.java
```diff
@@ -256,21 +256,22 @@ public RSP get() throws InterruptedException, ExecutionException {
...
+                    while (response == null && (end - now > 0L)) {
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1,end-now));
+                        Request.this.wait(Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now)));
+                        now = System.nanoTime();
                     }
-                    if(response==null)
+                    if (response == null)
```
jglick: Try to avoid whitespace changes to unrelated lines.

jglick:
:+1:

src/main/java/hudson/remoting/PingThread.java
stephenc: Also what is the logic in choosing 10ms... for GigE the latency between the master and slave should be less than 10ms and all you really want to have happen is that it wakes up at all, it would seem a 1ns minimum wait time might make more sense than 10ms... additionally you probably want this to wake up anyway every 10 seconds or so to allow you to check if the channel has fallen down and bomb out faster, so perhaps you want a `Math.min(TimeUnit.SECONDS.toNanos(10), Math.max(1, remaining))` instead so that it kind-of spin loops

src/main/java/hudson/remoting/Request.java
```diff
@@ -256,21 +256,22 @@ public RSP get() throws InterruptedException, ExecutionException {
...
-                synchronized(Request.this) {
+                synchronized (Request.this) {
                     // wait until the response arrives
                     // Note that the wait method can wake up for no reasons at all (AKA spurious wakeup),
-                    long end = System.currentTimeMillis() + unit.toMillis(timeout);
-                    long now;
-                    while(response==null && (now=System.currentTimeMillis())<end) {
+                    long now = System.nanoTime();
+                    long end = now + unit.toNanos(timeout);
+                    while (response == null && (end - now > 0L)) {
```
stephenc: what's wrong with `end > now` BTW

src/main/java/hudson/remoting/SynchronousExecutorService.java
```diff
@@ -33,12 +33,16 @@ public synchronized boolean isTerminated() {
...
-        long end = System.currentTimeMillis() + unit.toMillis(timeout);
+        long now = System.nanoTime();
+        long end = now + unit.toNanos(timeout);
 
         while (count!=0) {
-            long d = end - System.currentTimeMillis();
-            if (d<0)    return false;
-            wait(d);
+            long d = end - now;
+            if (d<0) {
```
stephenc: I think you mean `(d<=0)`

src/main/java/hudson/remoting/PingThread.java
jtnord: There is very little point waiting 0ns and expecting anything different to happen as we are waiting on remote calls it seemed silly to wait anything much smaller.

src/main/java/hudson/remoting/Request.java
jtnord: you need to account for overflow of the 64bit number.  It starts from an undetermined number - could easily be `Long.MAX_VALUE - 3` so end could be negative and now could be positive yet we still need to sleep.

src/main/java/hudson/remoting/SynchronousExecutorService.java
jtnord: yes, otherwise we wait forever :)

src/main/java/hudson/remoting/AtmostOneThreadExecutor.java
jtnord: if `timeout == 0` then `end = now + 0`  so (`end - now > 0 == false`) so we never enter the loop and call `wait(0)`

src/main/java/hudson/remoting/PingThread.java
jtnord: @stephenc if the channel is closed during this call then f.get(...) should throw an exception if I follow the code correctly.
So this should return as soon as
1) the result is available
or
2) the channel is closed
or
3) The timeout happens
or
4) the JVM does a spurious wakeup...

stephenc:
:+1:

PR #35, opened 2015-02-20
Prevent potential memory leak with ExportTable

christ66:
Switch to using IdentityHashMap instead of two tables. Previous implementation of the ExportTable used two hash maps to maintain reverse order in the table. Switching to using one hashmap with a unique identity reduces the amount of memory consumed by the ExportTable in the event a hashCode is written poorly for any Callables.

### DIFF
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -23,16 +23,13 @@
  */
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
+
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.ArrayList;
+import java.util.*;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.SEVERE;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -43,8 +40,11 @@
  * @author Kohsuke Kawaguchi
  */
 final class ExportTable {
-    private final Map<Integer,Entry<?>> table = new HashMap<Integer,Entry<?>>();
-    private final Map<Object,Entry<?>> reverse = new HashMap<Object,Entry<?>>();
+    /**
+     * Entry table map based on unique id.
+     */
+    private final Map<Integer, Entry<?>> etable = new IdentityHashMap<Integer, Entry<?>>();
```
jglick: Huh?? `IdentityHashMap<Integer,T>` makes little sense. The keys are autoboxed `int`s; the `System.identityHashCode` is meaningless for these.

Did you perhaps mean to leave `table` along and just make `reserve` an `IdentityHashMap` with the same signature?

jglick:
What exactly is the problem you are trying to solve? Some class somewhere with an incorrect `hashCode` implementation? But you are still relying on `equals` being correct.

The description also talks about `Callable`, whereas an `ExportTable` is for managing RPC calls to objects held on the remote side—an unrelated kind of `Request`.

christ66:
@stephenc yes that's correct.

@jglick This should definitely have annotation with `Do Not Merge` and/or `WIP`. This was more of a test to try reducing the amount of memory used by the export table, and I created this PR to get thoughts.

PR #37, opened 2015-03-31
[WiP] - Coverity fixes

oleg-nenashev:
Just saved the local branch during the local repo cleanup. The PR should be reworked before the merge.


KostyaSha:
What do you expect rework? Change LGTM
:+1:

oleg-nenashev:
Needs a rework in any case, because it makes sense to start using Java7 features in the remoting layer

PR #39, opened 2015-04-15
When a channel goes down, kill off all Pipes.

kohsuke:
See the commit message for more details of this change.

This prevents a hang in channel communication.

### DIFF
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -35,7 +35,7 @@ import java.util.Map;
 import java.util.ArrayList;
 import java.util.logging.Logger;
 
-import static java.util.logging.Level.SEVERE;
+import static java.util.logging.Level.*;
 
 /**
  * Manages unique ID for exported objects, and allows look-up from IDs.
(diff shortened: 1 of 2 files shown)


kohsuke:
@reviewbybees

rsandell:
:+1: Assuming the test is just flaky.

src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -338,6 +338,38 @@ Object get(int id) {
...
+    public void abort(Throwable e) {
+        List<Entry<?>> values;
+        synchronized (this) {
+            values = new ArrayList<Entry<?>>(table.values());
+        }
+        for (Entry<?> v : values) {
+            if (v.object instanceof ErrorPropagatingOutputStream) {
+                try {
+                    ((ErrorPropagatingOutputStream)v.object).error(e);
+                } catch (IOException x) {
```
tfennelly: Maybe catch `Exception`? Otherwise propagation stops.

tfennelly: Or `Throwable`.

jtnord:
Checking locally if this fixes `PipeTest` that is reliably broken on a single core machine.

jtnord:
`PipeTest.testReaderCloseWhileWriterIsStillWriting-local-compatibility` is still failing locally (modified test to fail after 3 minutes).  Would you have expected that to be fixed here?

src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -338,6 +338,38 @@ Object get(int id) {
...
 
+    /**
+     * Propagate a channel termination error to all the exported objects.
+     * 
+     * <p>
+     * Exported {@link Pipe}s are vulnerable to infinite blocking
+     * when the channel is lost and the sender side is cut off. The reader
+     * end will not see that the writer has disappeared.
+     */
+    public void abort(Throwable e) {
```
jglick: Should probably be package-private.

kohsuke:
Code review comments addressed, but I still need to look at the test failure, it sounds like.

kohsuke:
The tests pass locally for me, and CI build passed for 2bb1421.

So I believe the failure in `PipeTest.testReaderCloseWhileWriterIsStillWriting-local-compatibility` is real but different issue, and maybe I can work with James at some point to dig deeper if he can reproduce this more easily.

I'm going to merge this PR.

PR #41, opened 2015-04-23
[JENKINS-26580] Add JNLP3 protocol which aims to provide basic level of security for JNLP slaves

akshayabd:
Apologies for the size of the pull request, if its too difficult to review I will be happy to break it into smaller chunks.

The JnlpProtocol3 class doc provides a detailed description of how the protocol works.

Summary of major changes by this pull request:
- Refactor/update Engine & JnlpProtocol to allow the protocols to create the Channel after a successful handshake instead of returning just the greeting response from the handshake. This was needed since Jnlp3 needs to wrap the socket streams in cipher streams.
- Added implementation of JNLP3 which does the handshake securely and create an encrypted Channel after a successful handshake.
- Some tweaks to JnlpProtocol2Test
- Updated version of mockito & powermockito

Notes:
- It would be nice to get this reviewed by people proficient in security protocols
- Right now the new protocol doesn't get used - in a future pull request it will be enabled, probably through a flag
- I've done some end-to-end testing but a lot more should be done, would be nice to get help with this
- The corresponding work to Jenkins-core is available at: https://github.com/jenkinsci/jenkins/compare/master...akshayabd:jnlp3-protocol?expand=1, I still need to add tests, a pull request will be created for that soon

Best,
Akshay


akshayabd:
I should mention, I based the handshake on challenge-response authentication:
http://en.wikipedia.org/wiki/Challenge%E2%80%93response_authentication#Cryptographic_techniques

ie. A shared secret that is previously known to both parties is used for mutual authentication.

src/main/java/org/jenkinsci/remoting/engine/jnlp3/ChannelCiphers.java
```diff
@@ -0,0 +1,100 @@
...
+            SecretKey secretKey = new SecretKeySpec(aesKey, "AES");
+            IvParameterSpec spec = new IvParameterSpec(specKey);
+            Cipher encryptCipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
+            encryptCipher.init(Cipher.ENCRYPT_MODE, secretKey, spec);
+            Cipher decryptCipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
+            decryptCipher.init(Cipher.DECRYPT_MODE, secretKey, spec);
+
+            return new ChannelCiphers(aesKey, specKey, encryptCipher, decryptCipher);
+        } catch (Exception e) {
+            throw new IOException("Failed to create channel ciphers", e);
```
akshayabd: I feel a bit uneasy about ChannelCiphers and HandshakeCiphers throwing IOExceptions. I feel they should throw an Exception and JnlpProtocol.establishChannel should be allowed to bubble up an Exception to Engine, which can be caught and handled.

akshayabd:
Ah - I realized I cant make a pull request for the Jenkins core changes till this is merged in and a new release it out. So I'll hold off on that pull request till this is reviewed and merged in.

src/main/java/hudson/remoting/ChunkedOutputStream.java
```diff
@@ -94,4 +94,4 @@ private void sendFrame(boolean hasMore) throws IOException {
         base.write(buf,0,size);
         size = 0;
     }
-}
\ No newline at end of file
+}
```
akshayabd: I'm trying to revert this but cant seem to get it to go back to the previous state - sorry about the non-relevant whitespace change.

akshayabd:
I've noticed the pull requests tests are sometimes timing out (at 4 mins). The tests are actually going through and its near the end when it gets aborted.

Can one of the admins increase the timeout to 5 mins? There are more and more tests being added so giving it some more time would be fair.

akshayabd:
k havent heard on this for a while, I'm going to start breaking this up into smaller pull requests

jglick:
@kohsuke ping

src/main/java/org/jenkinsci/remoting/engine/EngineUtil.java
```diff
@@ -58,6 +58,23 @@ public static String readLine(InputStream inputStream) throws IOException {
...
 
     /**
+     * Read a certain amount of characters from the stream.
+     *
+     * @param inputStream The input stream to read from.
+     * @param len The amount of characters to read.
+     * @return The characters read.
+     * @throws IOException
+     */
+    public static String readChars(InputStream inputStream, int len) throws IOException {
```
oleg-nenashev: It was a method for tests only. Makes sense to use package-internal wrapper next time

src/main/java/hudson/remoting/Engine.java
```diff
@@ -239,40 +232,36 @@ public void run() {
...
-                if (!connected) {
+                if (channel == null) {
                     onConnectionRejected("None of the protocols were accepted");
                     continue;
                 }
 
-                final Channel channel = new ChannelBuilder("channel", executor)
-                        .withJarCache(jarCache)
-                        .withMode(Mode.BINARY)
-                        .build(inputStream, new BufferedOutputStream(jnlpSocket.getOutputStream()));
```
stephenc: @oleg-nenashev the old code would keep the BufferedInputStream to be used when creating the channel, with this change however, the buffered input stream (and it's read-ahead buffer) is thrown away because [establishChannel](https://github.com/jenkinsci/remoting/pull/41/files#diff-2743d6a321ae5abd61a9ec59c4488a95R72) throws the Buffer away

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
```diff
@@ -49,16 +57,44 @@ protected JnlpProtocol(String secretKey, String slaveName) {
...
+     * @param channelBuilder The {@link ChannelBuilder} to use.
+     * @return The established channel if successful, else null.
+     * @throws IOException
+     */
+    @Nullable
+    public Channel establishChannel(Socket socket, ChannelBuilder channelBuilder) throws IOException {
+        DataOutputStream outputStream = new DataOutputStream(socket.getOutputStream());
+        BufferedInputStream inputStream = new BufferedInputStream(socket.getInputStream());
+        if(performHandshake(outputStream, inputStream)) {
+            return buildChannel(socket, channelBuilder);
```
stephenc: And here we throw away the Buffer and all its read-ahead input causing the capabilities to be inferred as 0 and then the channel transport is configured without chunking and bamb

```
INFO: Protocol failed to establish channel
java.io.StreamCorruptedException: invalid stream header: 0A6CACED
	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:806)
	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:299)
	at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:48)
	at hudson.remoting.ChannelBuilder.makeTransport(ChannelBuilder.java:430)
	at hudson.remoting.ChannelBuilder.negotiate(ChannelBuilder.java:389)
	at hudson.remoting.ChannelBuilder.build(ChannelBuilder.java:310)
	at org.jenkinsci.remoting.engine.JnlpProtocol2.buildChannel(JnlpProtocol2.java:93)
	at org.jenkinsci.remoting.engine.JnlpProtocol.establishChannel(JnlpProtocol.java:79)
	at hudson.remoting.Engine.run(Engine.java:245)
```

stephenc: Where `0A6C` is the chunk length and `ACED` is the start of the `ObjectOutputStream` in that chunk

akshayabd: I'm guessing this pull request introduced a bug? If so apologies - could you provide a link to the bug or more context?

I think I see what you're saying, here a buffered input stream is created for the socket which is then not used to create the channel, which means some bytes read in the original buffered input stream could be lost - is that correct?

akshayabd: I saw your pending pull request and related bug - thanks for catching this

lskillen:
@akshayabd Hey - I realise this is closed but we're currently investigating utilising JNLP3 for encrypted master/slave communications as an alternative to TCP over something like stunnel.  On JENKINS-26580 you've mentioned about optionally enabling the protocol for testing but this hasn't happened yet - Is there anything that we can do to progress this?  Cheers!

akshayabd:
There is some work going on to refactor how the master uses the remoting module. Once that is done we'll be able to enable JNLP3 for testing. Apologies, I'm not sure what the status of that work is - @kohsuke may be able to tell.

For reference:
https://github.com/jenkinsci/remoting/commit/678c4c64bb98ddf11a31874261064d044bfd2745

PR #42, opened 2015-04-23
[FIXED JENKINS-28062] A Channel.Listener's onClose method that propagate...

stephenc:
...s an exception is a sign of a bad listener not a problem closing the channel.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -55,6 +55,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
+import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 
 /**
@@ -829,8 +830,16 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             } // JENKINS-14909: leave synch block
         } finally {
             if (e instanceof OrderlyShutdown) e = null;
-            for (Listener l : listeners.toArray(new Listener[0]))
-                l.onClosed(this, e);
+            for (Listener l : listeners.toArray(new Listener[0])) {
+                try {
+                    l.onClosed(this, e);
+                } catch (Throwable t) {
+                    LogRecord lr = new LogRecord(Level.SEVERE, "Listener {0} propagated an exception for channel {1}'s close: {2}");
+                    lr.setThrown(t);
+                    lr.setParameters(new Object[]{l, this, t.getMessage()});
+                    logger.log(lr);
+                }
+            }
         }
     }
 


src/main/java/hudson/remoting/Channel.java
```diff
@@ -829,8 +830,16 @@ public void terminate(IOException e) {
             } // JENKINS-14909: leave synch block
         } finally {
             if (e instanceof OrderlyShutdown) e = null;
-            for (Listener l : listeners.toArray(new Listener[0]))
-                l.onClosed(this, e);
+            for (Listener l : listeners.toArray(new Listener[0])) {
+                try {
+                    l.onClosed(this, e);
+                } catch (Throwable t) {
+                    LogRecord lr = new LogRecord(Level.SEVERE, "Listener {0} propagated an exception for channel {1}'s close: {2}");
```
jtnord: This would be so much cleaner with slf4j...

src/main/java/hudson/remoting/Channel.java
```diff
@@ -829,8 +830,16 @@ public void terminate(IOException e) {
...
+                    lr.setThrown(t);
+                    lr.setParameters(new Object[]{l, this, t.getMessage()});
```
jtnord: so we will see the exceptions message twice?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -829,8 +830,16 @@ public void terminate(IOException e) {
             } // JENKINS-14909: leave synch block
         } finally {
             if (e instanceof OrderlyShutdown) e = null;
-            for (Listener l : listeners.toArray(new Listener[0]))
-                l.onClosed(this, e);
+            for (Listener l : listeners.toArray(new Listener[0])) {
+                try {
+                    l.onClosed(this, e);
+                } catch (Throwable t) {
```
jtnord: As this is a listener should we be catching Errors?
Would RuntimeException not be better here?

src/main/java/hudson/remoting/Channel.java
stephenc: Agree, but Jenkins uses JUL

src/main/java/hudson/remoting/Channel.java
stephenc: Yes because sometimes the exception can get separated from the message, also some people use custom formatters that drop the stack trace (bold people)

src/main/java/hudson/remoting/Channel.java
jtnord: yeah, a pet peve of mine.

src/main/java/hudson/remoting/Channel.java
stephenc: Debatable. e.g. if it was an OutOfMemory error it may be that closing this channel will resolve the issue anyway. Now we could do the

```
} catch (RuntimeException e) { 
  // log
} catch (Error e) {
  throw e;
} catch (Throwable t) {
  // log
}
```
dance if you feel really strongly about it... though all that will do is propagate the Error to Channel#L487 where it gets caught ant logged (incorrectly) as a failure to create the close command

stephenc: The point being that when the listener is being called, the channel has actually been closed, so we don't want to log a failure to close where it is actually a fault of the listener

jtnord: fair enough.

src/main/java/hudson/remoting/Channel.java
jtnord: if you drop the stack trace in a JUL formatter you still get the message (I'm scarred by this).
If you drop logging the Throwable all together then I don't care about supporting those people and they need to learn.
JUL should log this atomically as the StreamHandler users a Writer that has internal locking semantics.  Again if this doesn't happen then IMHO its a buggy 3rd party logger implementation (likely not provided by the JDK).

Anyway - I'm not passionate enough about this here to vote -1 based on this

jtnord:
:+1: despite my comments on the use of JUL bits.

jglick:
:+1: (Other than the gratuitously complicated logger call which could have been a single line, but whatever.)

PR #43, opened 2015-04-28
[JENKINS-28127] The Checksum caches on the master should be singletons. ...

akshayabd:
...This way the master does not have to wastefully re-calculate file hashes for each slave interaction. The caches are ConcurrentHashMap so they should be thread-safe. The TODO comment mentioned moving the caches to a separate class, but thats not needed, they can just be made static.

### DIFF
--- a/src/main/java/hudson/remoting/JarLoaderImpl.java
+++ b/src/main/java/hudson/remoting/JarLoaderImpl.java
@@ -74,8 +74,8 @@ class JarLoaderImpl implements JarLoader, Serializable {
             throw new AssertionError(e);
         }
 
-        knownJars.put(v,jar);
-        checksums.put(jar,v);
+        knownJars.putIfAbsent(v,jar);
+        checksums.putIfAbsent(jar,v);
         return v;
     }
 
--- a/src/main/java/hudson/remoting/JarLoaderImpl.java
+++ b/src/main/java/hudson/remoting/JarLoaderImpl.java
@@ -17,8 +17,6 @@ import java.util.concurrent.ConcurrentMap;
 /**
  * Implements {@link JarLoader} to be called from the other side.
  *
- * {@link JarLoaderImpl}s.
- *
  * @author Kohsuke Kawaguchi
  */
 @edu.umd.cs.findbugs.annotations.SuppressWarnings("SE_BAD_FIELD")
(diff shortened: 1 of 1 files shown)


akshayabd:
Hmm looks like the pull request test got stuck while downloading a jar? Can one of the admins plz abort the stuck job (if it is still stuck) and trigger a re-run for this pull request?

For future reference - can others issue commands to run tests again? If so how? Thanks.

akshayabd:
Kool - looks like closing/opening the pull request triggers a re-test.

Plz review.

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -15,22 +13,20 @@
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ConcurrentSkipListSet;
 
 /**
  * Implements {@link JarLoader} to be called from the other side.
  *
- * TODO: move {@link #knownJars} and {@link #checksums} to another class to share it across
  * {@link JarLoaderImpl}s.
```
oleg-nenashev: You forgot to remove the second TODO line

akshayabd: Thanks - removed

akshayabd:
Updated - PTAL

oleg-nenashev:
@akshayabd 
The class is serializable, and the change IMO breaks possible serialization cases. Examples:
* sharing of JAR registries between two masters
* dumping a JAR registry somewhere (e.g. Support Core plugin's extension points may use the serialization to save support bundles)

I suppose we should think twice before merging the change

akshayabd:
Hmmm ic.

I really would like to update the master-sides behavior though to stop re-calculating file hashes all the time. Let me think of another way to make this change that keeps an internal non-static cache.

akshayabd:
I've been thinking about this more, there may be a way to do this without removing the instance-specific caches. I'll wait for https://github.com/jenkinsci/remoting/pull/44 to be merged in first to avoid conflicts as that touches JarLoaderImpl too.

akshayabd:
I'll have another pull request - closing this one.

PR #44, opened 2015-04-28
[JENKINS-26759] Make FileSystemJarCache more defensive

akshayabd:
[JENKINS-26759](https://issues.jenkins-ci.org/browse/JENKINS-26759)

After a jar is fetched verify the checksum matches the expected value, if not throw an exception.

### DIFF
--- a/src/main/java/hudson/remoting/Checksum.java
+++ b/src/main/java/hudson/remoting/Checksum.java
@@ -2,7 +2,13 @@ package hudson.remoting;
 
 import java.io.ByteArrayInputStream;
 import java.io.DataInputStream;
+import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.net.URL;
+import java.security.DigestOutputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
 /**
  * Represents 128bit checksum of a jar file.
@@ -17,7 +23,7 @@ final class Checksum {
         this.sum2 = sum2;
     }
 
-    Checksum(byte[] arrayOf16bytes, int numOfLong) {
+    private Checksum(byte[] arrayOf16bytes, int numOfLong) {
         try {
             DataInputStream in = new DataInputStream(new ByteArrayInputStream(arrayOf16bytes));
             long l1=0,l2=0;
@@ -54,4 +60,41 @@ final class Checksum {
     public String toString() {
         return String.format("%016X%016X",sum1,sum2);
     }
+
+    /**
+     * Returns the checksum for the given file.
+     */
+    static Checksum forFile(File file) throws IOException {
+        return forURL(file.toURI().toURL());
+    }
+
+    /**
+     * Returns the checksum for the given URL.
+     */
+    static Checksum forURL(URL url) throws IOException {
+        try {
+            MessageDigest md = MessageDigest.getInstance(DIGEST_ALGORITHM);
+            Util.copy(url.openStream(), new DigestOutputStream(new NullOutputStream(), md));
+            return new Checksum(md.digest(), md.getDigestLength() / 8);
+        } catch (NoSuchAlgorithmException e) {
+            throw new AssertionError(e);
+        }
+    }
+
+    private static class NullOutputStream extends OutputStream {
+        @Override
+        public void write(int b) {
+        }
+
+        @Override
+        public void write(byte[] b) {
+        }
+
+        @Override
+        public void write(byte[] b, int off, int len) {
+        }
+    }
+
+    public static final String DIGEST_ALGORITHM = System.getProperty(
+            JarLoaderImpl.class.getName() + ".algorithm", "SHA-256");
 }
--- a/src/main/java/hudson/remoting/Checksum.java
+++ b/src/main/java/hudson/remoting/Checksum.java
@@ -73,7 +73,7 @@ final class Checksum {
      */
     static Checksum forURL(URL url) throws IOException {
         try {
-            MessageDigest md = MessageDigest.getInstance(DIGEST_ALGORITHM);
+            MessageDigest md = MessageDigest.getInstance(JarLoaderImpl.DIGEST_ALGORITHM);
             Util.copy(url.openStream(), new DigestOutputStream(new NullOutputStream(), md));
             return new Checksum(md.digest(), md.getDigestLength() / 8);
         } catch (NoSuchAlgorithmException e) {
@@ -94,7 +94,4 @@ final class Checksum {
         public void write(byte[] b, int off, int len) {
         }
     }
-
-    public static final String DIGEST_ALGORITHM = System.getProperty(
-            JarLoaderImpl.class.getName() + ".algorithm", "SHA-256");
 }
--- a/src/main/java/hudson/remoting/FileSystemJarCache.java
+++ b/src/main/java/hudson/remoting/FileSystemJarCache.java
@@ -71,8 +71,14 @@ public class FileSystemJarCache extends JarCacheSupport {
                 tmp.renameTo(target);
 
                 if (target.exists()) {
+                    Checksum expected = new Checksum(sum1, sum2);
+                    Checksum actual = Checksum.forFile(target);
+                    System.out.println(actual.toString());
+                    if (!expected.equals(actual)) {
+                        throw new IOException("Checksum of retrieved jar doesn't match: " + target);
+                    }
                     // even if we fail to rename, we are OK as long as the target actually exists at this point
-                    // this can happen if two FileSystejarCache instances share the same cache dir
+                    // this can happen if two FileSystemJarCache instances share the same cache dir
                     return target.toURI().toURL();
                 }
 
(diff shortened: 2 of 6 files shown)


akshayabd:
@jglick plz review

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -71,8 +71,13 @@ protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException
...
 
                 if (target.exists()) {
+                    Checksum expected = new Checksum(sum1, sum2);
+                    Checksum actual = Checksum.forFile(target);
+                    if (!expected.equals(actual)) {
+                        throw new IOException("Checksum of retrieved jar doesn't match: " + target);
+                    }
                     // even if we fail to rename, we are OK as long as the target actually exists at this point
-                    // this can happen if two FileSystejarCache instances share the same cache dir
+                    // this can happen if two FileSystemJarCache instances share the same cache dir
```
akshayabd: I think this comment can be removed, it doesnt make sense to me. Agreed?

src/main/java/hudson/remoting/Checksum.java
```diff
@@ -54,4 +60,38 @@ public int hashCode() {
...
+        try {
+            MessageDigest md = MessageDigest.getInstance(JarLoaderImpl.DIGEST_ALGORITHM);
+            Util.copy(url.openStream(), new DigestOutputStream(new NullOutputStream(), md));
+            return new Checksum(md.digest(), md.getDigestLength() / 8);
+        } catch (NoSuchAlgorithmException e) {
+            throw new AssertionError(e);
+        }
+    }
+
+    private static class NullOutputStream extends OutputStream {
```
jglick: BTW this already exists in Apache Commons I/O and Guava, so if either of these are on the classpath, it might be nice to reuse them.

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -71,8 +71,13 @@ protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException
                 tmp.renameTo(target);
 
                 if (target.exists()) {
+                    Checksum expected = new Checksum(sum1, sum2);
+                    Checksum actual = Checksum.forFile(target);
+                    if (!expected.equals(actual)) {
+                        throw new IOException("Checksum of retrieved jar doesn't match: " + target);
```
jglick: I would suggest including the two checksums in the detail message, for possible future diagnosis.

src/main/java/hudson/remoting/FileSystemJarCache.java
jglick: It makes sense. Two instances can indeed share a cache directory. So a failure in `renameTo` is OK so long as the target actually exists _and_ the checksum is still correct.

jglick: Though I would expect the checksum verification to happen on `tmp` _before_ the rename. Otherwise we leave corrupt content behind for the next call. So:

* check for cache file; if present, use it, stop now
* download to `tmp`
* verify checksum
* try to rename to real file
   * if that succeeded, OK, stop
   * if that failed
      * and the real file is not there, die
      * and it is (a race condition with two caches downloading the same file at the same time), verify that checksum (should be correct, but just in case)

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -1,28 +1,19 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ConcurrentSkipListSet;
 
 /**
  * Implements {@link JarLoader} to be called from the other side.
  *
- * TODO: move {@link #knownJars} and {@link #checksums} to another class to share it across
- * {@link JarLoaderImpl}s.
- *
```
jglick: I think this comment is still potentially appropriate.

src/test/java/hudson/remoting/ChecksumTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertEquals(createdExpectedChecksum(hash1), Checksum.forURL(tmpFile1.toURI().toURL()));
+
+        assertEquals(createdExpectedChecksum(hash2), Checksum.forFile(tmpFile2));
+        assertEquals(createdExpectedChecksum(hash2), Checksum.forURL(tmpFile2.toURI().toURL()));
+
+        assertNotEquals(Checksum.forFile(tmpFile1), Checksum.forFile(tmpFile2));
+    }
+
+    private File createTmpFile(String name, String contents) throws Exception {
+        File tmpFile = File.createTempFile(name, ".txt");
```
jglick:
```java
@Rule public TemporaryFolder tmp = new TemporaryFolder();
```

src/test/java/hudson/remoting/FileSystemJarCacheTest.java
```diff
@@ -0,0 +1,97 @@
...
+    public ExpectedException expectedEx = ExpectedException.none();
+
+    @Mock private Channel mockChannel;
+    @Mock private JarLoader mockJarLoader;
+    private FileSystemJarCache fileSystemJarCache;
+    private Checksum expectedChecksum;
+
+    @Before
+    public void setUp() throws Exception {
+        File rootDir = Files.createTempDir();
```
jglick: `TemporaryFolder` rule

src/main/java/hudson/remoting/Checksum.java
akshayabd: Unfortunately neither of them are :(.

Guava is there but only for tests. I feel like it should be added as a dependency for remoting, I've come across a few places where having it would have been nice. It would increase the slave.jar size, but I haven't looked into by how much.

I'd prefer to leave adding Guava to a separate pull request if its decided to be included. Let me know if it is - I'll be happy to bring it in.

src/main/java/hudson/remoting/FileSystemJarCache.java
akshayabd: Done.

src/main/java/hudson/remoting/JarLoaderImpl.java
akshayabd: Sorry - this was meant only for https://github.com/jenkinsci/remoting/pull/43 - reverting.

src/main/java/hudson/remoting/FileSystemJarCache.java
akshayabd: Ack - updating

src/test/java/hudson/remoting/ChecksumTest.java
akshayabd: Ack

src/test/java/hudson/remoting/FileSystemJarCacheTest.java
akshayabd: Ack

akshayabd:
Updates made - PTAL.

Do you have suggestions on testing the rename failing other than switching to PowerMock?

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -56,6 +56,14 @@ protected URL lookInCache(Channel channel, long sum1, long sum2) throws IOExcept
     protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException, InterruptedException {
         File target = map(sum1, sum2);
         File parent = target.getParentFile();
+
+        if (target.exists()) {
+            // Assume its already been fetched correctly before. ie. We are not going to validate
+            // the checksum.
+            LOGGER.fine(String.format("Jar file already exists: %16X%16X", sum1, sum2));
+            return target.toURI().toURL();
```
akshayabd: Hmmm it may be worth verifying the checksum in this case too?

akshayabd:
ping - @jglick - should we verify the checksum even if the target already exists?

Also if everything else looks ok I'd like to get this merged in - I have some other changes I'm working on that depend on this pull request.

src/main/java/hudson/remoting/Checksum.java
jglick: Nah, forget it.

jglick:
> suggestions on testing the rename failing

Not really, it is hard to test this sort of thing.

src/main/java/hudson/remoting/FileSystemJarCache.java
jglick: Under the assumption that this is the rare case and so not performance-sensitive, yes it may be worth checking here. I do not know offhand if this is called only as a race condition, or if this is a normal cache hit which must be fast.

akshayabd: ok - i'll update it to verify the checksum here as well

jglick:
Assuming you have done some kind of sanity check (e.g., building Jenkins core with this patched remoting and verified that you can start a slave on a node with no cache dir, run a build, disconnect, reconnect, and run another build), :+1:

src/main/java/hudson/remoting/Checksum.java
kohsuke: Trying to keep remoting small in terms of dependency footprint as it's used by CLI and etc.

PR #48, opened 2015-05-06
Use the latest release of guava.

deadmoose:
Guava 14 was released ~2.5 years ago.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>14.0</version>
+      <version>18.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>


oleg-nenashev:
AFAIK there was an attempt to update Guava, but it was reverted due to compatibility issues

deadmoose:
This is in the specific context of running tests on remoting. It isn't actually used in the real code, and has only ever been using this one specific version since it was introduced in a504a5624fbad8f90ffddc9beb68b41a4d957fd6

I'll believe that e.g. updating Jenkins core to modern guava is a larger ordeal; it's on an even more ancient 11.0.1 (old enough that updating skips over an entire deprecate -> api break cycle), but this should be fine. If remoting were leaking its dependency up the tree, that would likely *already* cause issues since it's not old enough.

akshayabd:
+1 to this change

KostyaSha:
Hi, do you have any specific issues with current version?

deadmoose:
I'm a little fuzzy off the top of my head after ~3 months, but we've got a few internal feature adds and IIRC some of the test coverage for that used some newer guavaisms.

That said, like I mentioned before, in remoting it really is a test-only dependency, so it's pretty straight forward and safe; it doesn't leak to consumers of the library, and it demonstrably works there by virtue of the tests working.

kohsuke:
This is a test dependency, I don't see any big deals here.

PR #47, opened 2015-05-05
Unflake the JnlpProtocol2 tests

akshayabd:
Originally part of https://github.com/jenkinsci/remoting/pull/41. I'm breaking it out into its own pull request so it can get reviewed & merged in sooner.

The current JnlpProtocol2 tests incorrectly try to mock out Properties call to new Date(). This results in the tests sometimes failing. I could not get Powermock to correctly mock out the call and instead chose to write an ArgumentMatcher that would compare string-ified properties as needed.

PR #51, opened 2015-05-07
Deflake JnlpProtocol2Test

deadmoose:
PowerMockito wasn't being told to prep Properties.class to have its
'new Date()' call hijacked, so it wasn't actually doing it. That said,
even adding that didn't seem to help.

But instead of reaching into the implementation details, I've written a
custom Matcher to just verify that the output string representation of
the Properties parses back to the expected values. As an added benefit,
this no longer depends on two identical Properties objects writing keys
in the same order when store() is called, since the API there doesn't
actually guarantee that.

### DIFF
--- a/src/test/java/org/jenkinsci/remoting/engine/JnlpProtocol2Test.java
+++ b/src/test/java/org/jenkinsci/remoting/engine/JnlpProtocol2Test.java
@@ -27,6 +27,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.InOrder;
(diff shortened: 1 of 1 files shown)


akshayabd:
There is already a pull request to do this:
https://github.com/jenkinsci/remoting/pull/47

akshayabd:
This approach is better - closing https://github.com/jenkinsci/remoting/pull/47

jglick:
IMO this is OK to merge as is but @lacostej could file a follow-up cleaning things up a little.

kohsuke:
I agree, a room for further improvement shouldn't block the base change itself.

PR #52, opened 2015-05-18
[JENKINS-28127] Checksum caching to improve performance when dealing with high number of slaves

akshayabd:
Only calculate checksums for Jars once. This change reduces the stress on the Jenkins master when several slaves connect to it simultaneously, allowing Jenkins to scale more gracefully when dealing with a high number of ephemeral slaves that comes & go. This is particularly helpful when there are large Jars involved, as the case may be in some custom plugins or if the user puts custom jars in the Jenkins classpath.

### DIFF
--- a/src/main/java/hudson/remoting/Checksum.java
+++ b/src/main/java/hudson/remoting/Checksum.java
@@ -9,6 +9,8 @@ import java.net.URL;
 import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * Represents 128bit checksum of a jar file.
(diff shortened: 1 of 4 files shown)


akshayabd:
FYI - I think this change eliminates the need to try to get Channels to share JarLoaders, thats why I removed the related TODO comments.

akshayabd:
The Jenkins.war's WEB-INF/lib has around 125 jars - not all of which get transferred to the slave. If someone is using a good number of plugins that run on the slave then maybe a couple hundred jars would be involved? I would think a ConcurrentMap of that size should not be a problem. We don't have access to guava (its not a BUILD dependency of remoting, and from previous conversations it looks like it won't be) so we're not able to use guava caches.

src/main/java/hudson/remoting/Checksum.java
```diff
@@ -94,4 +128,8 @@ public void write(byte[] b) {
         public void write(byte[] b, int off, int len) {
         }
     }
+
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings("DMI_COLLECTION_OF_URLS")
```
KostyaSha: Please use http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html

PR #53, opened 2015-05-22
Swap PingThread#dead* abstraction

lacostej:
I noticed that the way PingThread#onDead(*) methods are deprecated/implemented could cause clients to not implement the method that gives the most feedback.

So here's a new implementation that
1. force subclasses to implement the most meaningful method
2. in a separate commit, get rid of the old method altogether

Existing clients should have to implement the new method anyway. The old one isn't used (except maybe by implementing clients?) and the proposed default implementation might give more feedback if ever called.

I am unable to run full tests due to a PipeTest hang when running mvn test.

### DIFF
--- a/src/main/java/hudson/remoting/PingThread.java
+++ b/src/main/java/hudson/remoting/PingThread.java
@@ -41,7 +41,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
  * or when the disconnection is not properly detected.
  *
  * <p>
- * {@link #onDead()} method needs to be overridden to define
+ * {@link #onDead(Throwable)} method needs to be overridden to define
  * what to do when a connection appears to be dead.
  *
  * @author Kohsuke Kawaguchi
@@ -141,9 +141,7 @@ public abstract class PingThread extends Thread {
      *
      * @since 2.9
      */
-    protected void onDead(Throwable diagnosis) {
-        onDead();   // fall back
-    }
+    protected abstract void onDead(Throwable diagnosis);
 
     private static final class Ping implements Callable<Void, IOException> {
         private static final long serialVersionUID = 1L;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/PingThread.java
```diff
@@ -134,16 +134,16 @@ private void ping() throws IOException, InterruptedException {
      *      and provide a fallback behaviour to be backward compatible with earlier version of remoting library.
      */
     @Deprecated
-    protected abstract void onDead();
+    protected void onDead() {
+        onDead(new RuntimeException("No cause"));
```
oleg-nenashev: What about a <code>@CheckForNull</code> method? seems to be a more convenient solution

lacostej: I am not sure what you mean here! :)

oleg-nenashev: Sorry for missing the response. My idea was to put the null check logic to <code>protected void onDead(Throwable diagnosis)</code>, but it's need

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -134,16 +134,16 @@ private void ping() throws IOException, InterruptedException {
...
 
     /**
      * Called when ping failed.
      *
      * @since 2.9
      */
-    protected void onDead(Throwable diagnosis) {
-        onDead();   // fall back
-    }
+    protected abstract void onDead(Throwable diagnosis);
```
oleg-nenashev: This is a public class, which could be used somewhere else. Probably you could just add a non-fatal default implementation. E.g.:

```java
protected void onDead(@CheckForNull Throwable diagnosis) {
    System.err.println("Ping failed. Cause: " + diagnosis);
    if (diagnosis != null) {
        diagnosis.printStackTrace(System.err);
    }
}
```

oleg-nenashev:
retriggered the CI

oleg-nenashev:
The PR needs to be polished a bit. I'll try to handle it

jeffret-b:
This has been open for a long time with no progress. The ideas might be useful but I don't see a priority. I'm closing it. If someone wants to finish it up, please re-open.

PR #54, opened 2015-06-10
[JENKINS-28844] Fix memory leak

stephenc:
See [JENKINS-28844](https://issues.jenkins-ci.org/browse/JENKINS-28844) 

* In larger installations of Jenkins the complex cycles of object references containing classloaders containing references to the channel which contains references back to the classloaders can confuse the garbage collector.
* The confusion arrises because classloaders are considered live while there are objects from the classloader live
      and the RemoteClassLoader and RemoteInvocationHandler's proxy both need a reference to the Channel
      which typically can be storing either some of the objects from the RemoteClassLoader or
      the RemoteInvocationHandler's proxy instances in the Channel properties. Thus although the entire subgraph
      is no longer live, the cycle is not detected as ClassLoaders are treated separately. The issue is worse in
      Java 7 where the ClassLoaders are on the PermGen heap though it is also present in Java 8.
* Typically multiple out of memory errors can cause the cycle to ultimately get cleaned up as one half
      gets cleared out breaking the cycle enough so that a subsequent out of memory removes the second half leaving
      a third out of memory to clear out the ClassLoaders... but after three out of memory errors in a row, Jenkins
      itself is typically well dead.
    
This PR (I believe) fixes these issues:
* We apply [Wheeler's theorem](http://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering) to
      the Channel references. This allows us to clear out the reference instance as soon as the Channel is terminated
      thus simplifying the work of the garbage collector.
* We clear out some unnecessary Channel references that were hiding in ThreadLocals
* We move from using an Object.finalize() implementation to a PhantomReference implementation of the
      unexporting operation. In my experiments the call to Object.finalize() on a running Jenkins instance would
      very often get postponed to almost never, and more critically, the presence of a finalizer means that all
      instances of RemoteInvocationHandler were put in the finalizer queue as opposed to only those that
      need to be tracked for auto unexporting. We can (and do) proactively remove the references from the
      reference queue as soon as the channel is closed, further reducing the work that is required by the
      garbage collector.
* Some other findbugs related fixes and adding a thread name since I copied the naming thread factory to provide a name for the unexporter thread

@reviewbybees (who may want to know that this references ZD-26018)

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -23,7 +23,6 @@
  */
 package hudson.remoting;
 
-import org.jenkinsci.remoting.CallableDecorator;
 import edu.umd.cs.findbugs.annotations.SuppressWarnings;
 import hudson.remoting.CommandTransport.CommandReceiver;
 import hudson.remoting.PipeWindow.Key;
@@ -31,8 +30,10 @@ import hudson.remoting.PipeWindow.Real;
 import hudson.remoting.forward.ForwarderFactory;
 import hudson.remoting.forward.ListeningPort;
 import hudson.remoting.forward.PortForwarder;
+import org.jenkinsci.remoting.CallableDecorator;
 import org.jenkinsci.remoting.RoleChecker;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.Closeable;
 import java.io.IOException;
@@ -189,6 +190,14 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * and therefore should be OK.
      */
     private final WeakHashMap<PipeWindow.Key, WeakReference<PipeWindow>> pipeWindows = new WeakHashMap<PipeWindow.Key, WeakReference<PipeWindow>>();
+    /**
+     * There are cases where complex object cycles can cause a closed channel to fail to be garbage collected,
+     * these typically arrise when an {@link #export(Class, Object)} is {@link #setProperty(Object, Object)}
+     * (a supported and intended use case), the {@link Ref} allows us to break the object cycle on channel
+     * termination and simplify the circles into chains which can then be collected easily by the garbage collector.
+     * @since FIXME after merge
+     */
+    private final Ref reference = new Ref(this);
 
     /**
      * Registered listeners. 
@@ -497,6 +506,17 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
                 Channel.this.terminate(e);
             }
         });
+    }
+
+    /**
+     * Gets the {@link Ref} for this {@link Channel}. The {@link Ref} will be {@linkplain Ref#clear()}ed when
+     * the channel is terminated in order to break any complex object cycles.
+     * @return the {@link Ref} for this {@link Channel}
+     * @since FIXME after merge
+     */
+    @Nonnull
+    /*package*/ Ref ref() {
+        return reference;
     }
 
     /**
@@ -809,6 +829,10 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             synchronized (this) {
                 if (e == null) throw new IllegalArgumentException();
                 outClosed = inClosed = e;
+                // we need to clear these out early in order to ensure that a GC operation while
+                // proceding with the close does not result in a batch of UnexportCommand instances
+                // being attempted to send over the locked and now closing channel.
+                RemoteInvocationHandler.notifyChannelTermination(this);
                 try {
                     transport.closeRead();
                 } catch (IOException x) {
@@ -828,6 +852,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
                         executingCalls.clear();
                     }
                     exportedObjects.abort(e);
+                    // break any object cycles into simple chains to simplify work for the garbage collector
+                    reference.clear();
                 } finally {
                     notifyAll();
                 }
@@ -1123,7 +1149,14 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     public synchronized void close(Throwable diagnosis) throws IOException {
         if(outClosed!=null)  return;  // already closed
 
-        send(new CloseCommand(this,diagnosis));
+        try {
+            send(new CloseCommand(this, diagnosis));
+        } catch (IOException e) {
+            // send should only ever - worst case - throw an IOException so we'll just catch that and not Throwable
+            logger.log(Level.WARNING, "Having to terminate early", e);
+            terminate(e);
+            return;
+        }
         outClosed = new IOException().initCause(diagnosis);   // last command sent. no further command allowed. lock guarantees that no command will slip inbetween
         notifyAll();
         try {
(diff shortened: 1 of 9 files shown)


jglick:
The test build got aborted for some reason. Timeout? Other cancellation?

jglick:
BTW removing `FIXED` annotation (which may not be honored by the JIRA link daemon anyway) since an actual fix involves integrating a new remoting release to core, so the issue should stay open until that happens.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -194,6 +190,14 @@
      * and therefore should be OK.
      */
     private final WeakHashMap<PipeWindow.Key, WeakReference<PipeWindow>> pipeWindows = new WeakHashMap<PipeWindow.Key, WeakReference<PipeWindow>>();
+    /**
+     * There are cases where complex object cycles can cause a closed channel to fail to be garbage collected,
+     * these typically arrise when an {@link #export(Class, Object)} is {@link #setProperty(Object, Object)}
+     * (a supported and intended use case), the {@link Ref} allows us to break the object cycle on channel
+     * termination and simplify the circles into chains which can then be collected easily by the garbage collector.
+     * @since FIXME after merge
```
jglick: No `@since` needed on non-API members. Anyone doing code archaeology should just `git-blame`.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -814,6 +829,10 @@ public void terminate(IOException e) {
             synchronized (this) {
                 if (e == null) throw new IllegalArgumentException();
                 outClosed = inClosed = e;
+                // we need to clear these out early in order to ensure that a GC operation while
+                // proceding with the close does not result in a batch of UnexportCommand instances
```
jglick: typo

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1484,4 +1516,77 @@ public static Channel current() {
         // to avoid situations like this, create proxy classes that we need during the classloading
         jarLoaderProxy=RemoteInvocationHandler.getProxyClass(JarLoader.class);
     }
+
+    private static class ThreadLastIoId extends ThreadLocal<int[]> {
```
jglick: IIRC this refactoring was to avoid a `$this` reference from an anonymous inner class, right? Might be worth it to mention that in Javadoc.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1484,4 +1516,77 @@ public static Channel current() {
...
+            return new int[1];
+        }
+    }
+
+    /**
+     * A reference for the {@link Channel} that can be cleared out on {@link #close()}/{@link #terminate(IOException)}.
+     * @since FIXME after merge
+     * @see #reference
+     */
+    /*package*/ static final class Ref {
```
jglick: BTW this whole class could probably be replaced with `AtomicReference<Channel>`, at the loss of a little clarity in naming.

src/main/java/hudson/remoting/NamingThreadFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Thread factory that sets thread name so we know who is responsible for so many threads being created.
+ * @since FIXME after merge
+ */
+public class NamingThreadFactory implements ThreadFactory {
```
jglick: I wonder if we should create a utilities module that can be shared with Jenkins core.

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -108,15 +133,43 @@
         // if the type is a JDK-defined type, classloader should be for IReadResolve
         if(cl==null || cl==ClassLoader.getSystemClassLoader())
             cl = IReadResolve.class.getClassLoader();
-        return type.cast(Proxy.newProxyInstance(cl, new Class[]{type,IReadResolve.class},
-            new RemoteInvocationHandler(channel,id,userProxy,autoUnexportByCaller,type)));
+        RemoteInvocationHandler handler = new RemoteInvocationHandler(channel, id, userProxy, autoUnexportByCaller, type);
```
jglick: It is legal to pass a null channel to this constructor?

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -175,11 +228,11 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         RPCRequest req = new RPCRequest(oid, method, args, userProxy ? dc.getClassLoader() : null);
         try {
             if(userProxy) {
-                if (async)  channel.callAsync(req);
-                else        return channel.call(req);
+                if (async)  channel().callAsync(req);
+                else        return channel().call(req);
```
jglick: And it it is null? Maybe there should be a `channelOrFail` method throwing `IOException`?

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -233,16 +292,217 @@ public int hashCode() {
...
+    /**
+     * Manages the cleanup of {@link RemoteInvocationHandler} instances that need to be auto unexported.
+     * @since FIXME after merge
+     */
+    private static class Unexporter implements Runnable {
+
+        /**
+         * Our executor service, we use at most one thread for all {@link Channel} instances in the current classloader.
+         */
+        private final ExecutorService svc = new AtmostOneThreadExecutor(
```
jglick: Should this be `static`?

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -233,16 +292,217 @@ public int hashCode() {
...
+         */
+        @Override
+        public void run() {
+            if (!isAlive.compareAndSet(false, true)) {
+                inQueue.set(false); // we have started execution so clear the flag, must happen after check of running
+                return;
+            }
+            inQueue.set(false); // we have started execution and running is true, so queued can be reset
+            try {
+                while (!referenceLists.isEmpty()) {
```
jglick: Too tricky to follow easily, so just trusting this is all correct…

jglick:
Minor comments only, :+1: nice work!

src/main/java/hudson/remoting/NamingThreadFactory.java
stephenc: Yep... or maybe just have Jenkins use this one directly now that it exists

src/main/java/hudson/remoting/RemoteInvocationHandler.java
stephenc: I couldn't find a case to prove it wasn't. We are dealing with remoting which will involve a network round-trip, so we are not CPU cycle critical and safer is better

src/main/java/hudson/remoting/RemoteInvocationHandler.java
stephenc: No need because `UNEXPORTER` is static. This leaves us with just the one static to track

src/main/java/hudson/remoting/RemoteInvocationHandler.java
stephenc: 395-426 gets one reference from the queue and cleans it up. 429-439 scans for any channels that have closed and removes their objects from even the phantom reference queue thus making it super easy for the GC to collect them... though having looked at that code in the cold light of day I now want to wrap the second one in a periodic run basis rather than every loop in case there are a large number of references cleared when there are many channels open

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -158,7 +158,7 @@ public void setConnectTo(String target) {
             System.err.println("Illegal parameter: "+target);
             System.exit(1);
         }
-        connectionTarget = new InetSocketAddress(tokens[0],Integer.valueOf(tokens[1]));
+        connectionTarget = new InetSocketAddress(tokens[0],Integer.parseInt(tokens[1]));
```
oleg-nenashev: Makes sense to process NumberFormatException and exit as well

oleg-nenashev:
The PR description has <code>@reviewbyees</code>, but should contain @reviewbybees
:bee: BTW

src/main/java/hudson/remoting/NamingThreadFactory.java
kohsuke: I'm more inclined just to hide it so that Jenkins core wouldn't start accidentally relying on this. Minor point.

PR #61, opened 2015-09-23
Add a utility main that can be used to initialize a node's jar cache

akshayabd:
Add a utility main that can populate a jar cache dir given another directory of jars.

This can be used to initialze a node's jar cache ahead of time before it connects to the master. This has performance benefits for the master as it won't have to transfer jars over to the node.

### DIFF
--- a/src/main/java/hudson/remoting/InitializeJarCacheMain.java
+++ b/src/main/java/hudson/remoting/InitializeJarCacheMain.java
@@ -1,0 +1,80 @@
+package hudson.remoting;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.FilenameFilter;
+
+/**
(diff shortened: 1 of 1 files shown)


akshayabd:
"Build timed out (after 5 minutes). Marking the build as aborted."

Could one of the admins look into this please? If its a time-out issue perhaps increase it?

oleg-nenashev:
Retriggering the build

oleg-nenashev:
At my previous company we used a similar approach via WinSW on Windows and [Startup plugin](https://wiki.jenkins-ci.org/display/JENKINS/Startup+Trigger). It LGTM to have this functionality in remoting directly.

Now the build fails on the deployment stage (I'm not sure it even needs it). To be investigated.

oleg-nenashev:
:+1:

PR #55, opened 2015-06-12
[JENKINS-28289] Ensure proxy exclusion list is used while getting proxy for JNLP clients

hypnoce:
Pull request as planned : https://issues.jenkins-ci.org/browse/JENKINS-28289

PR #62, opened 2015-10-01
[JENKINS-28289] Ensure proxy exclusion list is used while getting proxy for JNLP clients

evanx9:
Pull request as planned : https://issues.jenkins-ci.org/browse/JENKINS-28289

Ensure proxy exclusion list is used while getting proxy for JNLP
clients. Updated Jacoco version. Reworked version from
https://github.com/jenkinsci/remoting/pull/55

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -31,6 +31,7 @@ import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Socket;
 import java.net.URL;
@@ -304,31 +305,21 @@ public class Engine extends Thread {
         events.status(msg);
         int retry = 1;
         while(true) {
+        	boolean isHttpProxy = false;
+            InetSocketAddress targetAddress = null;
             try {
-                boolean isProxy = false;
-                Socket s;
-                if (System.getProperty("http.proxyHost") != null) {
-                    String proxyHost = System.getProperty("http.proxyHost");
-                    String proxyPort = System.getProperty("http.proxyPort", "80");
-                    s = new Socket(proxyHost, Integer.parseInt(proxyPort));
-                    isProxy = true;
-                } else {
-                    String httpProxy = System.getenv("http_proxy");
-                    if (httpProxy != null) {
-                        try {
-                            URL url = new URL(httpProxy);
-                            s = new Socket(url.getHost(), url.getPort());
-                            isProxy = true;
-                        } catch (MalformedURLException e) {
-                            System.err.println("Not use http_proxy environment variable which is invalid: "+e.getMessage());
-                            s = new Socket(host, Integer.parseInt(port));
-                        }
-                    } else {
-                        s = new Socket(host, Integer.parseInt(port));
-                    }
-                }
+            	Socket s = null;
+            	targetAddress = Util.getResolvedHttpProxyAddress(host, Integer.parseInt(port));
 
-                s.setTcpNoDelay(true); // we'll do buffering by ourselves
+            	if (targetAddress == null) {
+            		targetAddress = new InetSocketAddress(host, Integer.parseInt(port));;
+                } else {
+                	isHttpProxy = true;
+                }
+            	s = new Socket();
+            	s.connect(targetAddress);
+            	
+            	s.setTcpNoDelay(true); // we'll do buffering by ourselves
 
                 // set read time out to avoid infinite hang. the time out should be long enough so as not
                 // to interfere with normal operation. the main purpose of this is that when the other peer dies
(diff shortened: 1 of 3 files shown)


evanx9:
Looks like something is wrong with the artifact repo? The build portion looks good. @daniel-beck

evanx9:
@oleg-nenashev Noticed you are investigating the errors during deployment stage on another pull. Do let me know if the job can be fixed.

daniel-beck:
@evanx9 Don't worry, I think we can safely ignore brokenness like this that's unrelated to your changes.

oleg-nenashev:
retriggering the build

src/main/java/hudson/remoting/Engine.java
```diff
@@ -351,8 +342,15 @@ private Socket connect(String host, String port) throws IOException, Interrupted
...
-                if(retry++>10)
-                    throw (IOException)new IOException("Failed to connect to "+host+':'+port).initCause(e);
+            	if (retry++>10) {
+            		String suffix = "";
+            		if (isHttpProxy) {
+            			suffix = " through proxy " + targetAddress.toString();
+            		}
+            		// Cast -> InitCause method is returning a type Throwable.
+            		// Therefore we need to cast it to IOException.
+            		throw (IOException) new IOException("Failed to connect to " + host + ':' + port + suffix).initCause(e);
```
oleg-nenashev: NIT: I would decouple it to 3 lines and avoid any casts. BTW this approach should also work

src/main/java/hudson/remoting/Util.java
```diff
@@ -131,4 +136,37 @@ static URLConnection openURLConnection(URL url) throws IOException {
...
+    
+    static InetSocketAddress getResolvedHttpProxyAddress(String host, int port) throws IOException {
+    	InetSocketAddress targetAddress = null;
+    	Iterator<Proxy> proxies = ProxySelector.getDefault().select(URI.create(String.format("http://%s:%d", host, port))).iterator();
+    	// No break -> The while condition already checks if targetAddress is null.
+    	while (targetAddress == null && proxies.hasNext()) {
+    		Proxy proxy = proxies.next();
+    		if (proxy.type() == Proxy.Type.HTTP) {
+    			if (!(proxy.address() instanceof InetSocketAddress))
+    				throw new IOException("Unsupported proxy address type " + proxy.address().getClass());
```
oleg-nenashev: <code>proxy.address</code> may return null if it is a direct connection. In such case there will be an NPE in the routine. 
If I understand the code correctly, <code> getResolvedHttpProxyAddress</code> should return null in such case to invoke the no-proxy logic in the calling method

src/main/java/hudson/remoting/Util.java
```diff
@@ -131,4 +136,37 @@ static URLConnection openURLConnection(URL url) throws IOException {
...
+    		}
+    	}
+    	if (targetAddress == null) {
+    		String httpProxy = System.getenv("http_proxy");
+    		if (httpProxy != null) {
+    			try {
+    				URL url = new URL(httpProxy);
+    				targetAddress = new InetSocketAddress(url.getHost(), url.getPort());
+    			} catch (MalformedURLException e) {
+    				LOGGER.log(Level.WARNING, INVALID_HTTP_PROXY_MSG + " HTTP Proxy: " + httpProxy);
```
oleg-nenashev: Other calls from remoting or Jenkins slave may try to use this corrupted proxy setting. Probably it's safer to just throw <code>IOException</code> and fail here.

oleg-nenashev:
@evanx9 , @hypnoce 
See my comments above.

I suppose jacoco update is unrelated. For the future, it's better to use separate pull-requests for such updates. If you also resubmit a commit from other developer, please also make sure that you retain the authorship. It may be sensitive to them. If you make a patch on the top of the commit, just keep it as a separate commit.

I also have minor comments regarding the formatting. The project uses a subset of [Oracle code style](http://www.oracle.com/technetwork/java/codeconvtoc-136057.html) 
* The newly added code has tab indenting instead of the space one
* There are pretty much single-line <code>if</code> bodies without brackets. Since TAB indenting differs, it may confuse other developers if the code gets aligned improperly
* Sometimes annotations are helpful for static analysis. If your method may return a null, it's a good practice to annotate it via <code>@CheckForNull</code>

evanx9:
@oleg-nenashev Updated with your suggestions. Code is using spaces now. Noticed a problem with the testExportSecondary Test with Cloudbee's Jenkins. This test passes on my local system. I don't think my changes should have impacted this. Could you re-run the job to confirm?

hypnoce:
Is it possible to keep my commits in the git log ? It does not appear anymore. Thanks.

daniel-beck:
@evanx9 Please create commits that have @hypnoce as author if you're reusing his work.

@oleg-nenashev Time for another review?

daniel-beck:
Retriggering PR build.

evanx9:
@daniel-beck My commit messages clearly site both hypnoce and his original pull request, and is a modification on top of his changes. would that be sufficient?

The new build looks clean. Looking forward to merges with master and potentially the LTS, as we are having issues with the proxy and exclusion list right now with new nodes.

oleg-nenashev:
I'll try to review it tomorrow

hypnoce:
PR #55  has been updated.

oleg-nenashev:
#55 has been merged. I think this PR can be closed

PR #64, opened 2015-11-13
JENKINS-31510 Filter DiskFileItem with ClassFilter.DEFAULT

aldaris:
With this change Jenkins now filters DiskFileItem (in the lights of JENKINS -159). Once a remoting version with this fix is released, Jenkins needs to be updated, and that will also mean that having a custom commons-fileupload release won't be necessary either.


aldaris:
The [test failure](https://jenkins.ci.cloudbees.com/job/libraries/job/remoting/org.jenkins-ci.main$remoting/123/testReport/junit/hudson.remoting/ProxyWriterTest/testRemoteGC_NIO_pipe/)  appears to be unrelated to my changes, also the build works locally.

PR #65, opened 2015-11-24
[FIXED JENKINS-31718] Preserve the BufferedInputStream when building the transport

stephenc:
- Most likely some non-mock based tests would be better and likely would have caught this regression sooner.

See [JENKINS-31718](https://issues.jenkins-ci.org/browse/JENKINS-31718)

@reviewbybees @oleg-nenashev

### DIFF
--- a/src/main/java/hudson/remoting/Capability.java
+++ b/src/main/java/hudson/remoting/Capability.java
@@ -1,15 +1,14 @@
 package hudson.remoting;
 
 import hudson.remoting.Channel.Mode;
-
-import java.io.ObjectStreamClass;
-import java.io.Serializable;
-import java.io.UnsupportedEncodingException;
-import java.io.ObjectOutputStream;
-import java.io.OutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
 
 /**
  * Represents additional features implemented on {@link Channel}.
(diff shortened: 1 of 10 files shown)


oleg-nenashev:
The code looks good to me, but the test legitimately fails

stephenc:
@oleg-nenashev tests fixed

oleg-nenashev:
:bee: the change fixes both reproduction scenarios

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
```diff
@@ -96,9 +97,12 @@ abstract boolean performHandshake(DataOutputStream outputStream,
      *
      * @param socket The established {@link Socket} connection to the master.
      * @param channelBuilder The {@link ChannelBuilder} to use.
+     * @param inputStream The {@link BufferedInputStream} of {@link Socket#getInputStream()} or {@code null} if there
+     *                    is none (TODO remove the null possibility).
      * @return The constructed channel
      */
-    abstract Channel buildChannel(Socket socket, ChannelBuilder channelBuilder) throws IOException;
+    abstract Channel buildChannel(Socket socket, ChannelBuilder channelBuilder,
+                                  BufferedInputStream inputStream) throws IOException;
```
akshayabd: Perhaps it would make more sense to change the method interface to:

abstract Channel buildChannel(ChannelBuilder channelBuilder, BufferedInputStream inputStream, BufferedOutputStream outputStream) throws IOException;

Or just use InputStream and OutputStream in the method signature. That way implementations will not make the mistake of trying to get the inputstream directly from the socket.

jglick:
Follows up https://github.com/jenkinsci/remoting/pull/41 IIUC.

src/main/java/hudson/remoting/Capability.java
```diff
@@ -214,6 +213,69 @@ public static Capability read(InputStream is) throws IOException {
...
+            }
+            sb.append("Proxy writer 2.35");
+        }
+        if ((mask & MASK_CHUNKED_ENCODING) != 0) {
+            if (first) {
+                first = false;
+            } else {
+                sb.append(", ");
+            }
+            sb.append("Chunked encoding");
```
jglick: For this purpose I would just use the code name, e.g., `CHUNKED_ENCODING`.

src/main/java/hudson/remoting/Capability.java
```diff
@@ -214,6 +213,69 @@ public static Capability read(InputStream is) throws IOException {
...
+            } else {
+                sb.append(", ");
+            }
+            sb.append("Proxy writer 2.35");
+        }
+        if ((mask & MASK_CHUNKED_ENCODING) != 0) {
+            if (first) {
+                first = false;
+            } else {
+                sb.append(", ");
```
jglick: Maybe easier to construct a list and then use `StringUtils.join("|")`.

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
jglick: Agreed, better to pass `(ChannelBuilder, InputStream, OutputStream)`.

jglick:
:bug: due to the undesirable API signature.

Would be nice to reproduce the error in a test.

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
kohsuke: +1 on `abstract Channel buildChannel(ChannelBuilder, BufferedInputStream, BufferedOutputStream) throws IOException;` for better symmetry.

jglick:
:bee: on reconsideration since it turns out that although `JnlpProtocol` was `public`, it could not have been instantiated outside the package, so downgrading API signature to an :ant: that may be fixed later.

stephenc:
@reviewbybees done

kohsuke:
This PR was merged after rebasing so that it can ship as 2.53.1.

PR #66, opened 2015-12-17
[FIXED JENKINS-31510]

kohsuke:
Reworked PR #64 based on the tip.


oleg-nenashev:
Retriggering the CI. Seems it does not work due to the remoting layer

oleg-nenashev:
There are many usages in the Jenkins core and plugins:
https://github.com/search?p=5&q=user%3Ajenkinsci+%22org.apache.commons.fileupload%22&ref=searchresults&type=Code&utf8=✓

After a random check I have not found it's propagation over the channel, but seems a full review is required

kohsuke:
Note that we already uses patched commons-fileupload that is not serializable. I think this is a no-brainer change that can be just merged.

oleg-nenashev:
Looking into the commons-fileupload 's code

kohsuke:
I don't remember much about this PR but JENKINS-31510 is already marked as fixed.

PR #67, opened 2015-12-18
[JENKINS-31256] Use credentials in waitForServerToBack

wiizzard:
This patch addresses three regressions introduced with HUDSON-4071, 662b0f and JENKINS-6167 as these only modified run() resulting in no proxy and no authorization support in waitForServerToBack()

As a result the connection to the master could not be reestablished as connections were rejected with response code 401 as reported in JENKINS-31256.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -182,7 +182,7 @@ public class Engine extends Thread {
                     // find out the TCP port
                     HttpURLConnection con = (HttpURLConnection)Util.openURLConnection(salURL);
                     if (credentials != null) {
-                        // TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
+                        // Pass credentials to support running Jenkins behind a (reverse) proxy requiring authorization
                         String encoding = Base64.encode(credentials.getBytes("UTF-8"));
                         con.setRequestProperty("Authorization", "Basic " + encoding);
                     }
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
:+1: . I wonder why remoting supports only Basic auth, but It's another story. Maybe makes sense to move the duplicated credentials handling code to a new helper method.

wiizzard:
Any further comments or can this be merged?

oleg-nenashev:
Uhm, last changes caused a merge conflict. I'm planning a new release soon, so I'll check what can be done with this guy

wiizzard:
I fixed the merge conflict but I think the SSL stuff has also to go to waitForServerToBack(). I think some parts can be commonized into openURLConnection(). I'll take a closer look shortly.

oleg-nenashev:
> I think the SSL stuff has also to go to waitForServerToBack()

+1

oleg-nenashev:
Needs to be merged locally. NVM, I'll do it

nebulans:
Any ETA on this landing?

oleg-nenashev:
Sorry for the delay. I'll try to do it ASAP

src/main/java/hudson/remoting/Engine.java
```diff
@@ -424,6 +393,12 @@ private Socket connect(String host, String port) throws IOException, Interrupted
     private void waitForServerToBack() throws InterruptedException {
         Thread t = Thread.currentThread();
         String oldName = t.getName();
+        SSLSocketFactory sslSocketFactory = null;
+        try {
+            sslSocketFactory = getSSLSocketFactory();
+        } catch (Throwable e) {
```
oleg-nenashev: This approach is being used in other places of this file. It is not LGTM, but it does not increase the technical debt

oleg-nenashev:
Merged via #87

PR #68, opened 2016-01-29
Allow user to adjust socket timeout

fengxx:
we encounter slave offline issue when executing commands like 
`
sleep 40m
`
on jnlp agent, the socket read timeout is hard coded to 30 minutes and caused the issue

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -388,7 +388,7 @@ public class Engine extends Thread {
                 // to interfere with normal operation. the main purpose of this is that when the other peer dies
                 // abruptly, we shouldn't hang forever, and at some point we should notice that the connection
                 // is gone.
-                s.setSoTimeout(30*60*1000); // 30 mins. See PingThread for the ping interval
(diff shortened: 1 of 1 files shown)


fengxx:
the build was aborted for exceeding build time limit (8 minutes), maybe caused by slow maven download. I don't think the commit will affects the build

src/main/java/hudson/remoting/Engine.java
```diff
@@ -403,6 +403,8 @@ public static Engine current() {
 
     private static final Logger LOGGER = Logger.getLogger(Engine.class.getName());
 
+    static final int TIMEOUT = Integer.getInteger(Engine.class.getName()+".soTimeout",30*60*1000);
```
oleg-nenashev: Maybe "socketTimeout"? Readability is better than economy on several symbols :)

oleg-nenashev:
merge conflict

oleg-nenashev:
I want to make the timeout configurable before the next version release

fengxx:
Thanks for reviewing, I had renamed it to socketTimeout

oleg-nenashev:
Subject for a backporting. I'll create an issue

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-34808

PR #69, opened 2016-02-02
Activate JNLP3 protocol

kohsuke:
JNLP3 protocol support is live now, with some more massaging to move some of the server-side code into the remoting for better pairing.

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol3.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol3.java
@@ -103,6 +103,7 @@ import static org.jenkinsci.remoting.engine.Jnlp3Util.*;
  * 256bit sizes are supported.
  *
  * @author Akshay Dayal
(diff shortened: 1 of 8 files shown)


kohsuke:
cc @akshayabd @oleg-nenashev @jglick

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol3.java
```diff
@@ -37,7 +37,7 @@
 import java.util.Properties;
 
 import static org.jenkinsci.remoting.engine.EngineUtil.*;
-import static org.jenkinsci.remoting.engine.Jnlp3Util.createChallengeResponse;
+import static org.jenkinsci.remoting.engine.Jnlp3Util.*;
```
amuniz: We should avoid * imports, unless on test code.

kohsuke:
Turns out I still got some issues with this code. Please standby

kohsuke:
With the updated push this change should br good to go.

PR #70, opened 2016-02-04
Test failures on Java 8

jglick:
I have found that while tests reliably pass for me on Java 7, when using Java 8 I often get some failures about the “south” channel not being closed. Not sure why, exactly, but I suspect a race condition—`CloseCommand` not having been processed by the time the assertion is run?

@reviewbybees esp. @kohsuke, @oleg-nenashev


jglick:
And the CI builder appears to be using Java 7.

oleg-nenashev:
:bee:

PR #71, opened 2016-02-07
[JENKINS-32825] Deadlock in Channel Abort

andresrc:
See [JENKINS-32925](https://issues.jenkins-ci.org/browse/JENKINS-32825)

Additional `closed` state check introduced to break writing loop.

@reviewbybees


oleg-nenashev:
The change looks good to me, :bee: and :+1:
The failure has been caused by the build timeout on CI. Retriggering it

stephenc:
:bee:

andresrc:
@reviewbybees done

oleg-nenashev:
here we go. It's not a candidate for backporting into 2.53.3 as I see (cc @kohsuke and @stephenc )

PR #72, opened 2016-02-16
[FIXED JENKINS-32980] Adds a new CLI option to specify additional X.509 certs

stephenc:
PEM encoded certificates to trust when performing JNLP port discovery on the supplied Jenkins URLs

@jenkinsci/code-reviewers @reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -24,6 +24,26 @@
 package hudson.remoting;
 
 import hudson.remoting.Channel.Mode;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.security.AccessController;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManagerFactory;
 import org.jenkinsci.remoting.engine.JnlpProtocol;
 import org.jenkinsci.remoting.engine.JnlpProtocolFactory;
 
@@ -83,6 +103,11 @@ public class Engine extends Thread {
      * "http://foo.bar/jenkins/".
      */
     private List<URL> candidateUrls;
+    /**
+     * The list of {@link X509Certificate} instances to trust when connecting to any of the {@link #candidateUrls}
+     * or {@code null} to use the JVM default trust store.
+     */
+    private List<X509Certificate> candidateCertificates;
 
     /**
      * URL that points to Jenkins's tcp slave agent listener, like <tt>http://myhost/hudson/</tt>
@@ -145,6 +170,19 @@ public class Engine extends Thread {
         this.noReconnect = noReconnect;
     }
 
+    public void setCandidateCertificates(List<X509Certificate> candidateCertificates) {
+        this.candidateCertificates = candidateCertificates == null
+                ? null
+                : new ArrayList<X509Certificate>(candidateCertificates);
+    }
+
+    public void addCandidateCertificate(X509Certificate certificate) {
+        if (candidateCertificates == null) {
+            candidateCertificates = new ArrayList<X509Certificate>();
+        }
+        candidateCertificates.add(certificate);
+    }
+
     public void addListener(EngineListener el) {
         events.add(el);
     }
@@ -173,6 +211,25 @@ public class Engine extends Thread {
                 Throwable firstError=null;
                 String host=null;
                 String port=null;
+                SSLSocketFactory sslSocketFactory = null;
+                if (candidateCertificates != null && !candidateCertificates.isEmpty()) {
+                    KeyStore keyStore = getCacertsKeyStore();
+                    // load the keystore
+                    keyStore.load(null, null);
+                    int i = 0;
+                    for (X509Certificate c : candidateCertificates) {
+                        keyStore.setCertificateEntry(String.format("alias-%d", i++), c);
+                    }
+                    // prepare the trust manager
+                    TrustManagerFactory trustManagerFactory =
+                            TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+                    trustManagerFactory.init(keyStore);
+                    // prepare the SSL context
+                    SSLContext ctx = SSLContext.getInstance("TLS");
+                    ctx.init(null, trustManagerFactory.getTrustManagers(), null);
+                    // now we have our custom socket factory
+                    sslSocketFactory = ctx.getSocketFactory();
+                }
 
                 for (URL url : candidateUrls) {
                     String s = url.toExternalForm();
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -145,6 +170,19 @@ public void setNoReconnect(boolean noReconnect) {
         this.noReconnect = noReconnect;
     }
 
+    public void setCandidateCertificates(List<X509Certificate> candidateCertificates) {
+        this.candidateCertificates = candidateCertificates == null
+                ? null
+                : new ArrayList<X509Certificate>(candidateCertificates);
+    }
+
+    public void addCandidateCertificate(X509Certificate certificate) {
```
oleg-nenashev: :bug: Since the stuff is public, all methods must be synchronized

src/main/java/hudson/remoting/Engine.java
```diff
@@ -403,6 +463,101 @@ public static Engine current() {
...
+                File trustStoreFile;
+                if (trustStore != null) {
+                    trustStoreFile = new File(trustStore);
+                    trustStoreStream = getFileInputStream(trustStoreFile);
+                } else {
+                    String javaHome = properties.get("javaHome");
+                    trustStoreFile = new File(
+                            javaHome + File.separator + "lib" + File.separator + "security" + File.separator
+                                    + "jssecacerts");
+                    if ((trustStoreStream = getFileInputStream(trustStoreFile)) == null) {
```
oleg-nenashev: From what I see in Java documentation, jssecacerts and cacerts have no priority specifications. BTW here we ignore "cacerts" if we can get something from jssecacerts. Is it correct?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -403,6 +463,101 @@ public static Engine current() {
...
+            String trustStoreType = properties.get("trustStoreType");
+            String trustStoreProvider = properties.get("trustStoreProvider");
+            LOGGER.log(Level.FINE, "trustStore is: {0}", trustStore);
+            LOGGER.log(Level.FINE, "trustStore type is: {0}", trustStoreType);
+            LOGGER.log(Level.FINE, "trustStore provider is: {0}", trustStoreProvider);
+
+            if (trustStoreType.length() != 0) {
+                LOGGER.log(Level.FINE, "init truststore");
+
+                if (trustStoreProvider.length() == 0) {
```
oleg-nenashev: better to use StringUtils.isEmpty(). Just in case somebody changes the implementation above

src/main/java/hudson/remoting/Engine.java
```diff
@@ -403,6 +463,101 @@ public static Engine current() {
...
+
+                char[] trustStorePasswdChars = null;
+                String trustStorePasswd = properties.get("trustStorePasswd");
+                if (trustStorePasswd.length() != 0) {
+                    trustStorePasswdChars = trustStorePasswd.toCharArray();
+                }
+
+                keystore.load(trustStoreStream, trustStorePasswdChars);
+                if (trustStorePasswdChars != null) {
+                    for (int i = 0; i < trustStorePasswdChars.length; ++i) {
```
oleg-nenashev: Why do you need to zero items here?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -403,6 +463,101 @@ public static Engine current() {
...
+        } finally {
+            if (trustStoreStream != null) {
+                trustStoreStream.close();
+            }
+        }
+
+        return keystore;
+    }
+
+    private static FileInputStream getFileInputStream(final File file) throws PrivilegedActionException {
```
oleg-nenashev: :ant: CheckForNull

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -159,6 +172,68 @@ public Engine createEngine() {
...
+                            && length > "-----BEGIN CERTIFICATE-----\n-----END CERTIFICATE-----".length()) {
+                        FileInputStream fis = null;
+                        try {
+                            // we do basic size validation, if there are x509 certificates that have a PEM encoding
+                            // larger
+                            // than 64kb we can revisit the upper bound.
+                            cert = new byte[(int) length];
+                            fis = new FileInputStream(file);
+                            int read = fis.read(cert);
+                            if (cert.length == read) {
```
oleg-nenashev: :bug: <code>!=</code>

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -159,6 +172,68 @@ public Engine createEngine() {
...
+            try {
+                factory = CertificateFactory.getInstance("X.509");
+            } catch (CertificateException e) {
+                throw new IllegalStateException("Java platform specification mandates support for X.509", e);
+            }
+            List<X509Certificate> certificates = new ArrayList<X509Certificate>(candidateCertificates.size());
+            for (String certOrAtFilename : candidateCertificates) {
+                certOrAtFilename = certOrAtFilename.trim();
+                byte[] cert;
+                if (certOrAtFilename.startsWith("@")) {
```
oleg-nenashev: In the case of files it would be great to require 700 permissions on Unix. Just to ensure that certificate is not leakable

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -159,6 +172,68 @@ public Engine createEngine() {
...
+                        cert = certOrAtFilename.getBytes("US-ASCII");
+                    } catch (UnsupportedEncodingException e) {
+                        LOGGER.log(Level.WARNING, "Could not parse certificate " + certOrAtFilename, e);
+                        continue;
+                    }
+                }
+                try {
+                    certificates.add((X509Certificate) factory.generateCertificate(new ByteArrayInputStream(cert)));
+                } catch (ClassCastException e) {
+                    LOGGER.log(Level.WARNING, "Expected X.509 certificate from " + certOrAtFilename, e);
```
oleg-nenashev: Do we want to continue the cycle after it?

oleg-nenashev:
:bug: for the synchronization and erroneous CERT file comparison

src/main/java/hudson/remoting/Engine.java
stephenc: None of the other class settings are sync, so this class is not thread-safe

oleg-nenashev: I'm not convinced, but I agree to decrease the priority to :whale:

src/main/java/hudson/remoting/Engine.java
stephenc: Unspecified in the JVM spec but the order used by open JDK is that [jssecacerts is priority over cacerts](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/sun/security/ssl/TrustManagerFactoryImpl.java#L174)

src/main/java/hudson/remoting/Engine.java
stephenc: This is standard practice whenever you have a password in a `char[]`, you zero out the items after use

src/main/java/hudson/remoting/jnlp/Main.java
stephenc: why? anyone can get the cert by opening a HTTPS connection to the URL

src/main/java/hudson/remoting/jnlp/Main.java
stephenc: yes

src/main/java/hudson/remoting/Engine.java
stephenc: I added a `@NotThreadSafe` annotation... if you want to make it thread safe, that is a different change

src/main/java/hudson/remoting/Engine.java
oleg-nenashev: Agreed

src/main/java/hudson/remoting/Engine.java
oleg-nenashev: I agree, it's reasonable from the security perspective (at least if JIT is not smart enough to optimize it)

oleg-nenashev:
retriggering the CI

oleg-nenashev:
:bee:

oleg-nenashev:
Actually @reviewbybees done

stephenc:
@oleg-nenashev as maintainer who should merge?

PR #73, opened 2016-02-17
callAsync().get() does not return when the channel is closed

stephenc:
@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Request.java
+++ b/src/main/java/hudson/remoting/Request.java
@@ -238,7 +238,10 @@ abstract class Request<RSP extends Serializable,EXC extends Throwable> extends C
                             if (isCancelled()) {
                                 throw new CancellationException();
                             }
-                            Request.this.wait(); // wait until the response arrives
+                            if (channel.isInClosed()) {
+                                throw new ExecutionException(new RequestAbortedException(null));
+                            }
+                            Request.this.wait(30*1000); // wait until the response arrives
                         }
                     } catch (InterruptedException e) {
                         try {
@@ -266,7 +269,10 @@ abstract class Request<RSP extends Serializable,EXC extends Throwable> extends C
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now)));
+                        if (channel.isInClosed()) {
+                            throw new ExecutionException(new RequestAbortedException(null));
+                        }
+                        Request.this.wait(Math.min(30*1000,Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now))));
                         now = System.nanoTime();
                     }
                     if (response == null)
--- a/src/test/java/hudson/remoting/ForkEBCDICRunner.java
+++ b/src/test/java/hudson/remoting/ForkEBCDICRunner.java
@@ -12,4 +12,9 @@ public class ForkEBCDICRunner extends ForkRunner {
         r.add(0,"-Dfile.encoding=CP037");
         return r;
     }
+
+    @Override
+    public String getName() {
+        return "forkEBCDIC";
+    }
 }


src/main/java/hudson/remoting/Request.java
```diff
@@ -234,11 +234,14 @@ public boolean isDone() {
             public RSP get() throws InterruptedException, ExecutionException {
                 synchronized(Request.this) {
                     try {
-                        while(response==null) {
+                        while(response==null && !channel.isInClosed()) {
                             if (isCancelled()) {
                                 throw new CancellationException();
                             }
-                            Request.this.wait(); // wait until the response arrives
+                            Request.this.wait(30*1000); // wait until the response arrives
```
oleg-nenashev: Waits w/o timeout are evil, but such hardcoded timeout is not applicable in the case of network-robust instances

oleg-nenashev:
:bug: hardcoded timeout is too small IMHO. It should be configurable in any case

src/main/java/hudson/remoting/Request.java
stephenc: This is the exact same timeout for recheck the while condition as is used in `call()` hence using that timeout.

Spin-looping once every 30 seconds is acceptable to detect channel closed conditions. Longer timeouts do not seem to make sense, and shorter ones will just generate more loop. If you want to parameterize then all 3 uses of `30*1000` should be parameterized... i.e. a separate change from this

stephenc:
Timeout being configurable is a separate issue. The timeout is exactly [the same as used in the non-async `call()`](https://github.com/jenkinsci/remoting/pull/73/files#diff-c08fa2c44dfadb8726415c8e03fc4850R147)

src/main/java/hudson/remoting/Request.java
jtnord: @oleg-nenashev I would go the opposite of you and say "how can the limit be too small - its 30 seconds and is in a loop"?
I would hate to block to 5 seconds if the channel is being closed down - for me it is way too big :-)
calling `isCancelled()` and `channel.IsInClosed()` should be light enough that we can call them every second!

Anyway - 30 seconds is better than ∞

src/main/java/hudson/remoting/Request.java
```diff
@@ -262,13 +265,16 @@ public RSP get(long timeout, TimeUnit unit) throws InterruptedException, Executi
...
-                    while (response == null && (end - now > 0L)) {
+                    while (response == null && (end - now > 0L) && !channel.isInClosed()) {
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now)));
+                        Request.this.wait(Math.min(30*1000,Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now))));
                         now = System.nanoTime();
                     }
+                    if (response == null && end > now) {
```
jtnord: nanoTime Overflow end = LONG.MAX_LONG  now = LONG.MIN_LONG....

jtnord:
:bug: on nanoTime comparison using >

src/main/java/hudson/remoting/Request.java
jtnord: this comparison makes me twitchy - but I can;t put my finger on it....

jtnord:
:bug: on incorrect comparison of the nanoTime (now may have looped over and be very small - yet now may not have done)

stephenc:
@jtnord it's a fair cop on the `nanotime` comparison

jtnord:
:bee: but there the unit test failures smell like they are related.

stephenc:
@jtnord we'll see how my latest changes fair, they work locally

stephenc:
@jtnord in fact on be7d19f the tests pass locally for me also

stephenc:
@jtnord got to the bottom of the failures. Basically the ping thread was detecting the channel as closed first, declaring an error and calling `System.exit(-1)` which gives 255 not 0.

stephenc:
I'm going to squash the commits if https://github.com/jenkinsci/remoting/commit/1811a9b01d626de9447a419b4fe30a9d50a576eb passes

stephenc:
OK, so actually it was the exception I was throwing... counter-intuitively it is `RequestAbortedException` that you are supposed to throw not `IOException("Channel closed")` as that way `PingThread` will see it as normal termination and not exit with `-1`

stephenc:
@jtnord OK we have a pass in the test run :-) care to re-review

src/main/java/hudson/remoting/Request.java
```diff
@@ -238,7 +238,10 @@ public RSP get() throws InterruptedException, ExecutionException {
                             if (isCancelled()) {
                                 throw new CancellationException();
                             }
-                            Request.this.wait(); // wait until the response arrives
+                            if (channel.isInClosed()) {
+                                throw new ExecutionException(new RequestAbortedException(null));
```
jtnord: what no message "Channel was closed whilst waiting for a response"

src/main/java/hudson/remoting/Request.java
```diff
@@ -266,7 +269,10 @@ public RSP get(long timeout, TimeUnit unit) throws InterruptedException, Executi
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now)));
+                        if (channel.isInClosed()) {
+                            throw new ExecutionException(new RequestAbortedException(null));
```
jtnord: what no message "Channel was closed whilst waiting for a response"

src/main/java/hudson/remoting/Request.java
stephenc: Nope that's what this exception pattern means (see non-async call method above)

:sadpanda:

jtnord: oh - it takes a thorwable and no String message.
:cry:

src/test/java/hudson/remoting/ForkEBCDICRunner.java
```diff
@@ -12,4 +12,9 @@
         r.add(0,"-Dfile.encoding=CP037");
         return r;
     }
+
+    @Override
+    public String getName() {
```
jtnord: what's this in aid of?

jtnord:
:bee:

src/main/java/hudson/remoting/Request.java
```diff
@@ -266,7 +269,10 @@ public RSP get(long timeout, TimeUnit unit) throws InterruptedException, Executi
                         if (isCancelled()) {
                             throw new CancellationException();
                         }
-                        Request.this.wait(Math.max(1, TimeUnit.NANOSECONDS.toMillis(end - now)));
```
oleg-nenashev: Oh, it was not good

oleg-nenashev:
:bee: 
@reviewbybees done

src/test/java/hudson/remoting/ForkEBCDICRunner.java
stephenc: Oh look I have [two failed tests](https://jenkins.ci.cloudbees.com/job/libraries/job/remoting/147/testReport/)... and they have the same name... how on earth am I supposed to determine which is the one at fault... is it the `-fork` from `ForkRunner` or is it the `-fork` from `ForkEBCDICRunner`...

with this change you will worry no more 

(And when I was able to reproduce that failure locally it would only be for one of the forks at a time... because I have more cores and the threads were usually woken in the correct order to let the tests pass)

stephenc:
@oleg-nenashev right so as maintainer of this repo, do you want to merge or shall I?

PR #74, opened 2016-02-22
[WiP][FIXED JENKINS-33082] Record diagnosis information under a slave disconnection failure

fbelzunc:
This is just a work in progress...

For diagnosis purposes could be nice to record what is happening on the slave side once the pingThread has failed.

@andresrc @oleg-nenashev @jtnord WDYT about this? Might be useful?

### DIFF
--- a/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
+++ b/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
@@ -76,7 +76,9 @@ public class AgentFailuresAnalyzer {
         } catch (UnsupportedEncodingException e) {
             throw e;
         } finally {
-            writer.close();
+            if (writer != null) {
+                writer.close();
+            }
         }
         if (cause != null) {
             cause.printStackTrace(writer);
--- a/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
+++ b/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
@@ -27,12 +27,12 @@ public class AgentFailuresAnalyzer {
     /**
      * Channel we are going to study
      */
-    private Channel c;
+    private final Channel c;
 
     /**
      * The cause of the ping failure
      */
-    private Throwable cause;
+    private final Throwable cause;
 
 
     /**
@@ -71,6 +71,9 @@ public class AgentFailuresAnalyzer {
         PrintWriter writer = null;
         try {
             writer = new PrintWriter(file, "UTF-8");
+            if (cause != null) {
+                cause.printStackTrace(writer);
+            }
         } catch (FileNotFoundException e) {
             throw e;
         } catch (UnsupportedEncodingException e) {
@@ -79,9 +82,6 @@ public class AgentFailuresAnalyzer {
             if (writer != null) {
                 writer.close();
             }
-        }
-        if (cause != null) {
-            cause.printStackTrace(writer);
         }
     }
 
@@ -101,13 +101,7 @@ public class AgentFailuresAnalyzer {
         threadDumps.add(file);
 
         PrintWriter writer;
-        try {
-            writer = new PrintWriter(file, "UTF-8");
-        } catch (FileNotFoundException e) {
-            throw e;
-        } catch (UnsupportedEncodingException e) {
-            throw e;
-        }
+        writer = new PrintWriter(file, "UTF-8");
 
         ThreadMXBean mbean = ManagementFactory.getThreadMXBean();
         ThreadInfo[] threads;
@@ -126,23 +120,23 @@ public class AgentFailuresAnalyzer {
         long[] deadLocks;
         try {
             deadLocks = mbean.findDeadlockedThreads();
-        } catch (UnsupportedOperationException x) {
-            x.printStackTrace(writer);
-            deadLocks = null;
-        }
-        if (deadLocks != null && deadLocks.length != 0) {
-            writer.println(" Deadlock Found ");
-            ThreadInfo[] deadLockThreads = mbean.getThreadInfo(deadLocks);
-            for (ThreadInfo threadInfo : deadLockThreads) {
-                StackTraceElement[] elements = threadInfo.getStackTrace();
-                for (StackTraceElement element : elements) {
-                    writer.println(element.toString());
+            if (deadLocks != null && deadLocks.length != 0) {
+                writer.println(" Deadlock Found ");
+                ThreadInfo[] deadLockThreads = mbean.getThreadInfo(deadLocks);
+                for (ThreadInfo threadInfo : deadLockThreads) {
+                    StackTraceElement[] elements = threadInfo.getStackTrace();
+                    for (StackTraceElement element : elements) {
+                        writer.println(element.toString());
+                    }
                 }
             }
+        } catch (UnsupportedOperationException e) {
+            e.printStackTrace(writer);
+        } finally {
+            writer.println();
+            writer.flush();
+            writer.close();
         }
-        writer.println();
-        writer.flush();
-        writer.close();
     }
 
     /**
--- a/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
+++ b/src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
@@ -18,11 +18,11 @@ import java.util.Date;
  * Generates diagnosis information (threaddump + stacktrace) when the pingThread fails
  *
  * <p>
- * Useful to troubleshoot the slave disconnection failures on the agent side.
+ * Useful to troubleshoot agent disconnection failures
  *
  * @since 2.55
  */
-public class PingFailureAnalyzer {
+public class AgentFailuresAnalyzer {
 
     /**
      * Channel we are going to study
@@ -36,7 +36,7 @@ public class PingFailureAnalyzer {
 
 
     /**
-     * Maximum of files to store inside pingFailureAnalyzer/thread-dumps and pingFailureAnalyzer/stacktraces.
+     * Maximum of files to store inside agentFailuresAnalyzer/thread-dumps and pingFailureAnalyzer/stacktraces.
      * When the limit is reached out it starts overriding.
      */
     private final static int MAX_DIAGNOSIS_FILES = 5;
@@ -44,40 +44,43 @@ public class PingFailureAnalyzer {
     /**
      * Folder to store the thread dumps. Maintains most recent N files in a directory in cooperation with the writer.
      */
-    private final PingThreadFileListCap threadDumps = new PingThreadFileListCap(new File("pingFailureAnalyzer/thread-dumps"), MAX_DIAGNOSIS_FILES);
+    private final AgentFailuresFileListCap threadDumps = new AgentFailuresFileListCap(new File("agentFailuresAnalyzer/thread-dumps"), MAX_DIAGNOSIS_FILES);
 
     /**
      * Folder to store the stacktraces. Maintains most recent N files in a directory in cooperation with the writer.
      */
-    private final PingThreadFileListCap stacktraces = new PingThreadFileListCap(new File("pingFailureAnalyzer/stacktraces"), MAX_DIAGNOSIS_FILES);
+    private final AgentFailuresFileListCap stacktraces = new AgentFailuresFileListCap(new File("agentFailuresAnalyzer/stacktraces"), MAX_DIAGNOSIS_FILES);
 
-    public PingFailureAnalyzer(Channel c, Throwable cause) {
+    public AgentFailuresAnalyzer(Channel c, Throwable cause) {
         this.c = c;
         this.cause = cause;
     }
 
 
     /**
-     * Save the stacktrace which produces the ping failure into pingFailureAnalyzer/stacktraces
+     * Save the stacktrace which produces the ping failure into agentFailuresAnalyzer/stacktraces
      *
      * @param cause
      *      The Throwable
      */
-    public void saveStackTrace(Throwable cause) {
+    public void saveStackTrace(Throwable cause) throws FileNotFoundException, UnsupportedEncodingException {
         long iota = System.currentTimeMillis();
         final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
-        File file = stacktraces.file(format.format(new Date(iota++)) + ".txt");
+        File file = stacktraces.file("stacktrace-" + format.format(new Date(iota)) + ".txt");
         stacktraces.add(file);
         PrintWriter writer = null;
         try {
             writer = new PrintWriter(file, "UTF-8");
         } catch (FileNotFoundException e) {
-            e.printStackTrace();
+            throw e;
         } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+            throw e;
+        } finally {
+            writer.close();
         }
-        cause.printStackTrace(writer);
-        writer.close();
+        if (cause != null) {
+            cause.printStackTrace(writer);
+        }
     }
 
     /**
@@ -89,19 +92,19 @@ public class PingFailureAnalyzer {
             value = {"VA_FORMAT_STRING_USES_NEWLINE"},
             justification = "We don't want platform specific"
     )
-    public void takeThreadDump() {
+    public void takeThreadDump() throws FileNotFoundException, UnsupportedEncodingException {
         long iota = System.currentTimeMillis();
         final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
-        File file = threadDumps.file(format.format(new Date(iota++)) + ".txt");
+        File file = threadDumps.file(format.format("threadDump-" + new Date(iota)) + ".txt");
         threadDumps.add(file);
 
-        PrintWriter writer = null;
+        PrintWriter writer;
         try {
             writer = new PrintWriter(file, "UTF-8");
         } catch (FileNotFoundException e) {
-            e.printStackTrace();
+            throw e;
         } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+            throw e;
         }
 
         ThreadMXBean mbean = ManagementFactory.getThreadMXBean();
@@ -135,9 +138,9 @@ public class PingFailureAnalyzer {
                 }
             }
         }
-
         writer.println();
         writer.flush();
+        writer.close();
     }
 
     /**
@@ -152,21 +155,11 @@ public class PingFailureAnalyzer {
             justification = "We don't want platform specific"
     )
     public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBean mbean) {
-        long cpuPercentage;
-        try {
-            long cpuTime = mbean.getThreadCpuTime(t.getThreadId());
-            long threadUserTime = mbean.getThreadUserTime(t.getThreadId());
-            cpuPercentage = (cpuTime == 0) ? 0: 100 * threadUserTime / cpuTime;
-        } catch (UnsupportedOperationException x) {
-            x.printStackTrace(writer);
-            cpuPercentage = 0;
-        }
-        writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%%",
+        writer.printf("\"%s\" id=%d (0x%x) state=%s",
                 t.getThreadName(),
                 t.getThreadId(),
                 t.getThreadId(),
-                t.getThreadState(),
-                cpuPercentage);
+                t.getThreadState());
         final LockInfo lock = t.getLockInfo();
         if (lock != null && t.getThreadState() != Thread.State.BLOCKED) {
             writer.printf("\n    - waiting on <0x%08x> (a %s)",
--- a/src/main/java/hudson/remoting/AgentFailuresFileListCap.java
+++ b/src/main/java/hudson/remoting/AgentFailuresFileListCap.java
@@ -31,8 +31,9 @@ public class AgentFailuresFileListCap {
         this.folder = folder;
         this.size = size;
 
-        if (!folder.exists() && !folder.mkdirs())
-            throw new Error("Failed to create "+folder);
+        if (!folder.exists() && !folder.mkdirs()) {
+            throw new Error("Failed to create " + folder);
+        }
 
         File[] sortedFiles = folder.listFiles(filter);
         if (sortedFiles==null)  return;
@@ -50,10 +51,6 @@ public class AgentFailuresFileListCap {
         files.addAll(Arrays.asList(sortedFiles));
     }
 
-    public File getFolder() {
-        return folder;
-    }
-
     public synchronized void add(File f) {
         if (size <= files.size()) {
             Iterator<File> itr = files.iterator();
@@ -67,11 +64,6 @@ public class AgentFailuresFileListCap {
         }
 
         files.add(f);
-    }
-
-    public synchronized void touch(File f){
-        files.remove(f);
-        add(f);
     }
 
     /**
--- a/src/main/java/hudson/remoting/AgentFailuresFileListCap.java
+++ b/src/main/java/hudson/remoting/AgentFailuresFileListCap.java
@@ -15,19 +15,19 @@ import java.util.logging.Logger;
  * Useful to record incidents as files without bankrupting
  *
  */
-public class PingThreadFileListCap {
+public class AgentFailuresFileListCap {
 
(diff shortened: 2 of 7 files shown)


src/main/java/hudson/remoting/Launcher.java
```diff
@@ -516,19 +516,34 @@ public static void main(InputStream is, OutputStream os, Mode mode, boolean perf
...
 
-        Channel channel = cb.build(is, os);
+        final Channel channel = cb.build(is, os);
         System.err.println("channel started");
         long timeout = 1000 * Long.parseLong(
                 System.getProperty("hudson.remoting.Launcher.pingTimeoutSec", "240")),
              interval = 1000 * Long.parseLong(
                 System.getProperty("hudson.remoting.Launcher.pingIntervalSec", "600"));
+
+        boolean pingFailureAnalyzer = Boolean.parseBoolean(System.getProperty("hudson.remoting.Launcher.pingFailureAnalyzer", "false"));
```
jtnord: [`Boolean.getBoolean("hudson.remoting.Launcher.pingFailureAnalyzer")`](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String))

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -516,19 +516,34 @@ public static void main(InputStream is, OutputStream os, Mode mode, boolean perf
...
+
         if (performPing && timeout > 0 && interval > 0) {
-            new PingThread(channel, timeout, interval) {
+            new PingThread(channel, timeout, interval, pingFailureAnalyzer) {
```
jtnord: there are more ways to create/construct the PingThread than just here - why not put this in the base class - or in a class that the PingThread can call?

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * Generates diagnosis information (threaddump + stacktrace) when the pingThread fails
+ *
+ * <p>
+ * Useful to troubleshoot the slave disconnection failures on the agent side.
```
jtnord: should be on both sides - who is to say the fault is not on the masters side?

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Save the stacktrace which produces the ping failure into pingFailureAnalyzer/stacktraces
+     *
+     * @param cause
+     *      The Throwable
+     */
+    public void saveStackTrace(Throwable cause) {
+        long iota = System.currentTimeMillis();
+        final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
+        File file = stacktraces.file(format.format(new Date(iota++)) + ".txt");
```
jtnord: what is this iota++?

src/main/java/hudson/remoting/PingThreadFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+
+        files.addAll(Arrays.asList(sortedFiles));
+    }
+
+    public File getFolder() {
+        return folder;
+    }
+
+    public synchronized void add(File f) {
+        if (size <= files.size()) {
```
jtnord: why the check?

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -61,14 +61,24 @@
      */
     private final long interval;
 
-    public PingThread(Channel channel, long timeout, long interval) {
+    /**
+     * Specifis if troubleshooting information should be catched in case of pingFailure
+     */
+    private final boolean pingFailureAnalyzer;
+
+    public PingThread(Channel channel, long timeout, long interval, boolean pingFailureAnalyzer) {
```
jtnord: don't we want this on by default?

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+            value = {"VA_FORMAT_STRING_USES_NEWLINE"},
+            justification = "We don't want platform specific"
+    )
+    public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBean mbean) {
+        long cpuPercentage;
+        try {
+            long cpuTime = mbean.getThreadCpuTime(t.getThreadId());
+            long threadUserTime = mbean.getThreadUserTime(t.getThreadId());
+            cpuPercentage = (cpuTime == 0) ? 0: 100 * threadUserTime / cpuTime;
```
jtnord: this is a bug copied from the support core plugin - please don't duplicate it :-)

100% cpu time is for things that have been running without stopping since Jenkins started.
a thread that wakes up every day and runs for 1ms should not be listed at 100%

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+            for (ThreadInfo threadInfo : deadLockThreads) {
+                StackTraceElement[] elements = threadInfo.getStackTrace();
+                for (StackTraceElement element : elements) {
+                    writer.println(element.toString());
+                }
+            }
+        }
+
+        writer.println();
+        writer.flush();
```
jtnord: are you are going to close this writer somewhere ;)

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -516,19 +516,34 @@ public static void main(InputStream is, OutputStream os, Mode mode, boolean perf
...
+                @Deprecated
                 @Override
                 protected void onDead() {
                     System.err.println("Ping failed. Terminating");
                     System.exit(-1);
                 }
+
+                @Override
+                protected void onDead(Throwable diagnosis) {
+                    if (this.isPingFailureAnalyzer()) {
```
jtnord: Will this will spam the logs for orderly shutdown?

https://github.com/jenkinsci/remoting/pull/74/files#diff-ba801c8ba3eabb55537cfff979bb3655L117

src/main/java/hudson/remoting/PingFailureAnalyzer.java
andresrc: On server-side there is kind of an extension for that. Will have to check,

src/main/java/hudson/remoting/PingFailureAnalyzer.java
andresrc: I'd better use the date to identify the case and a different suffix for each diagnosis (stack trace, thread dump, etc), simpler to group the related files.

src/main/java/hudson/remoting/PingFailureAnalyzer.java
andresrc: +1

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+        PrintWriter writer = null;
+        try {
+            writer = new PrintWriter(file, "UTF-8");
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
+        cause.printStackTrace(writer);
+        writer.close();
```
andresrc: `close` in a `finally`

andresrc: And possible NPE.

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+
+        PrintWriter writer = null;
+        try {
+            writer = new PrintWriter(file, "UTF-8");
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
+
```
andresrc: `writer` can be `null` past this point.

src/main/java/hudson/remoting/PingThread.java
andresrc: Depending on whether we want to change current default behavior.

andresrc:
Besides line comments I would like to check the master-side use of the PingThread.

oleg-nenashev:
@fbelzunc Any plans to finish it?

fbelzunc:
@oleg-nenashev Tomorrow, I promise to go through all the problems pointed by everybody.

fbelzunc:
Taking a look at this now.

src/main/java/hudson/remoting/PingFailureAnalyzer.java
fbelzunc: @andresrc I think you are talking about http://javadoc.jenkins-ci.org/jenkins/slaves/PingFailureAnalyzer.html and I am already using it on my code. So I guess, this might be useful for master troubleshooting as well.

src/main/java/hudson/remoting/PingThread.java
fbelzunc: I would prefer to have it disabled in order ensure any bad effect of this PR on the remoting behaviour. In any case, I am happy to do whatever you think might be the best option here.

src/main/java/hudson/remoting/PingThreadFileListCap.java
fbelzunc: Because we only want to generate a fix number of troubleshooting files so it doesn't get bigger and bigger. It is just rotating to avoid size issues.

src/main/java/hudson/remoting/Launcher.java
fbelzunc: I don't understand this concern.

* Can you elaborate a bit more, please?

jtnord: if you shutdown the launcher (with a ctrl-c, service stop etc) - I assume the channel will be closed and `onDead` will be called back.
If this is true then this seems to causing excessive logging.

src/main/java/hudson/remoting/PingThreadFileListCap.java
jtnord: so you should be checking if you have too many files then not too few? `if (size >= files.size()) {...`

fbelzunc:
@andresrc @jtnord @oleg-nenashev What do you think now? I think I simplified the code.

I would like to fully test the feature, but it would be nice if I can have your insights about this proposition.

Many thanks!

src/main/java/hudson/remoting/PingThreadFileListCap.java
fbelzunc: No, because you specify a limit with `size`, so if you do `if (size >= files.size())` then you will be deleting from the beginning - so any file will be actually created.

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+        } finally {
+            if (writer != null) {
+                writer.close();
+            }
+        }
+        if (cause != null) {
+            cause.printStackTrace(writer);
+        }
+    }
+
```
andresrc: :bug: method is wrong, the content is written after closing the writer.

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+                for (StackTraceElement element : elements) {
+                    writer.println(element.toString());
+                }
+            }
+        }
+        writer.println();
+        writer.flush();
+        writer.close();
+    }
+
```
andresrc: Bad exception handling. In general:

```java
Resource res = getResource();
try {
    // use resource
} catch(InterestingException e) {
    // if any
} finally {
    // free resource
}

File: src/main/java/hudson/remoting/AgentFailuresFileListCap.java [hunk 14]
```diff
@@ -0,0 +1,83 @@
...
+        this(folder,null,size);
+    }
+
+    public AgentFailuresFileListCap(File folder, FilenameFilter filter, int size) {
+        this.folder = folder;
+        this.size = size;
+
+        if (!folder.exists() && !folder.mkdirs())
+            throw new Error("Failed to create "+folder);
+
```
andresrc: What if `file` exists but is not a directory?

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+        // Sorting in ascending order.
+        Arrays.sort(sortedFiles, new Comparator<File>() {
+            public int compare(File o1, File o2) {
+                long l = o1.lastModified() - o2.lastModified();
+                if (l<0)    return -1;
+                if (l>0)    return 1;
+                return 0;
+            }
+        });
+
```
andresrc: If the files are created by the system, with an specific pattern, maybe you can rely on it instead of `lastModified` which may provide surprises in some contexts.

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        files.add(f);
+    }
+
+    public synchronized void touch(File f){
+        files.remove(f);
+        add(f);
+    }
+
```
andresrc: where is this used?

andresrc:
In general I would go for somewhat different design. The `AgentFailuresFileListCap` would provide the `Writer` at incident time and when disposing it would simply maintain the desired number of files.

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * Channel we are going to study
+     */
+    private Channel c;
+
+    /**
+     * The cause of the ping failure
+     */
+    private Throwable cause;
```
batmat: Possibly make those fields final if the goal is to make that analyzing class immutable

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+        threadDumps.add(file);
+
+        PrintWriter writer;
+        try {
+            writer = new PrintWriter(file, "UTF-8");
+        } catch (FileNotFoundException e) {
+            throw e;
+        } catch (UnsupportedEncodingException e) {
+            throw e;
+        }
```
batmat: I must be missing something. Is that whole try/catch block even useful?

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+                    t.getLockOwnerName(),
+                    t.getLockOwnerId(),
+                    t.getLockOwnerId());
+        }
+
+        final StackTraceElement[] elements = t.getStackTrace();
+        final MonitorInfo[] monitors = t.getLockedMonitors();
+
+        for (int i = 0; i < elements.length; i++) {
+            final StackTraceElement element = elements[i];
```
batmat: NIT(?): why not java5 style? Probably leaner.
```
for(StackTraceElement element : elements) {
``` 
instead of those two lines and the `i` index "management".

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+            writer.printf("    at %s\n", element);
+            for (int j = 1; j < monitors.length; j++) {
+                final MonitorInfo monitor = monitors[j];
```
batmat: Same here, not need to have to cope with `i` and `j`.

src/main/java/hudson/remoting/AgentFailuresAnalyzer.java
```diff
@@ -0,0 +1,220 @@
...
+                if (monitor.getLockedStackDepth() == i) {
```
batmat: Ok, you're using `i` there at least.

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+    public AgentFailuresFileListCap(File folder, int size) {
+        this(folder,null,size);
+    }
+
+    public AgentFailuresFileListCap(File folder, FilenameFilter filter, int size) {
+        this.folder = folder;
+        this.size = size;
+
+        if (!folder.exists() && !folder.mkdirs())
+            throw new Error("Failed to create "+folder);
```
batmat: NIT: just me, but IMO never ever use blocks without braces. Just not worth the risk when touching that code later.

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+        File[] sortedFiles = folder.listFiles(filter);
+        if (sortedFiles==null)  return;
+
+        // Sorting in ascending order.
+        Arrays.sort(sortedFiles, new Comparator<File>() {
+            public int compare(File o1, File o2) {
+                long l = o1.lastModified() - o2.lastModified();
+                if (l<0)    return -1;
+                if (l>0)    return 1;
+                return 0;
```
batmat: replace the 4 lines above by `return o1.lastModified() - o2.lastModified()` ?

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+        files.add(f);
+    }
+
+    public synchronized void touch(File f){
+        files.remove(f);
+        add(f);
+    }
+
+    /**
+     * Creates a new file object in this directory without changing the
```
batmat: "the ..." missing word?

src/main/java/hudson/remoting/PingThreadFileListCap.java
jtnord: must have been late when I commented :-(

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -61,14 +61,24 @@
...
         super("Ping thread for channel "+channel);
         this.channel = channel;
         this.timeout = timeout;
         this.interval = interval;
+        this.pingFailureAnalyzer = pingFailureAnalyzer;
         setDaemon(true);
     }
 
+    public PingThread(Channel channel, long timeout, long interval) {
+        this(channel, 4*60*1000/*4 mins*/, interval, false);
```
jtnord: :bug: this should pass the timeout supplied and not replace it with something else...

src/main/java/hudson/remoting/PingThreadFileListCap.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Maintains most recent N files in a directory in cooperation with the writer.
+ *
+ * Useful to record incidents as files without bankrupting
+ *
+ */
+public class PingThreadFileListCap {
```
jtnord: does this need to be public - seems its internal and there is IIRC something in Jenkins to perform this for other users.

src/main/java/hudson/remoting/PingFailureAnalyzer.java
jtnord: :bug: writer is not closed.

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    @IgnoreJRERequirement
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings(
+            value = {"VA_FORMAT_STRING_USES_NEWLINE"},
+            justification = "We don't want platform specific"
+    )
+    public void takeThreadDump() {
+        long iota = System.currentTimeMillis();
+        final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
+        File file = threadDumps.file(format.format(new Date(iota++)) + ".txt");
```
jtnord: I do not see a point in increementing iota - it is not used after here, and it is not used iin a loop - so on next call you will just get something new in time - unless this happens within a 10ms granualatity on a low resolution timer - but if that is the case the iota++ will still not do anything!

src/main/java/hudson/remoting/PingFailureAnalyzer.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+        cause.printStackTrace(writer);
+        writer.close();
+    }
+
+    /**
+     * Dumps all of the threads' current information to an output stream.
+     *
+     */
+    @IgnoreJRERequirement
```
jtnord: this seems to be wrong - why are you ignoring animal sniffer results - and what was the cause of the check failing?

src/main/java/hudson/remoting/AgentFailuresFileListCap.java
fbelzunc: Then, `folder.mkdirs()` will fail

https://docs.oracle.com/javase/7/docs/api/java/io/File.html#mkdirs()

christ66:
FWIW I am :ant: on this issue.

* There is already an extension point for analyzing ping failures called `PingFailureAnalyzer`. I tried creating https://github.com/jenkinsci/jenkins/pull/1988 earlier which performed a thread dump on the node machine if it hit a ping timeout.
* I believe this should not be a JVM argument but some checkbox which can be enabled/disabled this option via the UI for all slaves.
* I'm more inclined to have this be a plugin which uses the `PingFailureAnalyzer` extension point. Sometimes thread dump information is not needed, and it might be good to implement more diagnostic tools other than thread dumps (like heap histogram, or networking checks).

fbelzunc:
@christ66 Answering you in-line:

> There is already an extension point for analyzing ping failures called PingFailureAnalyzer. 

I am aware of `PingFailureAnalyzer` - I am using it on the PR. But this is not enough to correctly troubleshooting the slave disconnection issues. 

> I tried creating jenkinsci/jenkins#1988 earlier which performed a thread dump on the node machine if it hit a ping timeout.

I am happy to close this PR if you plan to address this problem in short/med-term.

> I believe this should not be a JVM argument but some checkbox which can be enabled/disabled this option via the UI for all slaves.

If this is a concern, it can be addressed.

> I'm more inclined to have this be a plugin which uses the PingFailureAnalyzer extension point. Sometimes thread dump information is not needed, and it might be good to implement more diagnostic tools other than thread dumps (like heap histogram, or networking checks).

I agree my initial plan was to use this only for slaves - as we cannot address the problem with plugins. For Jenkins Master I agree it should be done through support-core plugin.

@oleg-nenashev @christ66 @batmat I have very limited time to work on these matters so before continuing here I would like to ensure that this PR will be fine addressing the comments. So basically create a checkbox on the UI and make this only available for slaves - for master will be support-core plugin.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -516,19 +518,40 @@ public static void main(InputStream is, OutputStream os, Mode mode, boolean perf
...
         long timeout = 1000 * Long.parseLong(
                 System.getProperty("hudson.remoting.Launcher.pingTimeoutSec", "240")),
              interval = 1000 * Long.parseLong(
                 System.getProperty("hudson.remoting.Launcher.pingIntervalSec", "600"));
+
+        final boolean agentFailureAnalyzer = Boolean.getBoolean("hudson.remoting.Launcher.agentFailureAnalyzer");
+
         if (performPing && timeout > 0 && interval > 0) {
             new PingThread(channel, timeout, interval) {
+                @Deprecated
```
christ66: It would be nice to have javadoc for deprecating a method. Something like:

```java
@Deprecated Use {@link PingThread#onDead(Throwable)}
```

oleg-nenashev:
I don't like the current design much as well. Makes sense to take it offline and review together

oleg-nenashev:
@fbelzunc Any plans to work on it?

fbelzunc:
@oleg-nenashev since you don't like the current design, I am transitioning this to close.

PR #77, opened 2016-03-04
The send and receive methods are supposed to return -1 on EOF

stephenc:
The current implementations do not catch the ChannelClosedException and return -1, instead relying on the implied
behaviour of the JVM to return -1 appropriately. This change fixes the code path so that on newer JVMs the -1
is propagated instead.

Without this change, if you use NioChannelHub to manage client sockets you will get

```
Mar 04, 2016 10:50:56 AM org.jenkinsci.remoting.nio.NioChannelHub run
WARNING: Communication problem
java.nio.channels.ClosedChannelException
	at sun.nio.ch.SocketChannelImpl.ensureReadOpen(SocketChannelImpl.java:256)
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:299)
	at org.jenkinsci.remoting.nio.FifoBuffer$Pointer.receive(FifoBuffer.java:137)
	at org.jenkinsci.remoting.nio.FifoBuffer.receive(FifoBuffer.java:310)
	at org.jenkinsci.remoting.nio.NioChannelHub.run(NioChannelHub.java:561)
	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

on close

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
+++ b/src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
@@ -6,6 +6,7 @@ import java.io.InputStream;
 import java.io.InterruptedIOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.nio.channels.ClosedChannelException;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.channels.WritableByteChannel;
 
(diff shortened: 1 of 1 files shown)


amuniz:
:bee:

src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
```diff
@@ -250,16 +251,19 @@ public int send(WritableByteChannel ch) throws IOException {
...
 
-                lock.notifyAll();
+                    lock.notifyAll();
 
-                if (sent==0)    // channel filled up
-                    return read;
+                    if (sent == 0)    // channel filled up
+                        return read;
+                } catch (ClosedChannelException e) {
+                    return -1; // propagate EOF
```
rsandell: It looks like there is a potential for `read > 0` and so there actually was something sent in a previous pass of the loop, so returning `-1` would be a lie in that case?

stephenc: Not in my opinion... as `-1` means we have reached the EOF... doesn't matter how much we sent, we can send no more...

If people have strong views on this I am ok removing the handling from `send`... it's the `receive` side that is annoying me with pointless stack traces

stephenc:
@reviewbybees done

PR #78, opened 2016-03-04
Need a line with known text to find failures

kohsuke:
Server who doesn't understand the JNLP3 protocol sends us a line like
"Unknown protocol:....", whereas the protocol expects an integer.
This results in the following exception:

    SEVERE: For input string: "Unknown protocol:Protocol:JNLP3-connect"
    java.lang.NumberFormatException: For input string: "Unknown protocol:Protocol:JNLP3-connect"
            at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
            at java.lang.Integer.parseInt(Integer.java:492)
            at java.lang.Integer.parseInt(Integer.java:527)
            at org.jenkinsci.remoting.engine.JnlpProtocol3.initiateAndValidateMaster(JnlpProtocol3.java:185)
            at org.jenkinsci.remoting.engine.JnlpProtocol3.performHandshake(JnlpProtocol3.java:141)
            at org.jenkinsci.remoting.engine.JnlpProtocol.establishChannel(JnlpProtocol.java:77)
            at hudson.remoting.Engine.run(Engine.java:308)

To gracefully handle this, we need JNLP3 protocol to send a known line
so that the client can easily distinguish a successful start of the
authentication/negotiation against a downright rejection. I'e added
the "Negotiate\n" line to do this here.

In JNLP1 and JNLP2, the first response thta the server sends is GREETING_SUCCESS,
so this problem doesn't apply. JNLP3 is the first protocol that requires
some back & forth, which revealed this hidden assumption.

Calling out @akshayabd and  @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
@@ -85,7 +85,11 @@ public abstract class JnlpProtocol {
      * Performs a handshake with the master.
      *
      * @param outputStream The stream to write into to initiate the handshake.
-     * @param inputStream The stream to read responses from the master.
+     * @param inputStream
+     *      The stream to read responses from the master.
+     *      If the server doesn't understand the protocol, the first line it sends to the client is
+     *      an error message in one line, so a protocol should send a known success message
+     *      (like {@link #GREETING_SUCCESS}) to help differentiate here.
      * @return true iff handshake was successful.
      * @throws IOException
      */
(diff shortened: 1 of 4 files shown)


kohsuke:
Corresponding change in Jenkins at https://github.com/jenkinsci/jenkins/pull/2010

kohsuke:
The tests have timed out. Retrying.

oleg-nenashev:
The change looks good to me itself, but seems it causes incompatibility between new and old JNLP3 protocol versions. So the communication won't work properly if Jenkins slaves are not updated. Right?

stephenc:
Well as v3 has not been turned on yet (as I understand it) it shouldn't be an issue if fixed now

kohsuke:
Correct, JNLP3 hasn't shipped yet in Jenkins

PR #79, opened 2016-04-04
[JENKINS-33999] Do not blacklist Spring’s Exceptions

jtnord:
`NestedRuntimeException` and others needs to be propagated across remoting for things
like remote authentication schemes.

@reviewbybees
[JENKINS-33999](https://issues.jenkins-ci.org/browse/JENKINS-33999)

### DIFF
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -60,7 +60,7 @@ public abstract class ClassFilter {
         ".*org[.]apache[.]xalan.*",
         "^org[.]codehaus[.]groovy[.]runtime[.].*",
         "^org[.]hibernate[.].*",
-        "^org[.]springframework[.].*",
+        "^org[.]springframework[.](?!core[.]NestedRuntimeException$).*",
         "^sun[.]rmi[.].*",
     };
 
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -60,7 +60,7 @@ public abstract class ClassFilter {
         ".*org[.]apache[.]xalan.*",
         "^org[.]codehaus[.]groovy[.]runtime[.].*",
         "^org[.]hibernate[.].*",
-        "^org[.]springframework[.](?!core[.]NestedRuntimeException$).*",
+        "^org[.]springframework[.](?!(\\p{Alnum}+[.])*\\p{Alnum}*Exception$).*",
         "^sun[.]rmi[.].*",
     };
 
--- a/src/test/java/hudson/remoting/DefaultClassFilterTest.java
+++ b/src/test/java/hudson/remoting/DefaultClassFilterTest.java
@@ -49,10 +49,11 @@ public class DefaultClassFilterTest {
     private static final List<String> defaultBadClasses = Arrays.asList("org.codehaus.groovy.runtime.Bob",
                                     "org.apache.commons.collections.functors.Wibble", "org.apache.xalan.Bogus",
                                     "com.sun.org.apache.xalan.bogus", "org.springframework.core.SomeClass", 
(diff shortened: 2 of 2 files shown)


oleg-nenashev:
Seems to be safe enough :+1:

jglick:
:bee:

jglick:
:bug: you are still missing an exemption for `org.springframework.dao.DataAccessException`, which is declared by `SecurityRealm.loadUserByUsername`.

jglick:
I would feel safer if you limited the whitelist to just `NestedRuntimeException` and `DataAccessException`, the ones we actually need. I cannot find any `SomethingException`s that look unsafe in `spring-core-2.5.6.SEC03.jar`, but you are now opening up classes in scarier-looking libraries like `spring-aop`. Have you actually inspected all exception classes in all Spring-namespaced libraries used by Jenkins core or plugins and satisfied yourself that they are all safe?

jtnord:
You can't just add  `DataAccessException` as it is `abstract` and so implementations would throw concrete subclasses and given the implementation will come from a plugin we are not in control of any of the which subclass this may be.
And the Actual exception if one is thrown can contain any other arbitrary `Throwable` as it supports chaining.  So IIUC the blacklist and filtering correctly we would need to allow more through anyway?
Either that or the other side needs to replace all exceptions should they occur and replace them with an exception that is safe to throw and introduce a new concrete subtype.

> Have you actually inspected all exception classes in all Spring-namespaced libraries used by Jenkins core or plugins and satisfied yourself that they are all safe

I cloned [spring-framework](https://github.com/spring-projects/spring-framework) checked out the v3.0.0.M1 tag (2.5.6 is not tagged in github I think close enough) and ran `grep -l -r <pattern>. | xargs -I FILE grep -l class.*Exception FILE` where pattern is `readObjectNoData` `readObject` and `readResolve`
I then looked at the matches to see if they where within a class ending with Exception (including inner classes).

Now there are several exceptions that have references to other objects via interfaces or the ilk so any one of those could be dangerous - but those particular classes should not be deserialized across remoting (again IIUC) if they are bad (unless they are not already filtered which means something is already bad so this does not introduce any new vectors)

jglick:
> You can't just add `DataAccessException` as it is `abstract` and so implementations would throw concrete subclasses

Only if the concrete subclasses, or intermediate types, are defined in Spring. The only subtype I can find which seems likely to be thrown by a `SecurityRealm` is `LdapDataAccessException`, which is in an Acegi namespace and directly extends `DataAccessException`. The rest seem specific to SQL/JDBC.

> the actual exception if one is thrown can contain any other arbitrary `Throwable` as it supports chaining

Sure; the question is whether there are other `Throwable` subtypes we are currently blacklisting which some legitimate Jenkins remoting call might actually be using. Whether these subtypes are used as `cause`s or as the primary exception is irrelevant.

:bee: since you say you inspected all of the types being whitelisted and saw nothing dangerous.

stephenc:
Could we auto whitelist any Throwable class that doesn't have the readXxx methods or did I miss something?

jtnord:
@stephenc we should whitelist all classes that don't have readXxx, but our filtering doesn't support that.  We would need a different filtering implementation.  Now that is something that we could possibly do...
but that at least is less likely to get back ported to an LTS branch near you!

oleg-nenashev:
I'm also aware about the performance of such implementation. Too much reflection if we start using reflection as a fallback for all classes. For exceptions it should be fine since we don't usually transfer them over the channel in the case of the stable work.

jtnord:
@reviewbybees done

oleg-nenashev:
Ready to go.
@jtnord Do we also want to have it in https://github.com/jenkinsci/remoting/tree/2.53.x ?

jtnord:
@oleg-nenashev yes please - as that's what is still in the upcoming LTS line.

PR #80, opened 2016-04-11
Ensure a message is logged if we can not create an override

jtnord:
If the first access of the ClassFilter is from a class being auto wired
then the root cause of the Issue can get swallowed and it is not clear in
the logs what is wrong.  So if something goes wrong always log the error
at the expense that it gets logged twice (which is not a bad thing as it
is fatal anyway and will cause Jenkins to pretty much not work at all)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -81,8 +81,8 @@ public abstract class ClassFilter {
      * values provide for user specified overrides - see {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
      */
     /*package*/ static ClassFilter createDefaultInstance() {
-        List<Pattern> patternOverride = loadPatternOverride();
         try {
+            List<Pattern> patternOverride = loadPatternOverride();
             if (patternOverride != null) {
                 LOGGER.log(Level.FINE, "Using user specified overrides for class blacklisting");
                 return new RegExpClassFilter(patternOverride);
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -82,12 +82,19 @@ public abstract class ClassFilter {
      */
     /*package*/ static ClassFilter createDefaultInstance() {
         List<Pattern> patternOverride = loadPatternOverride();
-        if (patternOverride != null) {
-            LOGGER.log(Level.FINE, "Using user specified overrides for class blacklisting");
-            return new RegExpClassFilter(patternOverride);
-        } else {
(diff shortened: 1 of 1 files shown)


kzantow:
🐝

mslusarczyk:
:bee:

jtnord:
@reviewbybees done

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -82,12 +82,19 @@ public final Class check(Class c) {
...
+                LOGGER.log(Level.FINE, "Using user specified overrides for class blacklisting");
+                return new RegExpClassFilter(patternOverride);
+            } else {
+                LOGGER.log(Level.FINE, "Using default in built class blacklisting");
+                return new RegExpClassFilter(DEFAULT_PATTERNS);
+            }
+        }
+        catch (Error e) {
+            // when being used by something like XStream the actual cause gets swallowed
+            LOGGER.log(Level.SEVERE, "Failed to initialize the default class filter", e);
```
oleg-nenashev: I'm not a big fan of this construct, because it may hide the original OOM if a new one happens within Logger listeners. But good for other error types

oleg-nenashev:
:bee: with a corner-case grumbling

src/main/java/hudson/remoting/ClassFilter.java
jtnord: this happens early on so if we are OOM there then wehey...
now you can argue this should be `Throwable` or a specific `Error` should be thrown.
or even a big fat :bug: that ` List<Pattern> patternOverride = loadPatternOverride();` should be inside the `try` block.

jtnord:
:bug: on my own code... it's late and I need to take a look tomorrow, but despite all the positive reviews I think this is not correct.

oleg-nenashev:
@jtnord any plans?
May be a subject for backporting

jtnord:
@oleg-nenashev need to fix it up, need to get some other bits done first so no a huge priority for me right now.

oleg-nenashev:
Sure, thanks for update!

oleg-nenashev:
@jtnord Would be great to get it integrated by the end of this week. I'm planning a new release, which may still get into the LTS

jtnord:
removing my own bug with a :bee:

oleg-nenashev:
🐝 from me

amuniz:
:bee:

jtnord:
@reviewbybees done (again)

jglick:
FTR this looks like [JDK-8051847](https://bugs.openjdk.java.net/browse/JDK-8051847).

PR #82, opened 2016-04-14
Force class load on UserRequest, see JENKINS-19445

jenkins-bbva:
This is a workaround to force class load of any class. This helps prevent deadlock on windows nodes when using JNA and Subversion.
Use property hudson.remoting.RemoteClassLoader.force to name the class to load. Ideally this forced load should happen earlier on the startup, however the classloader isn't available.
link: https://issues.jenkins-ci.org/browse/JENKINS-19445

### DIFF
--- a/src/main/java/hudson/remoting/UserRequest.java
+++ b/src/main/java/hudson/remoting/UserRequest.java
@@ -34,6 +34,8 @@ import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.io.NotSerializableException;
 import java.io.ObjectInputStream;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * {@link Request} that can take {@link Callable} whose actual implementation
@@ -86,7 +88,7 @@ final class UserRequest<RSP,EXC extends Throwable> extends Request<UserResponse<
         return result;
     }
 
-    static boolean workaroundDone = false;
+    private static boolean workaroundDone = false;
     protected UserResponse<RSP,EXC> perform(Channel channel) throws EXC {
         try {
             ClassLoader cl = channel.importedClassLoaders.get(classLoaderProxy);
@@ -96,6 +98,9 @@ final class UserRequest<RSP,EXC extends Throwable> extends Request<UserResponse<
             // @Related    https://issues.tmatesoft.com/issue/SGT-451
             final String clazz = System.getProperty(RemoteClassLoader.class.getName() + ".force", null);
             if ( clazz != null && !workaroundDone) {
+                // Optimistic logging set.
+                String eventMsg = "Loaded";
+                Level logLevel = Level.INFO;
                 // java.lang classes can only be instantiated by the bootstrap Classloader.
                 // Guarantees that *all* threads with whatever Classloader in use, have the
                 // same mutex instance:    an intance of java.lang.Class<java.lang.Object>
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -86,10 +86,32 @@ public UserRequest(Channel local, Callable<?,EXC> c) throws IOException {
...
 
+    static boolean workaroundDone = false;
     protected UserResponse<RSP,EXC> perform(Channel channel) throws EXC {
         try {
             ClassLoader cl = channel.importedClassLoaders.get(classLoaderProxy);
 
+            // Allow forcibly load of a class, allows to workaround:
+            // @See        https://issues.jenkins-ci.org/browse/JENKINS-19445
+            // @Related    https://issues.tmatesoft.com/issue/SGT-451
+            final String clazz = System.getProperty(RemoteClassLoader.class.getName() + ".force", null);
```
oleg-nenashev: I would vote for initializing it once as a static variable.
I don't see how the dynamic provisioning may help, but it decreases the performance a bit

src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -86,10 +86,32 @@ public UserRequest(Channel local, Callable<?,EXC> c) throws IOException {
         return result;
     }
 
+    static boolean workaroundDone = false;
```
oleg-nenashev: should be private

src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -86,10 +86,32 @@ public UserRequest(Channel local, Callable<?,EXC> c) throws IOException {
...
+            if ( clazz != null && !workaroundDone) {
+                // java.lang classes can only be instantiated by the bootstrap Classloader.
+                // Guarantees that *all* threads with whatever Classloader in use, have the
+                // same mutex instance:    an intance of java.lang.Class<java.lang.Object>
+                synchronized(java.lang.Object.class)
+                {
+                    workaroundDone = true;
+                    try {
+                        final Class<?> loaded = Class.forName( clazz, true, cl );
+                        System.err.println("Loaded class: '" + clazz + "' using classloader: " + cl);
```
oleg-nenashev: 1) Should be java.util.logging.Logger instead of the raw output
2) It would be also great to move such IO operation outside the Object class lock

src/main/java/hudson/remoting/UserRequest.java
jenkins-bbva: This is the only point we are certain to have the same ClassLoader with the same rules. Not sure how to obtain the same ClassLoader from a static block. We checked all the flow from Launcher to here, as the only point were it would be safe, and early enough to prevent the blockage.

src/main/java/hudson/remoting/UserRequest.java
jenkins-bbva: Agreed, many time away from coding in Java.

src/main/java/hudson/remoting/UserRequest.java
jenkins-bbva: 1) Which logger name?
2) Good idea

oleg-nenashev: In the Jenkins project we commonly use class name as a logger name

jenkins-bbva: Since we are using RemoteClassLoader for the property, I used the same class for logging.

oleg-nenashev: OK

oleg-nenashev:
👍

oleg-nenashev:
My action item is to document it somehow

jglick:
Is this some sort of bug in JNA? Is there a filed issue for that?

PR #83, opened 2016-05-12
[Related to JENKINS-34121] Allow disabling the protocols individually

stephenc:
- Also ensure that a fatal unexpected error in one protocol does not prevent a fall-back

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -302,12 +302,24 @@ public class Engine extends Thread {
                 Channel channel = null;
 
                 // Try available protocols.
+                boolean triedAtLeastOneProtocol = false;
                 for (JnlpProtocol protocol : protocols) {
+                    if (!protocol.isEnabled()) {
+                        events.status("Protocol " + protocol.getName() + " is not enabled, skipping");
+                        continue;
+                    }
+                    triedAtLeastOneProtocol = true;
                     events.status("Trying protocol: " + protocol.getName());
                     try {
                         channel = protocol.establishChannel(jnlpSocket, channelBuilder);
                     } catch (IOException ioe) {
-                        events.status("Protocol failed to establish channel", ioe);
+                        events.status("Protocol " + protocol.getName() + " failed to establish channel", ioe);
+                    } catch (RuntimeException e) {
+                        events.status("Protocol " + protocol.getName() + " encountered a runtime error", e);
+                    } catch (Error e) {
+                        events.status("Protocol " + protocol.getName() + " could not be completed due to an error", e);
+                    } catch (Throwable e) {
+                        events.status("Protocol " + protocol.getName() + " encountered an unexpected exception", e);
                     }
 
                     // On success do not try other protocols.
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -302,12 +302,24 @@ public void run() {
...
                     events.status("Trying protocol: " + protocol.getName());
                     try {
                         channel = protocol.establishChannel(jnlpSocket, channelBuilder);
                     } catch (IOException ioe) {
-                        events.status("Protocol failed to establish channel", ioe);
+                        events.status("Protocol " + protocol.getName() + " failed to establish channel", ioe);
+                    } catch (RuntimeException e) {
+                        events.status("Protocol " + protocol.getName() + " encountered a runtime error", e);
+                    } catch (Error e) {
+                        events.status("Protocol " + protocol.getName() + " could not be completed due to an error", e);
```
oleg-nenashev: Uhm. Do we really want to continue after it. Errors are commonly fatal

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol.java
```diff
@@ -58,6 +58,14 @@
     }
 
     /**
+     * Whether this protocol is enabled for connecting.
+     * @return {@code true} if this protocol is enabled.
+     */
+    public boolean isEnabled() {
+        return !Boolean.getBoolean(getClass().getName()+".disabled");
```
oleg-nenashev: so we retrieve the property every time. Useful for debugging, performance should not suffer in this place

src/main/java/hudson/remoting/Engine.java
stephenc: IOError is one that we should continue after and I fear could be one that gets thrown

stephenc: feel free to update this PR to whatever you want, but if JnlpProtocol3 throws an Error, e.g. LinkerError, etc we should still try the others

oleg-nenashev:
Yay, remoting build failed due to the remoting bug.

```
Tests run: 173, Failures: 0, Errors: 0, Skipped: 0

[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 7 iota=14
Build was aborted
Putting comment on the pull request
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 7 iota=14
channel stopped
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: java.io.IOException: Backing channel is disconnected.
```

src/main/java/hudson/remoting/Engine.java
oleg-nenashev: I think it's acceptable. If the error is not handleable, it will just blow up in another place

oleg-nenashev:
🐝

oleg-nenashev:
Created a separate issue for it: https://issues.jenkins-ci.org/browse/JENKINS-34819

PR #84, opened 2016-05-26
[FIXED JENKINS-32326] Support no_proxy environment variable

etiennebec:
https://issues.jenkins-ci.org/browse/JENKINS-32326

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -108,6 +108,55 @@ class Util {
     }
 
     /**
+     * Check if given URL is in the exclusion list defined by the no_proxy environment variable.
+     * On most *NIX system wildcards are not supported but if one top domain is added, all related subdomains will also
+     * be ignored. Both "mit.edu" and ".mit.edu" are valid syntax.
+     * http://www.gnu.org/software/wget/manual/html_node/Proxies.html
+     *
+     * Regexp:
+     * - \Q and \E: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html
+     * - To match IPV4/IPV/FQDN: Regular Expressions Cookbook, 2nd Edition (ISBN: 9781449327453)
+     *
+     * Warning: this method won't match shortened representation of IPV6 address
+     */
+    static boolean inNoProxyEnvVar(String host) {
+        String noProxy = System.getenv("no_proxy");
+        if (noProxy != null) {
+            noProxy = noProxy.trim()
+                    // Remove spaces
+                    .replaceAll("\\s+", "")
+                    // Convert .foobar.com to foobar.com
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Util.java
```diff
@@ -108,6 +108,55 @@ static String indent(String s) {
     }
 
     /**
+     * Check if given URL is in the exclusion list defined by the no_proxy environment variable.
+     * On most *NIX system wildcards are not supported but if one top domain is added, all related subdomains will also
```
oleg-nenashev: does not help much with redirects, etc.

oleg-nenashev:
In Jenkins project we commonly use System properties. IMHO it would be useful to support both options. System property could have a higher priority in this case. On the other hand, such approach may lead to security concerns

src/main/java/hudson/remoting/Util.java
etiennebec: You mean HTTP 301/302 redirection? The goal of this function is **only** to determine whether the slave needs to use a proxy or not to reach its master. The actual selection of proxy address, redirection/connection to the proxy will be handled by other parts: `Util.openURLConnection()` relies on `java.net.URL.openConnection()` to do so, and the Engine does it manually: https://github.com/jenkinsci/remoting/blob/remoting-2.59/src/main/java/hudson/remoting/Engine.java#L411

I tried to mimic as much as possible the behaviour of well known tools as curl or wget which supports the `no_proxy` environment variable.

etiennebec:
I would also prefer to only rely on system properties. But previous PR #27 introduced the use of `http_proxy` environment variable for no reason without supporting `no_proxy` exception list (equivalent of `http.nonProxyHosts` but less powerful).

So since #27 was merged, **every** remoting release includes a bug which prevent a slave to connect to its master when `http_proxy` was set. Because even if the system properties have precedence over this variable (see `Util.openURLConnection()` and `Util.getResolvedHttpProxyAddress()`), `http_proxy` was used as a fallback when they were not defined.

To sum up if you must use a proxy to access internet and has configured `no_proxy` so apt/curl/wget for instance could download files, your slave cannot currently contact its master which is on the same network.

Please have a look to the last comments of [JENKINS-28289](https://issues.jenkins-ci.org/browse/JENKINS-28289?focusedCommentId=258469&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-258469) and [JENKINS-32326](https://issues.jenkins-ci.org/browse/JENKINS-32326) for further details.

src/main/java/hudson/remoting/Util.java
oleg-nenashev: Agreed, it should work

oleg-nenashev:
LGTM 👍 
Needs some streamlining for System Properties, but it's a follow-up task IMHO

PR #86, opened 2016-06-06
[JENKINS-22722] - Make AsynchronousCommandTransport tolerant against Socket timeouts

oleg-nenashev:
This change prevents failures of the remoting channel on SocketTimeout exceptions if the channel is in the operational state after it. The original behavior can be restored by setting a system property.

It should be a more graceful fix of the issue, which @fengxx was trying to fix in https://github.com/jenkinsci/remoting/pull/68

@reviewbybees @KostyaSha @fengxx @jtnord

### DIFF
--- a/src/main/java/hudson/remoting/SynchronousCommandTransport.java
+++ b/src/main/java/hudson/remoting/SynchronousCommandTransport.java
@@ -3,6 +3,7 @@ package hudson.remoting;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InterruptedIOException;
+import java.net.SocketTimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/SynchronousCommandTransport.java
```diff
@@ -17,6 +18,15 @@
...
     protected Channel channel;
 
+    private static final String RDR_SOCKET_TIMEOUT_PROPERTY_NAME=SynchronousCommandTransport.class.getName() + ".failOnSocketTimeoutInReader";
+    
+    /**
+     * Enables the original aggressive behavior, when the channel reader gets 
+     * interrupted on any {@link SocketTimeoutException}.
+     * @since TODO
+     */
+    private static boolean RDR_FAIL_ON_SOCKET_TIMEOUT = Boolean.getBoolean(RDR_SOCKET_TIMEOUT_PROPERTY_NAME);
```
KostyaSha: After you merged SystemVariables... or how it was called?

KostyaSha: It's remoting, nevermind... So library also contributing more and more system variables :)

oleg-nenashev: Exactly. As I said in the SystemProperties conversation, it was just the first baby step. We will need to move the logic outside Jenkins core into a shared lib

jtnord:
:bee: :+1:

src/main/java/hudson/remoting/SynchronousCommandTransport.java
oleg-nenashev: @kostyasha deleted his comments

andresrc:
:bee:

jglick:
:bee: if you have found some means of testing this at least interactively. Maybe there is some Docker or `iptables` magic you can use to simulate this kind of condition?

oleg-nenashev:
@jglick 
I've simulated it in the following way:
* Disable PingThreads
* Reduce SocketTimeout to 1 minute
* Launch the build with "sleep 100500" on the slave.

fengxx:
:thumbsup:

oleg-nenashev:
Had a soak testing of the change with 1-second Socket timeout for 48 hours.
Slave and master logs were flooded by warnings, but there was no channel breakage.

Integrating the PR

PR #87, opened 2016-06-09
[JENKINS-31256] Use credentials in waitForServerToBack

oleg-nenashev:
It's a follow-up to #67 

> This patch addresses three regressions introduced with HUDSON-4071, 662b0f and JENKINS-6167 as these only modified run() resulting in no proxy and no authorization support in waitForServerToBack()

> As a result the connection to the master could not be reestablished as connections were rejected with response code 401 as reported in JENKINS-31256.

https://issues.jenkins-ci.org/browse/JENKINS-31256

CC @wiizzard @reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -182,7 +182,7 @@ public class Engine extends Thread {
                     // find out the TCP port
                     HttpURLConnection con = (HttpURLConnection)Util.openURLConnection(salURL);
                     if (credentials != null) {
-                        // TODO /tcpSlaveAgentListener is unprotected so why do we need to pass any credentials?
+                        // Pass credentials to support running Jenkins behind a (reverse) proxy requiring authorization
                         String encoding = Base64.encode(credentials.getBytes("UTF-8"));
                         con.setRequestProperty("Authorization", "Basic " + encoding);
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -441,6 +410,12 @@ private Socket connect(String host, String port) throws IOException, Interrupted
     private void waitForServerToBack() throws InterruptedException {
         Thread t = Thread.currentThread();
         String oldName = t.getName();
+        SSLSocketFactory sslSocketFactory = null;
+        try {
+            sslSocketFactory = getSSLSocketFactory();
+        } catch (Throwable e) {
```
oleg-nenashev: This approach is being used in other places of this file. It is not LGTM, but it does not increase the technical debt

oleg-nenashev:
Had to remove my changes and to force push in order to squash with the correct author

oleg-nenashev:
Works locally for me, restarted the tests

oleg-nenashev:
Crap, even such merge retains the author.
@wiizzard is it principal to you? If yes, I'll revert and resubmit the stuff

oleg-nenashev:
Done, reintegrated the change. 2.60 is a target baseline

PR #90, opened 2016-07-21
[JENKINS-33287] Do not consider connection rejection as a non-recoverab…

rcgroot:
This helps against the master rejecting connections because it doesn't yet know the agent previous connection has been terminated.

https://issues.jenkins-ci.org/browse/JENKINS-33287

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -303,10 +303,12 @@ public class Engine extends Thread {
 
                 // If no protocol worked.
                 if (channel == null) {
+                    final String rejected = "The server rejected the connection: ";
                     if (triedAtLeastOneProtocol) {
-                        onConnectionRejected("None of the protocols were accepted");
+                        events.status(rejected + "None of the protocols were accepted");
+                        Thread.sleep(10 * 1000);
                     } else {
-                        onConnectionRejected("None of the protocols are enabled");
+                        events.error(new Exception(rejected + "None of the protocols were enabled"));
                         return; // exit
                     }
                     continue;


oleg-nenashev:
I definitely see how it helps with [JENKINS-33287](https://issues.jenkins-ci.org/browse/JENKINS-33287), but I'm not 100% sure it's safe for other cases. Actually it causes infinite hanging of slave process in the case of the call duplication. 

Maybe there should be a timeout, after which we consider the rejection as non-recoverable. @olivergondza WDYT?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -303,10 +303,12 @@ public void run() {
...
                 // If no protocol worked.
                 if (channel == null) {
+                    final String rejected = "The server rejected the connection: ";
                     if (triedAtLeastOneProtocol) {
-                        onConnectionRejected("None of the protocols were accepted");
+                        events.status(rejected + "None of the protocols were accepted");
+                        Thread.sleep(10 * 1000);
                     } else {
-                        onConnectionRejected("None of the protocols are enabled");
+                        events.error(new Exception(rejected + "None of the protocols were enabled"));
```
olivergondza: Provided the protocols are enabled/disabled on master side, I do not think it makes sense to wait for it to change.

olivergondza:
It seems that the code has changed a lot compared to the version I used for investigation. Can we somehow tell if it got rejected as there is already an existing connection?

oleg-nenashev:
@rcgroot gentle ping

src/main/java/hudson/remoting/Engine.java
rcgroot: The observed behavior is that the master will reject connections from a client when it falsely believes the client is already connected. This state on the master will change after it realizes the client is disconnected and a retry will succeed.

When none of the protocols are enabled an error event is fired. One of the listeners will exit the process as a result. The 'return' is effectively dead code. I found that a bit confusing when trying to debug our Jenkins agent connection reliability.

oleg-nenashev:
No response from the committer. I am closing it to cleanup the pull request queue

PR #91, opened 2016-07-26
Improve diagnostics for JarCache write errors (related to JENKINS-36947)

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-36947

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -1,5 +1,7 @@
 package hudson.remoting;
 
+import org.jvnet.animal_sniffer.IgnoreJRERequirement;
+
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
@@ -14,8 +16,10 @@ import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.SocketAddress;
 import java.net.URL;
+import javax.annotation.Nonnull;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLSocketFactory;
+import java.nio.file.Files;
 import java.util.Iterator;
 
 /**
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
👍

src/main/java/hudson/remoting/Util.java
```diff
@@ -247,4 +251,26 @@ static InetSocketAddress getResolvedHttpProxyAddress(String host, int port) thro
...
+    @IgnoreJRERequirement @SuppressWarnings("Since15")
+    static void mkdirs(@Nonnull File file) throws IOException {
+        if (file.isDirectory()) return;
+
+        try {
+            Class.forName("java.nio.file.Files");
+            Files.createDirectories(file.toPath());
+            return;
+        } catch (ClassNotFoundException e) {
+            // JDK6
```
oleg-nenashev: I think we can finally drop support of Java 6 in remoting.
@stephenc is going to introduce some incompatible code, and maybe we will need remoting 3 in any case

src/main/java/hudson/remoting/Util.java
```diff
@@ -247,4 +251,26 @@ static InetSocketAddress getResolvedHttpProxyAddress(String host, int port) thro
...
+        } catch (ExceptionInInitializerError e) {
+            // JDK7 on multibyte encoding (http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7050570)
+        }
+
+        // Fallback
+        if (!file.mkdirs()) {
+            if (!file.isDirectory()) {
+                throw new IOException("Directory not created");
```
oleg-nenashev: Maybe not ideal. E.g. we do not check this directory is writable, etc.
But it's definitely better than the original code

oleg-nenashev:
re-:+1:

src/main/java/hudson/remoting/Util.java
olivergondza: Is there any chance to cut a JDK6 release with this fix, first?

src/main/java/hudson/remoting/Util.java
olivergondza: Which directory, the target one? I think that is out of the scope of the method.

src/main/java/hudson/remoting/Util.java
oleg-nenashev: Sure. I've forgotten you're on 1.580.
Right now there is no strong need to migrate, so I would not like to hurry too much

olivergondza: Not any more, it is 1.609 but the limitation remains.

src/main/java/hudson/remoting/Util.java
oleg-nenashev: Maybe

PR #92, opened 2016-07-27
[JENKINS-36871] JNLPProtocol4

stephenc:
This is a JNLP protocol that uses TLS upgrade over the plaintext socket before exchanging any connection secrets and allows for end-to-end security leveraging the TLS certificates on the Jenkins master HTTPS end-point where the master is using TLS on the web UI.

Initial measurements show that the TLS encryption using SSLEngine coupled with the ByteBuffer backed NIO implementation does not seem to have any negative performance impact. Additionally the new NIO implementation shows significantly better scalability and performance due to the offloading of I/O processing to the worker threads (this was necessary for to use SSLEngine, but a nice side-effect when compared with the previous NIO implementation (i.e. JNLPProtocol2) which does the actual I/O processing on the selector thread and thus can cause excessive transport delays when the selector thread becomes overloaded)

See [JENKINS-36871](https://issues.jenkins-ci.org/browse/JENKINS-36871) for the tracking issue.

This code is being contributed to the Jenkins Community by CloudBees, Inc. (though some of this code was work developed by me in personal time, but I am contributing that to the Jenkins Community anyway)

Work in progress while I complete our internal process for all the files required to be released.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/remoting/92)


stephenc:
# TODO

- [ ] Remove the `JnlpProtocol` / `JnlpProtocol1` / `JnlpProtocol2` / `JnlpProtocol3` classes
- [ ] Remove dependencies of the `JnlpProtocol` classes
- [ ] Remove the `JnlpServerHandshake` and related classes
- [ ] Refactor tests of the `JnlpProtocol` classes to be tests of `JnlpProtocolHandler` classes
- [ ] Add tests of `JnlpProtocol4Handler`
- [ ] Refactor Jenkins core to use the `JnlpProtocol_Handler` classes
- [ ] Some magic in Jenkins core to make `JnlpAgentReceiver` work / replace with `JnlpConnectionStateListener`

stephenc:

stephenc:
@oleg-nenashev this is done good enough to let this be opt-in

stephenc:
Todo

- [ ] OSS the `AgentEndpoint` and `AgentEndpointLocator` code
- [ ] Refactor `Engine` to use `AgentEndpointLocator`
- [ ] Wire up the instance identity checks to the client connect with JNLP4-connect

stephenc:
@oleg-nenashev Assuming the tests pass, this is fully baked

oleg-nenashev:
I think we should replace this Maven template by Jenkinsfile.
The PR builder goes crazy (and now I know why)

LICENSE.md
```diff
@@ -0,0 +1,22 @@
+The MIT License
```
oleg-nenashev: Unrelated, but OK

pom.xml
```diff
@@ -80,16 +80,28 @@ THE SOFTWARE.
...
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.11</version>
+      <version>4.12</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.hamcrest</groupId>
-      <artifactId>hamcrest-core</artifactId>
+      <artifactId>hamcrest-all</artifactId>
```
oleg-nenashev: Generally it's a bad practice since there are known issues with compatibility of classes (IIRC with HtmlUnit). Why have you needed it?

NOTICE
```diff
@@ -0,0 +1,5 @@
+Jenkins Remoting
+Copyright (c) 2004-2016, Sun Microsystems, Inc., Kohsuke Kawaguchi, Daniel Dyer, Stephen Connolly, and others
+
+This product includes software developed at
```
oleg-nenashev: Maybe you want to put the detailed NOTICE to the root as well. Otherwise it's strange to have detailed licenses in secondary assemblies only

src/assembly/agent-load-test/agent-load-test
```diff
@@ -0,0 +1,128 @@
...
+# ----------------------------------------------------------------------------
+#
+#   Copyright (c) 2001-2006 The Apache Software Foundation.  All rights
+#   reserved.
+
+
+# resolve links - $0 may be a softlink
+PRG="$0"
+
+while [ -h "$PRG" ]; do
```
oleg-nenashev: I would prefer to have tests based on a standard framework instead, but uh OK

src/assembly/agent-load-test/agent-load-test.bat
```diff
@@ -0,0 +1,113 @@
+@REM ----------------------------------------------------------------------------
+@REM  Copyright 2001-2006 The Apache Software Foundation.
```
oleg-nenashev: Would it be possible to remove such duplication and to require Cygwin if we want to keep low-level scripting?

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+    private ByteBufferQueue sendStaging = new ByteBufferQueue(transportFrameSize);
+
+    /**
+     * Write the packet.
+     *
+     * @param header the header to write.
+     * @param data   the data to write.
+     * @throws IOException if the data could not be written.
+     */
+    protected abstract void write(ByteBuffer header, ByteBuffer data) throws IOException;
```
oleg-nenashev: So InterruptedException is not an option here? The API is not symmetric then, and this command transport may be not so good for sync protocols

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+                        readState = READ_STATE_FRAME_BODY;
+                    } else {
+                        // store the first byte for resume
+                        readFrameHeader = data.get();
+                        readState = READ_STATE_MORE_HEADER;
+                    }
+                    break;
+                case READ_STATE_MORE_HEADER:
+                    // we have one byte already
+                    readFrameHeader = ChunkHeader.parse(readFrameHeader, data.get());
```
oleg-nenashev: This state machine change looks suspicious, but I still need to dig into

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+            } catch (IOException e1) {
+                LOGGER.log(Level.WARNING, "Failed to construct Command", e1);
+            } catch (ClassNotFoundException e11) {
+                LOGGER.log(Level.WARNING, "Failed to construct Command", e11);
+            } finally {
+                int available = is.available();
+                if (available > 0) {
+                    is.skip(available);
+                }
+                IOUtils.closeQuietly(is);
```
oleg-nenashev: Why? Maybe better to let the exception go since API allows it

oleg-nenashev: Or maybe logging at least

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private void processCommand() throws IOException {
+        try {
+            FastByteBufferQueueInputStream is = new FastByteBufferQueueInputStream(receiveQueue, readCommandSizes[0]);
+            try {
+                ObjectInputStreamEx ois = new ObjectInputStreamEx(is, channel.baseClassLoader, channel.classFilter);
+                receiver.handle(Command.readFrom(channel, ois));
+            } catch (IOException e1) {
+                LOGGER.log(Level.WARNING, "Failed to construct Command", e1);
```
oleg-nenashev: So we do not throw an error if we are unable to handle the command. Smells bad

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /**
+     * Set the frame size.
+     *
+     * @param transportFrameSize the new frame size.
+     */
+    public void setFrameSize(int transportFrameSize) {
+        if (transportFrameSize <= 0 || transportFrameSize >= Short.MAX_VALUE) {
+            throw new IllegalArgumentException();
```
oleg-nenashev: No description. Also makes sense to mention restrictions and the potential exception in Javadoc

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -0,0 +1,286 @@
...
+        this.channel = channel;
+        this.receiver = receiver;
+        try {
+            // just in case any data was queued while we were waiting for the call to setup, let's trigger the
+            // flush of the receive queue.
+            while (receiver != null && readCommandIndex > 0) {
+                processCommand();
+            }
+        } catch (IOException e) {
+            LOGGER.log(Level.WARNING, "Could not flush receive buffer queue", e);
```
oleg-nenashev: So we stop on the first error and do not try other commands

src/main/java/hudson/remoting/Asynchronous.java
```diff
@@ -31,7 +31,7 @@
  * // this method actually gets executed. For example, if you invoke two async
  * // calls, they may execute in the reverse order.
  * foo.zot();
- *
+ * </pre>
```
oleg-nenashev: Thanks for catching. Makes sense to run Javadoc  linter in as a part of the build

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1065,8 +1063,8 @@ public synchronized void join(long timeout) throws InterruptedException {
     /**
      * Notifies the remote peer that we are closing down.
      *
-     * Execution of this command also triggers the {@link SynchronousCommandTransport.ReaderThread} to shut down
-     * and quit. The {@link CloseCommand} is always the last command to be sent on
+     * Execution of this command also triggers the {@code SynchronousCommandTransport.ReaderThread} to shut down
```
oleg-nenashev: Why?

src/main/java/hudson/remoting/ChunkHeader.java
```diff
@@ -11,6 +14,18 @@
  * @author Kohsuke Kawaguchi
  */
 public class ChunkHeader {
+    public static int read(ByteBuffer buf) {
+        return parse(buf.get(), buf.get());
+    }
+
+    public static int peek(ByteBuffer buf) {
```
oleg-nenashev: YAGNI?

src/main/java/hudson/remoting/CommandTransport.java
```diff
@@ -108,7 +108,7 @@
      * 
      * <h2>Closing the read pump</h2>
      * <p>
-     * {@link Channel} implements {@linkplain Channel.CloseCommand its own "end of command stream" marker}, and
+     * {@link Channel} implements {@code Channel.CloseCommand} its own "end of command stream" marker, and
```
oleg-nenashev: Why not <code>@link</code>?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -196,11 +203,87 @@ public void removeListener(EngineListener el) {
...
+                }
+                KeyManagerFactory kmf;
+                try {
+                    kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+                } catch (NoSuchAlgorithmException e) {
+                    throw new IllegalStateException("Java runtime specification requires support for default key manager", e);
+                }
+                try {
+                    kmf.init(store, password);
+                } catch (KeyStoreException e) {
```
oleg-nenashev: And no logging?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -213,133 +296,147 @@ public void run() {
...
-                                if (!name.isEmpty()) {
-                                    agentProtocolNames.add(name);
+                        triedAtLeastOneProtocol = true;
+                        events.status("Trying protocol: " + protocol.getName());
+                        try {
+                            channel = protocol.connect(jnlpSocket, headers, new JnlpConnectionStateListener() {
+
+                                @Override
+                                public void beforeProperties(@Nonnull JnlpConnectionState event) {
+                                    if (event instanceof Jnlp4ConnectionState) {
```
oleg-nenashev: 🐜 So it leaks JNLP4 implementation to base classes. Does not seem to be very good from remoting extensibility PoV. It would be preferable to add handling logic to the protocol class via new method

src/main/java/hudson/remoting/Engine.java
```diff
@@ -213,133 +296,147 @@ public void run() {
...
-                    // On failure form a new connection.
-                    jnlpSocket.close();
-                    jnlpSocket = connect(host,port);
-                }
+                    // If no protocol worked.
+                    if (channel == null) {
+                        if (triedAtLeastOneProtocol) {
+                            onConnectionRejected("None of the protocols were accepted");
+                        } else {
+                            onConnectionRejected("None of the protocols are enabled");
```
oleg-nenashev: 🐜   "are enabled and supported by the remote server"

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -215,6 +230,9 @@ public void run() throws Exception {
                 }
             });
         }
+        if (candidateCertificates != null && !candidateCertificates.isEmpty()) {
+            HttpsURLConnection.setDefaultSSLSocketFactory(getSSLSocketFactory());
```
oleg-nenashev: Add warning?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -227,6 +245,12 @@ public void run() throws Exception {
             if (this.noReconnect) {
                 jnlpArgs.add("-noreconnect");
             }
+            if (candidateCertificates != null && !candidateCertificates.isEmpty()) {
+                for (String c: candidateCertificates) {
+                    jnlpArgs.add("-cert");
+                    jnlpArgs.add(c);
```
oleg-nenashev: I've thought that it maps only comma-separated params. Have you tested the multiple certificate case?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -244,6 +268,90 @@ public void run() throws Exception {
         System.exit(0);
     }
 
+    private SSLSocketFactory getSSLSocketFactory()
```
oleg-nenashev: Seems SSLSocketFactory nullability implicitly spreads widely outside this method. Maybe safe, but annotations definitely do not hurt ;)

src/main/java/hudson/remoting/Engine.java
```diff
@@ -196,11 +203,87 @@ public void removeListener(EngineListener el) {
...
+        for (URL url: candidateUrls) {
+            jenkinsUrls.add(url.toExternalForm());
+        }
+        JnlpAgentEndpointResolver resolver = new JnlpAgentEndpointResolver(jenkinsUrls);
+        resolver.setCredentials(credentials);
+        resolver.setProxyCredentials(proxyCredentials);
+        resolver.setTunnel(tunnel);
+        try {
+            resolver.setSslSocketFactory(getSSLSocketFactory());
+        } catch (Exception e) {
```
oleg-nenashev: We just throw a dozen of different exceptions from the method in order to handle them in such way

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -244,6 +268,90 @@ public void run() throws Exception {
...
+                            continue;
+                        }
+                    } else {
+                        if (file.isFile()) {
+                            LOGGER.log(Level.WARNING,
+                                    "Could not read certificate from {0}. File size is not within " +
+                                            "the expected range for a PEM encoded X.509 certificate",
+                                    file.getAbsolutePath());
+                        } else {
+                            LOGGER.log(Level.WARNING, "Could not read certificate from {0}. File not found",
```
oleg-nenashev: Maybe should be fatal error since these paths have been explicitly specified

NOTICE
stephenc: No because the root NOTICE is for remoting.jar whereas the NOTICE in the agent-load-test is for all the files in that, which includes test scoped dependencies

pom.xml
stephenc: This is a test scoped dependency and we do not use HtmlUnit in remoting

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -244,6 +268,90 @@ public void run() throws Exception {
...
+                                    file.getAbsolutePath());
+                        }
+                        continue;
+                    }
+                } else {
+                    try {
+                        cert = certOrAtFilename.getBytes("US-ASCII");
+                    } catch (UnsupportedEncodingException e) {
+                        LOGGER.log(Level.WARNING, "Could not parse certificate " + certOrAtFilename, e);
```
oleg-nenashev: US-ASCII is a standard encoding in Java IIRC. Maybe it makes sense to throw IllegalStateException like you do for X.509

src/assembly/agent-load-test/agent-load-test
stephenc: This was basically a modified appassembler script because appassembler does not handle apps from the test scope

src/assembly/agent-load-test/agent-load-test.bat
stephenc: This is just for a tool that we can distribute to users who want to test their own system limits

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
stephenc: This is just to ensure that we read whatever we committed to read so that we can stay aligned... not required by the underlying api contract, but nice to have

src/test/java/org/jenkinsci/remoting/engine/HandlerLoopbackLoadStress.java
```diff
@@ -0,0 +1,782 @@
...
+        char[] password = "password".toCharArray();
+
+        KeyStore store = KeyStore.getInstance("jks");
+        store.load(null, password);
+        store.setKeyEntry("alias", keyPair.getPrivate(), password, new Certificate[]{certificate});
+
+        KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+        kmf.init(store, password);
+
+        context = SSLContext.getInstance("TLS");
```
oleg-nenashev: NoSuchAlgorithmException is not handled

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
stephenc: Look at how the other command transports implement

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
stephenc: Nope that's the way remoting works... in part because other command transports are not able to ensure that they stay synced

src/main/java/hudson/remoting/Channel.java
stephenc: not `protected` or `public` so you cannot `{@link}`

src/main/java/hudson/remoting/ChunkHeader.java
stephenc: I need the peek and I need the read

src/main/java/hudson/remoting/CommandTransport.java
stephenc: package scope, so cannot link in javadoc

src/main/java/hudson/remoting/Engine.java
stephenc: These exceptions **will never happen** on a valid JVM... if they happen you are not running on a JVM

src/main/java/hudson/remoting/Engine.java
stephenc: More the other connection state types do not have a certificate property

src/main/java/hudson/remoting/Engine.java
```diff
@@ -213,133 +296,147 @@ public void run() {
...
+                } catch (Exception e) {
+                    events.error(e);
+                    return;
+                }
+                if (endpoint == null) {
+                    events.status("Could not resolve server among " + candidateUrls);
+                    return;
+                }
+
+                events.status(String.format("Agent discovery successful%n"
```
oleg-nenashev: Do we really need a multi-line event here?

src/main/java/hudson/remoting/Engine.java
stephenc: Nope because you may have disabled them locally by system properties also

src/main/java/hudson/remoting/Launcher.java
stephenc: why?

src/main/java/hudson/remoting/Util.java
```diff
@@ -218,40 +218,6 @@ static URLConnection openURLConnection(URL url) throws IOException {
         return openURLConnection(url, null, null, null);
     }
 
-    static InetSocketAddress getResolvedHttpProxyAddress(String host, int port) throws IOException {
```
oleg-nenashev: 4 reviewers: t has been moved to JnlpAgentEndpointResolver

src/main/java/org/jenkinsci/remoting/engine/Jnlp3ConnectionState.java
```diff
@@ -0,0 +1,91 @@
...
+import java.net.Socket;
+import java.util.List;
+import javax.annotation.Nonnull;
+
+/**
+ * Represents the connection state of a {@link JnlpProtocol3Handler} connection.
+ *
+ * @since FIXME
+ */
+@Deprecated
```
oleg-nenashev: Deprecated along with the entire protocol, right?

src/main/java/org/jenkinsci/remoting/engine/Jnlp3Util.java
```diff
@@ -29,6 +29,7 @@
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.Arrays;
+import org.jenkinsci.remoting.util.Charsets;
```
oleg-nenashev: uh oh

src/main/java/org/jenkinsci/remoting/engine/Jnlp4ConnectionState.java
```diff
@@ -0,0 +1,83 @@
...
+     * @throws ConnectionRefusalException if the connection has been refused.
+     */
+    void fireBeforeProperties(X509Certificate certificate) throws ConnectionRefusalException {
+        if (this.certificate != null) {
+            throw new IllegalStateException("fireBeforeProperties has been called already");
+        }
+        this.certificate = certificate;
+        try {
+            super.fireBeforeProperties();
+        } catch (IllegalStateException e) {
```
oleg-nenashev: Maybe undo on any exception excluding fireBeforeProperties?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpoint.java
```diff
@@ -0,0 +1,258 @@
...
+    /**
+     * Constructor for a remote {@code Jenkins} instance.
+     *
+     * @param host      the hostname.
+     * @param port      the port.
+     * @param publicKey the {@code InstanceIdentity.getPublic()} of the remote instance (if known).
+     * @param protocols The supported protocols.
+     */
+    public JnlpAgentEndpoint(@Nonnull String host, int port, @CheckForNull RSAPublicKey publicKey,
+                             @javax.annotation.CheckForNull Set<String> protocols) {
```
oleg-nenashev: NIT: It's imported

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpoint.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    /**
+     * Gets the socket address.
+     *
+     * @return the socket address
+     */
+    @Nonnull
+    public InetSocketAddress getAddress() {
+        return new InetSocketAddress(host, port);
```
oleg-nenashev: Port range is not verified in the constructor, we can get IllegalArgumentException here

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpoint.java
```diff
@@ -0,0 +1,258 @@
...
+                String[] responseLineParts = line.split(" ");
+                if (responseLineParts.length < 2 || !responseLineParts[1].equals("200"))
+                    throw new IOException("Got a bad response from proxy: " + line);
+                while (!readLine(is).isEmpty()) {
+                    // Do nothing, scrolling through headers returned from proxy
+                }
+            }
+            return socket;
+        } catch (IOException e) {
+            IOUtils.closeQuietly(channel);
```
oleg-nenashev: Generally we need our own closeQuietly() implementation, which prints the message to logger at least

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+        Thread t = Thread.currentThread();
+        String oldName = t.getName();
+        try {
+            int retries = 0;
+            while (true) {
+                Thread.sleep(1000 * 10);
+                try {
+                    // Jenkins top page might be read-protected. see http://www.nabble
+                    // .com/more-lenient-retry-logic-in-Engine.waitForServerToBack-td24703172.html
+                    URL url = toAgentListenerURL(first(jenkinsUrls));
```
oleg-nenashev: maybe we should try different jenkins URLs in the loop since we have this loop in any case

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+
+                    HttpURLConnection con =
+                            (HttpURLConnection) openURLConnection(url, credentials, proxyCredentials, sslSocketFactory);
+                    con.setConnectTimeout(5000);
+                    con.setReadTimeout(5000);
+                    con.connect();
+                    if (con.getResponseCode() == 200) {
+                        return;
+                    }
+                    LOGGER.info("Master isn't ready to talk to us. Will retry again: response code=" + con
```
oleg-nenashev: Would be nice to mention the URL, which has been tried

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * Gets URL connection.
+     * If http_proxy environment variable exists,  the connection uses the proxy.
+     * Credentials can be passed e.g. to support running Jenkins behind a (reverse) proxy requiring authorization
+     */
+    static URLConnection openURLConnection(URL url, String credentials, String proxyCredentials,
+                                           SSLSocketFactory sslSocketFactory) throws IOException {
+        String httpProxy = null;
+        // If http.proxyHost property exists, openConnection() uses it.
+        if (System.getProperty("http.proxyHost") == null) {
```
oleg-nenashev: Should it be in <code>getResolvedHttpProxyAddress</code> instead?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+                ProxySelector.getDefault().select(URI.create(String.format("http://%s:%d", host, port))).iterator();
+        while (targetAddress == null && proxies.hasNext()) {
+            Proxy proxy = proxies.next();
+            if (proxy.type() == Proxy.Type.DIRECT) {
+                break;
+            }
+            if (proxy.type() == Proxy.Type.HTTP) {
+                final SocketAddress address = proxy.address();
+                if (!(address instanceof InetSocketAddress)) {
+                    System.err.println(
```
oleg-nenashev: 🐜 Why not just loggers?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+            } catch (MalformedURLException e) {
+                System.err.println(
+                        "Not use http_proxy property or environment variable which is invalid: " + e.getMessage());
+                con = url.openConnection();
+            }
+        } else {
+            con = url.openConnection();
+        }
+        if (credentials != null) {
+            String encoding = Base64.encode(credentials.getBytes("UTF-8"));
```
oleg-nenashev: UnsupportedEncoding may happen here, but it's an illegal state in any case

src/main/java/hudson/remoting/Launcher.java
stephenc: perhaps... but in any case this path will never be hit on a JVM, so who cares!

src/test/java/org/jenkinsci/remoting/engine/HandlerLoopbackLoadStress.java
stephenc: TLS is a mandatory algorithm in JVM 7.0+ (may even be 6.0+)

src/main/java/hudson/remoting/Engine.java
stephenc: With JUL logging outputting all 4 lines as separate events is ugly and cramming the information into 1 line looks ugly too IMHO, so yes!

src/main/java/org/jenkinsci/remoting/engine/Jnlp3ConnectionState.java
stephenc: Yep, but I refactored how the protocol gets instantiated and I'm not removing the protocol in this PR

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+                    // According to Wikipedia (no RFC defines it), 128 is the max number of subdivision for a valid
+                    // FQDN:
+                    // https://en.wikipedia.org/wiki/Subdomain#Overview
+                    while (host.matches("^([a-z0-9]+(-[a-z0-9]+)*\\.)+[a-z]{2,}$") && depth < 128) {
+                        ++depth;
+                        // Check if the no_proxy contains the host
+                        if (noProxy.matches(".*(^|,)\\Q" + host + "\\E($|,).*")) {
+                            return true;
+                        }
+                        // Remove first subdomain: master.jenkins.acme.com -> jenkins.acme.com
```
oleg-nenashev: Smells like a hack. What if we have 2 subdomains?

src/main/java/org/jenkinsci/remoting/engine/Jnlp4ConnectionState.java
stephenc: Nope... though there is a risk of an ISE being propagated by the super's listeners and thus causing an ISE... but that is a sign of a bad listener... and fail fast is the best way to fix those

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -0,0 +1,422 @@
...
+                    }
+                }
+            }
+        }
+
+        return false;
+    }
+
+    @CheckForNull
+    private static List<String> header(@Nonnull HttpURLConnection connection, String... headerNames) {
```
oleg-nenashev: :ant: So we return fields for the first matching header only. Is it OK?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpoint.java
stephenc: Done

src/main/java/org/jenkinsci/remoting/engine/JnlpClientDatabase.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    @Nonnull
+    public ValidationResult validateCertificate(@Nonnull String clientName, @Nonnull X509Certificate certificate) {
+        return ValidationResult.UNCHECKED;
+    }
+
+    /**
+     * The types of certificate validation results.
+     */
+    enum ValidationResult {
```
oleg-nenashev: public method exposes non-public enum

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: I'm just moving the method from Utils

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: I'm just moving the method from Utils

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: Just moving the existing code

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: Just moving the existing code

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: Yes. This is the preference order of headers to search for.

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java
```diff
@@ -0,0 +1,531 @@
...
+        }
+    }
+
+    /**
+     * Advances the connection state to indicate that a connection has been "secured" and the property exchange
+     * is about to take place.
+     *
+     * @throws ConnectionRefusalException if the connection has been refused.
+     */
+    /*package*/ void fireBeforeProperties() throws ConnectionRefusalException {
```
oleg-nenashev: NIT: fire.* maybe not a perfect method naming pattern since the methods not only trigger listeners, but also change the internal state of the object

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java
```diff
@@ -0,0 +1,531 @@
...
+            public void invoke(JnlpConnectionStateListener listener, JnlpConnectionState event) {
+                listener.afterChannel(event);
+            }
+        });
+    }
+
+    /**
+     * Advances the connection state to indicate that the channel has been closed.
+     *
+     * @param cause
```
oleg-nenashev: NIT: Undefined Javadoc entry

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionStateListener.java
```diff
@@ -0,0 +1,111 @@
...
+    public void beforeChannel(@Nonnull JnlpConnectionState event) {
+    }
+
+    /**
+     * Callback to notify the {@link JnlpConnectionState#approve()} listener that the {@link Channel} has been
+     * created and is available via {@link JnlpConnectionState#getChannel()}.
+     *
+     * @param event the event.
+     */
+    public abstract void afterChannel(@Nonnull JnlpConnectionState event);
```
oleg-nenashev: Why do you require implementing after*() methods in the listeners?

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void onClosed(Channel channel, IOException cause) {
+            event.fireChannelClosed(cause);
+            try {
+                event.getSocket().close();
```
oleg-nenashev: so we do not remove listener in this method like it's being done for nClosed(ProtocolStack stack, IOException cause)

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
```diff
@@ -0,0 +1,384 @@
...
+    /**
+     * Creates an {@link SSLEngine} for the specified {@link Socket}. The {@link SSLContext} may have retained session
+     * state and thus enable a shorter handshake.
+     *
+     * @param socket the socket.
+     * @return the {@link SSLEngine}.                                             s
+     */
+    private SSLEngine createSSLEngine(Socket socket) {
+        SocketAddress remoteSocketAddress = socket.getRemoteSocketAddress();
+        if (remoteSocketAddress instanceof InetSocketAddress) {
```
oleg-nenashev: remoteSocketAddress can be null, and I'm not 100% sure going to else branch is really what you want here

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
```diff
@@ -0,0 +1,384 @@
...
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void onHandshakeCompleted(SSLSession session) throws ConnectionRefusalException {
+            X509Certificate remoteCertificate;
+            try {
+                remoteCertificate = (X509Certificate) session.getPeerCertificates()[0];
+            } catch (ClassCastException e) {
+                // shouldn't happen as handshake is completed, but just in case
```
oleg-nenashev: May also happen if remote classloading rejects a class or any subclass of X509Certificate

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocolHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Whether this protocol is enabled for connecting.
+     *
+     * @return {@code true} if this protocol is enabled.
+     * @since 2.59
+     */
+    public boolean isEnabled() {
+        // map to the legacy names
+        return !Boolean.getBoolean(
+                "org.jenkinsci.remoting.engine." + getClass().getSimpleName().replace("Handler", ".disabled"));
```
oleg-nenashev: some dark magic, but ok

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocolHandler.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Get the name of the protocol.
+     */
+    public abstract String getName();
+
+    /**
+     * Whether this protocol is enabled for connecting.
+     *
+     * @return {@code true} if this protocol is enabled.
+     * @since 2.59
```
oleg-nenashev: 🐜 wrong since

oleg-nenashev:
*[src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java, line 361 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjYLHoNqZ-4YSzZJGz:-KOjYLHoNqZ-4YSzZJH-:beuqn1o) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java#L361)):*
> ```Java
>                 event.getSocket().close();
>             } catch (IOException e) {
>                 // ignore
> ```

Maybe fine log at least?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void onHandshakeCompleted(SSLSession session) throws ConnectionRefusalException {
+            X509Certificate remoteCertificate;
+            try {
+                remoteCertificate = (X509Certificate) session.getPeerCertificates()[0];
```
oleg-nenashev: Suppress unchecked cast for FindBugs?

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionStateListener.java
stephenc: That is how you pass the channel into the Node or whatever else (*cough* Shared Slave *cough* Shared JNLP Cloud) that needs to receive it

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
stephenc: if it is null then we do not have the information to short-cut the handshaking, so the else branch is the correct behaviour

oleg-nenashev:
As expected, merge conflict after #96

oleg-nenashev:
Reviewed 9 of 135 files at r1.
Review status: 9 of 135 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 129 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjd3ufKN-zNdi2N2uZ:-KOjd3ufKN-zNdi2N2u_:bt33acg) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L129)):*
> ```Java
>         this.selector = Selector.open();
>         this.executor = executor;
>         this.bufferPool = new DirectByteBufferPool(16916, Runtime.getRuntime().availableProcessors() * 4);
> ```

Would be better to allow fine-tuning of these settings via System properties

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 191 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjdQJx-bG-142I2qYw:-KOjdQJx-bG-142I2qYx:bw2aar5) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L191)):*
> ```Java
>     public void executeOnSelector(Runnable task) {
>         if (task == null) {
>             throw new NullPointerException();
> ```

NIT: <code>"task is null"</code>

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 219 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjddAbZ0guS0suQXLz:-KOjddAbZ0guS0suQXM-:bb7od0r) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L219)):*
> ```Java
>     public Future<?> executeLater(Runnable task, long delay, TimeUnit units) {
>         if (task == null) {
>             throw new NullPointerException();
> ```

same

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 416 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjf2hKKDoGC8fBqulO:-KOjf2hKKDoGC8fBqulP:bi4ymdi) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L416)):*
> ```Java
>                         // we did some work that is anticipated to either take some time or have likely resulted
>                         // in an immediately ready selection key, hence we use the non-blocking form
>                         selector.selectNow();
> ```

From what I see from Javadoc, you should fallback to select() if this method returns 0

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 422 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjfIlcK10253D2s5px:-KOjfIlcK10253D2s5py:b9udjux) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L422)):*
> ```Java
>                     Set<SelectionKey> keys = selector.selectedKeys();
>                     if (keys.isEmpty()) {
>                         // don't stress the GC by creating an iterator for an empty set
> ```

You can even check the return value of the select() method then

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 487 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjg26CuGWfCvKaWBYb:-KOjg26CuGWfCvKaWBYc:bdifgee) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L487)):*
> ```Java
>             scheduledTasks.drainTo(scheduledWork);
>             for (DelayedRunnable task : scheduledWork) {
>                 if (!task.isCancelled()) {
> ```

else exclude it from the queue?

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 765 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjgXR-tEu1dWtSCiZ4:-KOjgXR-tEu1dWtSCiZ5:blc5h37) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L765)):*
> ```Java
>         private Throwable failure;
>         /**
>          * The {@link System#currentTimeMillis()} after which the task should be executed.
> ```

Strange link in Javadoc. Just milliseconds?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

oleg-nenashev:
Reviewed 1 of 135 files at r1.
Review status: 10 of 135 files reviewed at latest revision, 10 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHubReadyListener.java, line 70 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjhBpES8wSFrxwHEUK:-KOjhBpES8wSFrxwHEUL:bnekcbi) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHubReadyListener.java#L70)):*
> ```Java
>      *                {@link IOHub#addInterestWrite(SelectionKey)} to request additional callbacks for this ready state.
>      */
>     void ready(boolean accept, boolean connect, boolean read, boolean write);
> ```

Does not seem to be very extensible for the future. Maybe split to different methods and convert to the abstract class?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

oleg-nenashev:
Reviewed 15 of 135 files at r1.
Review status: 25 of 135 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/SettableFuture.java, line 142 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjl00NRbNtmspGcm1t:-KOjl00NRbNtmspGcm1u:bxhpeoa) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/SettableFuture.java#L142)):*
> ```Java
>      * @return {@code true} if the future is now completed, {@code false} if the future has already been completed.
>      */
>     public boolean setException(@Nonnull Throwable throwable) {
> ```

So it is not possible to return both value and exception. Seems to be reasonable, but maybe makes sense to mention in Javadoc

---

*[src/main/java/org/jenkinsci/remoting/util/SettableFuture.java, line 307 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjjolgmSsD10rQsO62:-KOjjolgmSsD10rQsO63:b-4ksct1) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/SettableFuture.java#L307)):*
> ```Java
>      * Verifies that a value is non-null without findbugs complaining too much.
>      *
>      * @param o the object that should be non-null.
> ```

NIT: Add throws to Javadoc

---

*[src/main/java/org/jenkinsci/remoting/util/SettableFuture.java, line 309 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjjW7SW1PKQ7E4hdvs:-KOjjW7SW1PKQ7E4hdvt:b-4ebh2n) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/SettableFuture.java#L309)):*
> ```Java
>      * @param o the object that should be non-null.
>      */
>     @edu.umd.cs.findbugs.annotations.SuppressWarnings
> ```

Obsolete annotation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: ok

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: agreed

src/test/java/org/jenkinsci/remoting/engine/HandlerLoopbackLoadStress.java
oleg-nenashev: From Oracle's Java SE 7 Javadoc: "Every implementation of the Java platform is required to support the following standard SSLContext protocol: TLSv1". Nothing is said for SE6, which needs to be supported by remoting right now. Just "TLS" smells like a bug in any case

src/main/java/hudson/remoting/Engine.java
oleg-nenashev: So we need even better message :(

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
stephenc: Same as `AbstractByteArrayCommandTransport`

src/main/java/hudson/remoting/Launcher.java
stephenc:
```
        Main m = new Main();
        CmdLineParser p = new CmdLineParser(m);
        p.parseArgument("-cert", "foo", "-cert", "fbar", "-cert=x,y", "-cert", "z,w");
        System.out.println(m.candidateCertificates);
```

prints

```
[foo, fbar, x,y, z,w]
```

So it handles multiple certificates. it does not do comma separated

src/test/java/org/jenkinsci/remoting/engine/HandlerLoopbackLoadStress.java
stephenc: This stress tester app needs Java 7 APIs anyway

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
stephenc: I don't see findbugs complaining as we catch the `ClassCastException`

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 416 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjf2hKKDoGC8fBqulO:-KOkx9rvP-Tc_e9o3yPY:b-s95glm) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L416)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> From what I see from Javadoc, you should fallback to select() if this method returns 0
</details>
If we want selectNow it is because we spent some time doing other things. There could be more "other things" that have been scheduled and as we are not in a select loop already then the `selector.wakeup()` will have been a no-op. So if we run a `select` then those tasks will be delayed, potentially indefinitely...

So actually the correct thing is to loop back so we can check the registrations, selector tasks, etc

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 422 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjfIlcK10253D2s5px:-KOkym9rxwEwH5mOl9g1:b2nha9w) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L422)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> You can even check the return value of the select() method then
</details>
ACK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 487 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjg26CuGWfCvKaWBYb:-KOkywG1HHB-W0yZCg7g:b-ikulsj) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L487)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> else exclude it from the queue?
</details>
Already removed from the queue as we drained the queue to this list

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHub.java, line 765 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjgXR-tEu1dWtSCiZ4:-KOkzAF8mgfU9f0229eu:b-5kjnbq) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L765)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Strange link in Javadoc. Just milliseconds?
</details>
Well technically its the number of milliseconds since midnight UTC 1/Jan/1970 but as we compare with System.currentTimeMillis() to check for execution and that can drift for correction seconds it is not actually strictly correct to say "number of milliseconds since midnight UTC 1/Jan/1970"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/protocol/IOHubReadyListener.java, line 70 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjhBpES8wSFrxwHEUK:-KOkzpan_7u45HXQeqGC:b-lgxiwl) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/protocol/IOHubReadyListener.java#L70)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Does not seem to be very extensible for the future. Maybe split to different methods and convert to the abstract class?
</details>
Actually you want this as a single method given that multiple of these could be true concurrently, e.g. read & write and you may have to enforce a sequencing between the two of them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/ByteBufferUtils.java, line 80 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjp2TsdfDLq4A12-lV:-KOl-7nhyIgQdr2r7YUS:b-h5j097) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/ByteBufferUtils.java#L80)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Needs to be documented somewhere. It does not seem to be a never-ever case
</details>
This is the implicit limit from DataOutputStream and is a consequence of the "two byte length". I don't think we need to document every IAE

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/ByteBufferUtils.java, line 104 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjpHOl7abpOBF8lFSy:-KOl05XU8yOJ-OG-7Q-g:bmpz027) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/ByteBufferUtils.java#L104)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Why not just IOException or any other non-runtime one?
</details>
We are mimicking the ByteBuffer contract, which is to throw a BUE

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/DirectByteBufferPool.java, line 38 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjncEQnq4iMl8WslT-:-KOl0pGlU6IquLkWxn8R:b-vrjmkv) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/DirectByteBufferPool.java#L38)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Thread unsafe annotation?
</details>
It's thread safe. I'll document the guards

---

*[src/main/java/org/jenkinsci/remoting/util/DirectByteBufferPool.java, line 84 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjnT0bsW_pas3rbnq-:-KOl154Q74S_Pr7g-nC6:b6edc46) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/DirectByteBufferPool.java#L84)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> So we allocate new big buffer, and then add it back to the pool when we release it. Maybe makes sense to add it to the pool immediately
</details>
The pool only contains buffers that are available for use. We are giving this one to the requester directly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/FastByteBufferQueueInputStream.java, line 34 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjmLGyUV8Zu_jXBz9t:-KOl1GvdjBowhyrc8SH6:be0wkpt) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/FastByteBufferQueueInputStream.java#L34)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Same for reset(). for mark() you could throw exception since there is a boolean method verifying the support state
</details>
I looked at other streams that do not support mark and the convention seems to be to just be a no-op

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
Review status: 35 of 135 files reviewed at latest revision, 22 unresolved discussions, some commit checks broke.

---

*[src/main/java/org/jenkinsci/remoting/util/IOUtils.java, line 58 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/remoting/92#-KOjlUvFq3vZrqwFiFAW:-KOl1T0pI5TG_lWlfg53:b-5bxh7q) ([raw file](https://github.com/jenkinsci/remoting/blob/c7f2c9c2a6ebd058c4b6f926af29bfed1a3a78ec/src/main/java/org/jenkinsci/remoting/util/IOUtils.java#L58)):*
<details><summary><i>Previously, oleg-nenashev (Oleg Nenashev) wrote…</i></summary>
> Makes sense to log the exception at the FINE+ level instead of just ignoring it
</details>
I disagree

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/remoting/92)*

stephenc:
@oleg-nenashev I think I have addressed all comments

oleg-nenashev:
We still need to fix the PR builder at least.
I'll re-check comments ASAP

stephenc:
@oleg-nenashev so I have changed the elastic timeout from 150% to 300% for libraries... that *should* fix this PR builder (as we are adding a load of tests)

stephenc:
```
Caused by: org.jenkinsci.remoting.protocol.impl.ConnectionRefusalException: Incorrect master challenge response from agent
	at org.jenkinsci.remoting.engine.JnlpProtocol3Handler.receiveHandshake(JnlpProtocol3Handler.java:325)
	at org.jenkinsci.remoting.engine.JnlpProtocol3Handler.receiveHandshake(JnlpProtocol3Handler.java:123)
	at org.jenkinsci.remoting.engine.LegacyJnlpProtocolHandler$1.call(LegacyJnlpProtocolHandler.java:108)
	at org.jenkinsci.remoting.engine.LegacyJnlpProtocolHandler$1.call(LegacyJnlpProtocolHandler.java:104)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
```

I wonder is there *another* encoding bug in the JNLP3 protocol...

stephenc:
Yep! *Another* bug in JNLP3... https://github.com/jenkinsci/remoting/blob/master/src/main/java/org/jenkinsci/remoting/engine/Jnlp3Util.java#L93

stephenc:
Let's see if we can get lucky

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-37302 filed

stephenc:
Retriggering to confirm that flaky tests have been fixed

stephenc:
And let's see

stephenc:
Timeout... Ffs

oleg-nenashev:
The last changes look good.
So 🐝 from me for merging it to a remoting-3.0 branch and then perform some soak testing.

stephenc:
@oleg-nenashev I'd rather cut a 2.0 branch pre-merge and merge this on master

oleg-nenashev:
We agreed that we will follow master and stable branch approach

PR #93, opened 2016-07-28
[FIXED JENKINS-37031] tcpSlaveAgentListener should publish supported agent protocols to speed up connection setup

stephenc:
See [JENKINS-37031](https://issues.jenkins-ci.org/browse/JENKINS-37031) and https://github.com/jenkinsci/jenkins/pull/2486

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -39,7 +39,9 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import javax.annotation.CheckForNull;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -214,6 +216,7 @@ public class Engine extends Thread {
                 Throwable firstError=null;
                 String host=null;
                 String port=null;
+                Set<String> agentProtocolNames = null;
                 SSLSocketFactory sslSocketFactory = getSSLSocketFactory();
 
                 for (URL url : candidateUrls) {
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Should be safe for backporting.
The code looks good to me 🐝 

@stephenc I would kindly ask to create PRs from a fork. Otherwise Jenkins automatic issue updater goes crazy

oleg-nenashev:
@reviewbybees done

src/main/java/hudson/remoting/Engine.java
```diff
@@ -247,6 +250,16 @@ public void run() {
                         }
                         host = con.getHeaderField("X-Jenkins-JNLP-Host"); // controlled by hudson.TcpSlaveAgentListener.hostName
                         if (host == null) host=url.getHost();
+                        String names = con.getHeaderField("X-Jenkins-Agent-Protocols");
+                        agentProtocolNames = new HashSet<String>();
+                        if (names != null) {
+                            for (String name: names.split(",")) {
+                                name = name.trim();
+                                if (!name.isEmpty()) {
```
oleg-nenashev: Why not just StringUtils.isNotBlank()?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -277,6 +290,10 @@ public void run() {
                         events.status("Protocol " + protocol.getName() + " is not enabled, skipping");
                         continue;
                     }
+                    if (agentProtocolNames != null && !agentProtocolNames.contains(protocol.getName())) {
```
oleg-nenashev: 🐛 agentProtocolNames **will** be empty if X-Jenkins-Agent-Protocols is missing since you initialize the container before if clause. In such case this code will also skip protocols

oleg-nenashev:
Finally there is a 🐛 from what I see.

oleg-nenashev:
Change in Jenkins core is unlikely a subject for backporting

src/main/java/hudson/remoting/Engine.java
stephenc: Because StringUtils is not on the classpath in remoting.jar (which is supposed to be self-contained)

oleg-nenashev:
🐝

PR #95, opened 2016-08-03
[FIXED JENKINS-37140] Roll the dice again if we are unlucky

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-37140

@reviewbybees


oleg-nenashev:
🐝 since I know the context

oleg-nenashev:
So we need to get it in .3

stephenc:
I'm merging this, you can sort out your backporting ;-)

PR #97, opened 2016-08-04
[JENKINS-22853] - Be robust against delayed EOF command when unexporting streams

oleg-nenashev:
In particular cases the objects can be unexported before the EOF command arrives from another channel side.  In my case it happens, because <code>UserRequest#releaseExports</code> already does it. In such case remoting prints horrible exceptions to the log. So we actually want not to unexport items in EOF Command if we're late.

Changes:
* [x] Improve error propagation in the code (no IllegalStateExceptions anymore)
* [x] All 3 EOF commands do not fail if there is no stream object. Unexport commands work as before
* [ ] TBD - Close closeable items when we release them in <code>ExportTable</code>

https://issues.jenkins-ci.org/browse/JENKINS-22853

@reviewbybees (esp. @stephenc)


oleg-nenashev:
PR Builder fails with "[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 7 iota=14", lol . Just another issue to tackle

src/main/java/hudson/remoting/InvalidObjectIdException.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package hudson.remoting;
+
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Exception for Invalid object processing in {@link ExportTable}
+ * @author Oleg Nenashev
+ */
+public class InvalidObjectIdException extends ExecutionException {
```
oleg-nenashev: I would use RemotingSystemException, but it's Runtime one

stephenc:
:bee:

oleg-nenashev:
I'll likely have to replace InvalidObjectIdException by just ExecutionException. I nest the exception in particular cases, and it gets serialized over the channel from agent to master. Classloading in this direction does not work, hence we get a ClassNotFound exception if agent remoting version is greater than the core's one

Self-:bug:

oleg-nenashev:
@stephenc need re-bee

stephenc:
:bee:

oleg-nenashev:
@reviewbybees done

PR #99, opened 2016-08-10
[NO_MERGE][JENKINS-23271] - Allow exporting objects with disabled automatic unexporting

oleg-nenashev:
Particular remoting calls may cause asynchronous operations on Jenkins agents.

E.g. RemoteProcess, which is being created by hudson.Launcher. In such case we export object of the remote process in the ExportTable, and then create an invocation handler, which is being serialized back to the master. After that the local instance of the wrapped object is not being used anymore, hence it can be garbage collected along with the exported object instance. This garbage collection causes unexported of the object.

According to my experiments, on Java8 such garbage collection of RemoteInvocationHandler frequently happens between <code>RemoteProсess#start()</code> and <code>RemoteProcess#join()</code> calls when running many short operations on the agent (e.g. sh steps), which leads to JENKINS-23271.

So I decided to introduce new API, which prohibits automatic unexport for such async calls. It should be mostly safe since RemoteProcess calls commonly end automatically or get terminated by external side. But additional feedback would be appreciated.

https://issues.jenkins-ci.org/browse/JENKINS-23271

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -589,6 +589,23 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     public <T> T export(Class<T> type, T instance) {
         return export(type, instance, true);
     }
+    
+    /**
+     * Exports object, which involves asynchronous operation.
+     * Such objects may be returned and serialized back to the channel before the real operation completes.
+     * In such case we have to disable automatic unexport, because otherwise the original object
+     * on our side may be deallocated by garbage Collector, which will trigger unexport from the {@link ExportTable}.
+     * <p/>
+     * This method should be invoked if and only if the call presumes async calls.
+     * For example Jenkins' {@code RemoteLauncher} starts the process and then calls {@code join()} to monitor it.
+     * In such case usage of this method is valid, because otherwise the {@code RemoteInvocationHandler} for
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Bug investigation braindump is here: https://docs.google.com/document/d/1nu7jD2bNqEzD7kOM1R0VUpExycPNUq8wOTTLrr9xvOY/edit

stephenc:
no tests?

feiniao0308:
When will this change submit? I'm waiting for it :-). Thanks!

oleg-nenashev:
@feiniao0308 This change is not going to fly in such way. It works, but actually it just suppresses the root cause of the problem with Garbage collector.
I'm going to propose another fix later.

feiniao0308:
@oleg-nenashev Thanks a lot, then i'll wait for the fix :-)

jglick:
> I'm going to propose another fix later.

So, should this PR be marked a werk-in-progress (or even closed)?

oleg-nenashev:
@jglick sorry, fixed

feiniao0308:
When will another fix came out?

oleg-nenashev:
When I find time to fix it :(

oleg-nenashev:
Closing in favor of https://github.com/jenkinsci/jenkins/pull/2635

PR #100, opened 2016-08-18
[JENKINS-25218] - Hardening of FifoBuffer operation logic

oleg-nenashev:
The most of [JENKINS-25218](https://issues.jenkins-ci.org/browse/JENKINS-25218) has been handled in https://github.com/jenkinsci/remoting/commit/3ea38c8b95a8d4ca8e245fc0999d7e4a9a6b1424 by @andresrc. But there are still potential issues in the FifoBuffer behavior.

Changes:
- [x] - Add closedRequested and handle it in multiple threads, which used to prevent <code>close()</code> calls previously
- [x] Fix potential NPE in FifoBuffer receiver
- [x] Do not rely on external code when we receive <code>ClosedChannelException</code> from the underlying channel

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
+++ b/src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
@@ -195,6 +195,8 @@ public class FifoBuffer implements Closeable {
     public void setLimit(int newLimit) {
         synchronized (lock) {
             limit = newLimit;
+            // We resized the buffer, hence read/write threads may be able to proceed
+            lock.notifyAll();
         }
     }
 
@@ -363,12 +365,15 @@ public class FifoBuffer implements Closeable {
         while (len>0) {
             int chunk;
 
+            final boolean shouldWaitToCleanTheBuffer;
             synchronized (lock) {
+                if (closeRequested) {
+                    handleCloseRequest();
+                    throw new IOException("closed during write operation");
+                }
+                
                 while ((chunk = Math.min(len,writable()))==0) {
-                    if (closeRequested) {
-                        handleCloseRequest();
-                        throw new IOException("closed during write operation");
-                    }
+                    
                     lock.wait(100);
                 }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
```diff
@@ -339,11 +369,14 @@ public void write(byte[] buf, int start, int len) throws InterruptedException, I
...
             int chunk;
 
-            synchronized (lock) {
+            final boolean shouldWaitToCleanTheBuffer;
+            synchronized (lock) { 
                 while ((chunk = Math.min(len,writable()))==0) {
-                    if (closed)
-                        throw new IOException("closed during write operation");
-
+                    if (closeRequested) {
```
oleg-nenashev: I'm still not comfortable about not having timeout if the buffer is full && does not get freed by other threads

jglick:
:bee: without much comprehension. I will assume this has gotten some interactive testing at least?

jglick:
From the description, it sounds like there ought to be a way to write a test demonstrating a fix of the known bug condition.

oleg-nenashev:
It's being soaked for ~4 days now. Seems to be stable on frequent data interactions.

Regarding tests, I can likely reproduce https://github.com/jenkinsci/remoting/pull/100/commits/b6e71fa34bbb5e533b6b2991246fa77649bcc01b, but it's pretty obvious. Other changes are related to some parallel code fun, so I do not see a way to reproduce them in tests without PowerMock bytecode magic.

Since it's just hardening, maybe we do not want to have it in the stable branch (excepting NPE fix)

oleg-nenashev:
I decided to merge this PR towards the 3.0 branch, which will have some stabilization interval in weekly releases

stephenc:
:bee: AIUI

oleg-nenashev:
I will merge it into 3.2 since we need a `stable` hotfix for JENKINS-39596 in 3.1

oleg-nenashev:
The PR has been soak tested enough. 2.62.4 is unlikely going into Jenkins LTS, so I will just merge it

PR #103, opened 2016-08-20
[JENKINS-37565] - Upgrade source level to Java 7

oleg-nenashev:
This change does not update the code to the Java7 style, but we will do it later. It requires https://github.com/jenkinsci/pom/pull/11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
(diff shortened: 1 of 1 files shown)


jglick:
You need to `deploy` the snapshot.

jglick:
:bee: if that is done (though upstream I suggested `7` be the default, so merely updating the parent would suffice)

oleg-nenashev:
@jglick No. I need reviews of https://github.com/jenkinsci/pom/pull/11 ;)

oleg-nenashev:
@reviewbybees ready to go forward

PR #104, opened 2016-08-20
[CID-152200,CID-152202] - Resource leak in Cipher I/O streams on exceptional paths

oleg-nenashev:
* https://scan8.coverity.com/reports.htm#v14462/p10499/fileInstanceId=12836377&defectInstanceId=4427377&mergedDefectId=152202
* https://scan8.coverity.com/reports.htm#v14462/p10499/fileInstanceId=12836377&defectInstanceId=4427377&mergedDefectId=152200

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
@@ -107,7 +107,7 @@ public abstract class JnlpServer3Handshake extends JnlpServerHandshake {
             // Something went wrong. We want to gracefully close Cipher stream in the case they were open
             EngineUtil.closeAndLogFailures(in, LOGGER, "CipherInputStream", socket.toString());
             EngineUtil.closeAndLogFailures(out, LOGGER, "CipherOutputStream", socket.toString());
-            if (ex instanceof RuntimeException) {
+            if (ex instanceof IOException) {
                 throw (IOException) ex;
             } else {
                 throw new IOException("Unexpected runtime exception during creation of the channel", ex);
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
@@ -95,11 +95,25 @@ public abstract class JnlpServer3Handshake extends JnlpServerHandshake {
                 Jnlp3Util.keyFromString(aesKeyString),
                 Jnlp3Util.keyFromString(specKeyString));
 
-        Channel channel = createChannelBuilder(nodeName).build(
-                new CipherInputStream(SocketChannelStream.in(socket),
-                        channelCiphers.getDecryptCipher()),
-                new CipherOutputStream(SocketChannelStream.out(socket),
-                        channelCiphers.getEncryptCipher()));
+
+        CipherInputStream in = null;
+        CipherOutputStream out = null;
+        final Channel channel;
+        try {
+            in = new CipherInputStream(SocketChannelStream.in(socket), channelCiphers.getDecryptCipher());
+            out = new CipherOutputStream(SocketChannelStream.out(socket), channelCiphers.getEncryptCipher());
+            channel = createChannelBuilder(nodeName).build(in, out);
+        } catch (Exception ex) {
+            // Something went wrong. We want to gracefully close Cipher stream in the case they were open
+            EngineUtil.closeAndLogFailures(in, LOGGER, "CipherInputStream", socket.toString());
+            EngineUtil.closeAndLogFailures(out, LOGGER, "CipherOutputStream", socket.toString());
+            if (ex instanceof RuntimeException) {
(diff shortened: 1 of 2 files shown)


oleg-nenashev:

src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
```diff
@@ -95,11 +95,25 @@ public Channel connect() throws IOException, InterruptedException {
...
+        try {
+            in = new CipherInputStream(SocketChannelStream.in(socket), channelCiphers.getDecryptCipher());
+            out = new CipherOutputStream(SocketChannelStream.out(socket), channelCiphers.getEncryptCipher());
+            channel = createChannelBuilder(nodeName).build(in, out);
+        } catch (Exception ex) {
+            // Something went wrong. We want to gracefully close Cipher stream in the case they were open
+            EngineUtil.closeAndLogFailures(in, LOGGER, "CipherInputStream", socket.toString());
+            EngineUtil.closeAndLogFailures(out, LOGGER, "CipherOutputStream", socket.toString());
+            if (ex instanceof RuntimeException) {
+                throw (IOException) ex;
```
olivergondza: The cast seems unsafe.

oleg-nenashev: It is unsafe... I've messed up something

oleg-nenashev:
@olivergondza fixed

src/main/java/org/jenkinsci/remoting/engine/JnlpServer3Handshake.java
```diff
@@ -95,11 +95,25 @@ public Channel connect() throws IOException, InterruptedException {
...
+        CipherOutputStream out = null;
+        final Channel channel;
+        try {
+            in = new CipherInputStream(SocketChannelStream.in(socket), channelCiphers.getDecryptCipher());
+            out = new CipherOutputStream(SocketChannelStream.out(socket), channelCiphers.getEncryptCipher());
+            channel = createChannelBuilder(nodeName).build(in, out);
+        } catch (Exception ex) {
+            // Something went wrong. We want to gracefully close Cipher stream in the case they were open
+            EngineUtil.closeAndLogFailures(in, LOGGER, "CipherInputStream", socket.toString());
+            EngineUtil.closeAndLogFailures(out, LOGGER, "CipherOutputStream", socket.toString());
```
jglick: :ant: better (under Java 7) to use `Throwable.addSuppressed`. Or simpler still:

```java
Channel channel = null;
CipherInputStream in = new CipherInputStream(SocketChannelStream.in(socket), channelCiphers.getDecryptCipher());
try {
    CipherOutputStream out = new CipherOutputStream(SocketChannelStream.out(socket), channelCiphers.getEncryptCipher());
    try {
        return channel = createChannelBuilder(nodeName).build(in, out);
    } finally {
        if (channel == null) {
            out.close();
        }
    }
} finally {
    if (channel == null) {
        in.close();
    }
}
```

jglick: Even better would be a utility (should I file a suggestion for Guava?):

```java
public static class ClosePartials implements AutoCloseable {
    private final List<AutoCloseable> things = new ArrayList<>();
    public ClosePartials() {}
    public <T extends AutoCloseable> T with(T thing) {
        things.add(thing);
        return thing;
    }
    /** @param <T> recommended to also be {@link AutoCloseable} */
    public <T> T done(T thing) {
        things.clear();
        return thing;
    }
    @Override
    public void close() throws Exception {
        Exception x = null;
        for (AutoCloseable thing : things) {
            try {
                thing.close();
            } catch (Throwable t) {
                if (x == null) {
                    if (t instanceof Exception) {
                        x = (Exception) t;
                    } else {
                        x = new Exception(t);
                    }
                } else {
                    x.addSuppressed(t);
                }
            }
        }
        if (x != null) {
            throw x;
        }
    }
}
```

Example usage:

```java
public static Channel example(Socket s, Cipher c, ChannelBuilder cb) throws Exception {
    try (ClosePartials cp = new ClosePartials();
         InputStream sin = cp.with(SocketChannelStream.in(s));
         CipherInputStream cin = cp.with(new CipherInputStream(sin, c));
         OutputStream sout = cp.with(SocketChannelStream.out(s));
         CipherOutputStream cout = cp.with(new CipherOutputStream(sout, c))) {
        return cp.done(cb.build(cin, cout));
    }
}
```

In general, methods which deliberately create closeable _[sic]_ objects but then conceal them as references inside some higher-level closeable object are risky, because even if the ultimate caller is careful to `close` the final object, its `close` method might not be foolproof. `Channel.close` does not even pretend to guarantee that all streams it was originally passed will be closed by the time the method completes.

jglick:
Anyway, :bee: as it stands.

oleg-nenashev:
All these PRs go against stable-2.x, which will stay on Java6

oleg-nenashev:
@reviewbybees done

PR #105, opened 2016-08-20
[CID-48110] - Resource leak on exceptional path in Util#makeResource()

oleg-nenashev:
It may happen in the case of the remote classloader.
https://scan8.coverity.com/reports.htm#v14462/p10499/fileInstanceId=12836386&defectInstanceId=4427346&mergedDefectId=48110

@jenkinsci/code-reviewers

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -60,8 +60,11 @@ class Util {
         resource.getParentFile().mkdirs();
 
         FileOutputStream fos = new FileOutputStream(resource);
-        fos.write(image);
-        fos.close();
+        try {
+            fos.write(image);
+        } finally {
+            fos.close();
+        }
 
         deleteDirectoryOnExit(tmpFile);
 


src/main/java/hudson/remoting/Util.java
```diff
@@ -60,8 +60,11 @@ static File makeResource(String name, byte[] image) throws IOException {
         resource.getParentFile().mkdirs();
 
         FileOutputStream fos = new FileOutputStream(resource);
-        fos.write(image);
-        fos.close();
+        try {
+            fos.write(image);
+        } finally {
```
jglick: :ant: simpler to use `try`-with-resources in `-target 7`

jglick:
:bee:

jglick:
BTW is FindBugs not getting these? Seems like standard FB fare.

oleg-nenashev:
@jglick It does

oleg-nenashev:
I just used CID since it allows referencing something for tracking

src/main/java/hudson/remoting/Util.java
oleg-nenashev: Sure. But the PR is about Java 6. Upgrade to Java 7 is planned to remoting 3. See #103

daniel-beck:
🐝

oleg-nenashev:
@reviewbybees done

PR #106, opened 2016-09-19
[FIXED JENKINS-38344] Catch uncaught exceptions, log and close connection

stephenc:
See [JENKINS-38344](https://issues.jenkins-ci.org/browse/JENKINS-38344)

@jenkinsci/code-reviewers @reviewbybees (esp @Vlatombe who will have to do some reverting and reapplication to keep internal codebase synced 😉 )

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -701,6 +701,9 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
                     record.setParameters(new Object[]{workerThread.getName(), listener, e.getClass().getSimpleName()});
                     LOGGER.log(record);
                 }
+                if (e instanceof Error) {
+                    throw (Error)e;
+                }
             } finally {
                 workerThread.setName(oldName);
             }
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -50,6 +50,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.AbstractQueuedSynchronizer;
 import java.util.logging.Level;
+import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import javax.annotation.Nonnull;
 import javax.annotation.OverridingMethodsMustInvokeSuper;
@@ -693,6 +694,13 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
                         (ops & SelectionKey.OP_CONNECT) == SelectionKey.OP_CONNECT,
                         (ops & SelectionKey.OP_READ) == SelectionKey.OP_READ,
                         (ops & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE);
+            } catch (Throwable e) {
+                if (LOGGER.isLoggable(Level.SEVERE)) {
+                    LogRecord record = new LogRecord(Level.SEVERE, "[{0}] Listener {1} propagated an uncaught {2}");
+                    record.setThrown(e);
+                    record.setParameters(new Object[]{workerThread.getName(), listener, e.getClass().getSimpleName()});
+                    LOGGER.log(record);
+                }
             } finally {
                 workerThread.setName(oldName);
             }
--- a/src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
@@ -23,6 +23,7 @@
  */
 package org.jenkinsci.remoting.protocol.impl;
 
+import java.util.logging.LogRecord;
 import javax.annotation.Nonnull;
 import java.io.EOFException;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


oleg-nenashev:
Will also require backporting to the stable branch

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
```diff
@@ -248,10 +249,23 @@ public void run() {
                         } catch (ClosedChannelException e) {
                             onRecvClosed();
                             return;
-                        } catch (EOFException e) {
+                        } catch (IOException e) {
+                            if (LOGGER.isLoggable(Level.FINER)) {
+                                // will likely be reported elsewhere, so we just trace this at FINER
```
oleg-nenashev: I'm not 100% sure, but OK

andresrc:
:bee:

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
```diff
@@ -273,6 +287,13 @@ public void run() {
                 } finally {
                     release(buffer);
                 }
+            } catch (Throwable e) {
+                if (LOGGER.isLoggable(Level.SEVERE)) {
```
oleg-nenashev: likely  <code>if (true)</code> for sane use-cases

stephenc:
@oleg-nenashev really? did you merge the JNLP4 stuff to stable?

stephenc:
@reviewbybees done

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
```diff
@@ -273,6 +287,13 @@ public void run() {
                 } finally {
                     release(buffer);
                 }
+            } catch (Throwable e) {
```
oleg-nenashev: 🐛 needs rethrow after logging. Otherwise propagation of serious issues like OOM may be suppressed

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -693,6 +694,13 @@ public void run() {
                         (ops & SelectionKey.OP_CONNECT) == SelectionKey.OP_CONNECT,
                         (ops & SelectionKey.OP_READ) == SelectionKey.OP_READ,
                         (ops & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE);
+            } catch (Throwable e) {
+                if (LOGGER.isLoggable(Level.SEVERE)) {
+                    LogRecord record = new LogRecord(Level.SEVERE, "[{0}] Listener {1} propagated an uncaught {2}");
+                    record.setThrown(e);
+                    record.setParameters(new Object[]{workerThread.getName(), listener, e.getClass().getSimpleName()});
+                    LOGGER.log(record);
+                }
```
oleg-nenashev: 🐛 rethrow. Otherwise it suppresses <code>Error</code>s, which are supposed to be fatal here

oleg-nenashev:
> did you merge the JNLP4 stuff to stable?

Nope. Just mean that there should be similar patterns

stephenc:
> Just mean that there should be similar patterns

May/may not... perhaps already covered. This was just a gap in my new code that I found

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
stephenc: We're not exactly on the critical path here... and if we decide to change the level it makes it easier to change it later

oleg-nenashev:
just a timeout

oleg-nenashev:
:bee: if the build passes

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
```diff
@@ -248,10 +249,23 @@ public void run() {
                         } catch (ClosedChannelException e) {
                             onRecvClosed();
                             return;
-                        } catch (EOFException e) {
+                        } catch (IOException e) {
```
jtnord: Not sure you want to catch `java.net.SocketTimeoutException` which would be thrown for a read timeout but is non fatal...  [javadoc link](https://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#setSoTimeout(int))

Not sure if the timeout is ever set here - but still - was the cause of a previous bug in remoting...

jtnord: I think we may have a timeout set [here](https://github.com/jenkinsci/remoting/blob/a4f5af3617277b53c0b350d64423b7b7009bfc8c/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpoint.java#L183) - so this would introduce a regression IIUC.

jtnord:
I think this may introduce a regression if the read from the remote side does not complete within the specified socket timeout.

src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
stephenc: We were catching the `IOException` anyway... but I agree that we should be just looping for a `SocketTimeoutException`

Likely not an issue for most cases as we use the NIOEngine exclusively for most code paths... but yes it would be a good idea to fix anyway

stephenc: https://issues.jenkins-ci.org/browse/JENKINS-38541 and #111

PR #107, opened 2016-09-26
Merge stable-2.x into master

oleg-nenashev:
Integrates pending changes from the stable branch

### DIFF
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -212,8 +212,6 @@ public class Main {
                         } catch (IOException e) {
                             LOGGER.log(Level.WARNING, "Could not read certificate from " + file, e);
                             continue;
-                        } finally {
-                            IOUtils.closeQuietly(fis);
                         }
                     } else {
                         if (file.isFile()) {
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -191,14 +191,18 @@ public class Main {
                     if (file.isFile()
                             && (length = file.length()) < 65536
                             && length > "-----BEGIN CERTIFICATE-----\n-----END CERTIFICATE-----".length()) {
-                        FileInputStream fis = null;
                         try {
                             // we do basic size validation, if there are x509 certificates that have a PEM encoding
                             // larger
                             // than 64kb we can revisit the upper bound.
                             cert = new byte[(int) length];
-                            fis = new FileInputStream(file);
-                            int read = fis.read(cert);
+                            FileInputStream fis = new FileInputStream(file);
+                            final int read;
+                            try {
+                                read = fis.read(cert);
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -191,14 +191,18 @@ public Engine createEngine() {
...
                         try {
                             // we do basic size validation, if there are x509 certificates that have a PEM encoding
                             // larger
                             // than 64kb we can revisit the upper bound.
                             cert = new byte[(int) length];
-                            fis = new FileInputStream(file);
-                            int read = fis.read(cert);
+                            FileInputStream fis = new FileInputStream(file);
+                            final int read;
+                            try {
```
oleg-nenashev: Can be reworked later since we're on Java 7 now

oleg-nenashev:
@reviewbybees @stephenc

src/main/java/org/jenkinsci/remoting/engine/EngineUtil.java
```diff
@@ -91,4 +96,24 @@ protected static Properties readResponseHeaders(InputStream inputStream) throws
...
+
+    /**
+      * Closes the item and logs error to the log in the case of error.
+      * Logging will be performed on the {@code WARNING} level.
+      * @param toClose Item to close. Nothing will happen if it is {@code null}
+      * @param logger Logger, which receives the error
+      * @param closeableName Name of the closeable item
+      * @param closeableOwner String representation of the closeable holder
+      */
+    static void closeAndLogFailures(@CheckForNull Closeable toClose, @Nonnull Logger logger,
```
stephenc: is this used?

oleg-nenashev: Seems "no", the merge process removed the usage. But it will be definitely used when I start cleaning up FindBugs

oleg-nenashev: So I want to keep it if there os no strong -1s

src/main/java/hudson/remoting/jnlp/Main.java
svanoort: Reworked using try with reseources?

src/main/java/org/jenkinsci/remoting/engine/EngineUtil.java
jglick: :ant:; this code idiom was never a good idea in Java 6- IMO (better to just let any hypothetical exception from `close` be thrown up the stack), and is completely unnecessary in Java 7+, since `try`-with-resources calls `addSuppressed` when necessary. So delete this unused method, and when fixing FB warnings, just use the modern `try` style and you are fine.

oleg-nenashev:
The method is still useful for particular cases, but I'll remove it in order to unblock the PR and further changes

oleg-nenashev:
Merging as per previous feedback

PR #108, opened 2016-09-26
[JENKINS-37565] - Really allow compiling with Java 7

oleg-nenashev:
Just a leftover fix

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -405,7 +405,7 @@ THE SOFTWARE.
         <configuration>
           <signature>
             <groupId>org.codehaus.mojo.signature</groupId>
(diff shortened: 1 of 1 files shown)


stephenc:
:bee:

jglick:
:bee: though should this not be picked up automatically from the parent POM when you set the right `java.level`?

oleg-nenashev:
> though should this not be picked up automatically from the parent POM when you set the right java.level?

It should, but https://github.com/jenkinsci/pom/blob/master/pom.xml does not include Animal Sniffer at all. Maybe needs a fix

oleg-nenashev:
@reviewbybees done

PR #110, opened 2016-09-27
[FIXED JENKINS-38539] Turn on SO_KEEPALIVE and provide CLI option to turn it off again

stephenc:
See [JENKINS-38539](https://issues.jenkins-ci.org/browse/JENKINS-38539)

Will need to be merged up to master also

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -190,7 +190,7 @@ public class Launcher {
     public boolean noReconnect = false;
 
     @Option(name = "-noKeepAlive",
-            usage = "Do not open the socket to the master with SO_KEEPALIVE enabled")
+            usage = "Disable TCP socket keep alive on connection to the master.")
     public boolean noKeepAlive = false;
 
     public static void main(String... args) throws Exception {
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/Launcher.java
```diff
@@ -189,6 +189,10 @@ public boolean verify(String s, SSLSession sslSession) {
     @Option(name="-noReconnect",usage="Doesn't try to reconnect when a communication fail, and exit instead")
     public boolean noReconnect = false;
 
+    @Option(name = "-noKeepAlive",
+            usage = "Do not open the socket to the master with SO_KEEPALIVE enabled")
```
olivergondza: This does not tell me much as a user. Why would I want that?

stephenc: I would argue that you do not want this unless you know why you do not want it... I could make it a system property rather than a CLI option... should only be required if we have a fundamental misunderstanding of how sockets work...

oleg-nenashev: I would also rephrase it a bit

rsandell:
:bee:

batmat:
:bee:

oleg-nenashev:
@reviewbybees done

src/main/java/hudson/remoting/Launcher.java
oleg-nenashev: @olivergondza My plan is to create a documentation page, which will clarify options and their potential impact. By now I think the current summary string is acceptable. Do you agree?

oleg-nenashev:
It's in the plan for the next release and hopefully 2.19.2

src/main/java/hudson/remoting/Launcher.java
olivergondza: Ack

PR #114, opened 2016-10-03
Add a simple Jenkinsfile for building on ci.jenkins.io

rtyler:
This should be able to be replaced at some point in the (hopefully) near future
by some steps in our shared library
(https://github.com/jenkins-infra/pipeline-library) but that's not quite ready
yet

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,34 @@
+#!/usr/bin/env groovy
+
+/* Only keep the 10 most recent builds. */
+properties([[$class: 'BuildDiscarderProperty',
+                strategy: [$class: 'LogRotator', numToKeepStr: '10']]])
+
+
+node('java') {
+    timestamps {
+        stage('Checkout') {
+            checkout scm
+        }
+
+        stage('Build') {
+            withEnv([
+                "JAVA_HOME=${tool 'jdk8'}",
+                "PATH+MVN=${tool 'mvn'}/bin",
+                'PATH+JDK=$JAVA_HOME/bin',
+            ]) {
+                timeout(30) {
+                    sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+                }
+            }
+        }
+
+        stage('Archive') {
+            /* Archive the test results */
+            junit '**/target/surefire-reports/TEST-*.xml'
+
+            /* Archive the build artifacts */
+            archiveArtifacts artifacts: 'target/*.hpi,target/*.jpi'
+        }
+    }
+}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,13 @@ for (int i = 0; i < platforms.size(); ++i) {
                         'PATH+JDK=$JAVA_HOME/bin',
                     ]) {
                         timeout(30) {
-                            sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+                            String command = 'mvn clean install -Dmaven.test.failure.ignore=true'
+                            if (isUnix()) {
+                                sh command
+                            }
+                            else {
+                                bat command
+                            }
                         }
                     }
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,7 @@ node('java') {
             junit '**/target/surefire-reports/TEST-*.xml'
 
             /* Archive the build artifacts */
-            archiveArtifacts artifacts: 'target/*.hpi,target/*.jpi'
+            archiveArtifacts artifacts: 'target/**/*.jar'
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,30 +5,44 @@ properties([[$class: 'BuildDiscarderProperty',
                 strategy: [$class: 'LogRotator', numToKeepStr: '10']]])
 
 
-node('java') {
-    timestamps {
-        stage('Checkout') {
-            checkout scm
-        }
+/* These platforms correspond to labels in ci.jenkins.io, see:
+ *  https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
+ */
+List platforms = ['linux', 'windows']
+Map branches = [:]
 
-        stage('Build') {
-            withEnv([
-                "JAVA_HOME=${tool 'jdk8'}",
-                "PATH+MVN=${tool 'mvn'}/bin",
-                'PATH+JDK=$JAVA_HOME/bin',
-            ]) {
-                timeout(30) {
-                    sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+for (int i = 0; i < platforms.size(); ++i) {
+    String label = platforms[i]
+    branches[label] = {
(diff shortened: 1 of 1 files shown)


rtyler:
@oleg-nenashev I believe the failure in jenkins.ci.c.c is unrelated, but let me know if that's not the case and I can dig further

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        stage('Build') {
+            withEnv([
+                "JAVA_HOME=${tool 'jdk8'}",
+                "PATH+MVN=${tool 'mvn'}/bin",
+                'PATH+JDK=$JAVA_HOME/bin',
+            ]) {
+                timeout(30) {
+                    sh 'mvn clean install -Dmaven.test.failure.ignore=true'
```
oleg-nenashev: Why do you ignore the test failure here?

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
...
+            checkout scm
+        }
+
+        stage('Build') {
+            withEnv([
+                "JAVA_HOME=${tool 'jdk8'}",
+                "PATH+MVN=${tool 'mvn'}/bin",
+                'PATH+JDK=$JAVA_HOME/bin',
+            ]) {
+                timeout(30) {
```
oleg-nenashev: I would set a 60-min timeout just in case the build queue is packed.

Jenkinsfile
oleg-nenashev: Ah, just because we do not want the build to fail...

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
...
+                    sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+                }
+            }
+        }
+
+        stage('Archive') {
+            /* Archive the test results */
+            junit '**/target/surefire-reports/TEST-*.xml'
```
oleg-nenashev: Should be the last invocation, which fails the build in the case of test errors

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env groovy
```
oleg-nenashev: Not required anymore

oleg-nenashev:
Build failure is unrelated. CC @stephenc, who has written this test

Jenkinsfile
rtyler: 30 minutes is the existing timeout value in the freestyle job

Jenkinsfile
rtyler: I add this out of habit to ensure the editor knows how to syntax highlight the document.

Jenkinsfile
oleg-nenashev: Which times out sometimes

Jenkinsfile
rtyler: This does not fail the Pipeline if there are test failures, it just marks the build as unstable. The `archiveArtifacts` step will still execute (as far as I am aware)

Jenkinsfile
rtyler: Is that not the correct behavior? Taking more than 30 minutes sounds like a problem with tests taking a long time. If this is going to spin for 60 minutes and still fail then that's costing us double the amount of compute $$$ :frowning:

Jenkinsfile
oleg-nenashev: OK

Jenkinsfile
oleg-nenashev: OK, let's keep it as is. At least it does not include the queue wait time

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
...
+
+
+node('java') {
+    timestamps {
+        stage('Checkout') {
+            checkout scm
+        }
+
+        stage('Build') {
+            withEnv([
```
stephenc: isn't `withMaven` supposed to set all this up for you (with the advantage of knowing about the settings) (not that I like the `withMaven` step... but still)

stephenc:
LGTM :bee:

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env groovy
+
+/* Only keep the 10 most recent builds. */
+properties([[$class: 'BuildDiscarderProperty',
+                strategy: [$class: 'LogRotator', numToKeepStr: '10']]])
+
+
+node('java') {
```
jtnord: windows... windows windows...

Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+                }
+
+                stage('Build') {
+                    withEnv([
+                        "JAVA_HOME=${tool 'jdk8'}",
+                        "PATH+MVN=${tool 'mvn'}/bin",
+                        'PATH+JDK=$JAVA_HOME/bin',
+                    ]) {
+                        timeout(30) {
+                            sh 'mvn clean install -Dmaven.test.failure.ignore=true'
```
jtnord: sh - on windows?

Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+                    checkout scm
+                }
+
+                stage('Build') {
+                    withEnv([
+                        "JAVA_HOME=${tool 'jdk8'}",
+                        "PATH+MVN=${tool 'mvn'}/bin",
+                        'PATH+JDK=$JAVA_HOME/bin',
+                    ]) {
+                        timeout(30) {
```
jtnord: timeout is too short to be usefull when tests fail (due to retrying the tests multiple times).

Takes 34ish minutes on my machine with gen6 i7 and SSD :-/

Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+                            sh 'mvn clean install -Dmaven.test.failure.ignore=true'
+                        }
+                    }
+                }
+
+                stage('Archive') {
+                    /* Archive the test results */
+                    junit '**/target/surefire-reports/TEST-*.xml'
```
jtnord: not that I care at this point (I wanta `Jenkinsfile` that builds on windows slaves) but what about findbugs?

Jenkinsfile
rtyler: Oh right, I forgot. It's still maddening that `bat` and `sh` are different steps

Jenkinsfile
rtyler: a little late to [this discussion](https://github.com/jenkinsci/remoting/pull/114?#discussion_r81625991)

Jenkinsfile
rtyler: not in this PR

jtnord:
@rtyler @oleg-nenashev ship-it please!!

Jenkinsfile
jtnord: Yes would be nice to have something akin to xshell support for pipeline (running a durable task...)

rtyler:
[SHAZAM](https://ci.jenkins.io/job/Core/job/remoting/)

PR #115, opened 2016-10-06
[JENKINS-38690] do not cause a IOHub Thread Storm

jtnord:
[JENKINS-38690](https://issues.jenkins-ci.org/browse/JENKINS-38690)
We should only add the READ interested OPS when we have fully drained the
read queue.

In the case we had a lot of data to read (where lots >> 8192 bytes) we
would read a buffers worth of data add READ to the interested opps then
read another buffer in the same thread.  If we where reading 81920 bytes
for example this would be 9 calls to add READ OPs - which would run on the
selctor thread and cause an immediate wakeup to be called followed by a
new onReady().
However the onReady will be added as a defered callback and the ops
removed, but the read thread is still going and as it takes time will call
back again to add interested ops... which will wake the selector see that
READ ops are valid and cause a new deffered callback...

So do not add read interested ops inside the loop, add it only when we
have drained the buffer.

@reviewbybees esp @stephenc

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
@@ -147,12 +147,12 @@ public class NIONetworkLayer extends NetworkLayer implements IOHubReadyListener 
                             while (recv.hasRemaining()) {
                                 onRead(recv);
                             }
-                            if (in.isOpen() && recvKey.isValid()) {
-                                getIoHub().addInterestRead(recvKey);
-                            } else {
-                                recvKey.cancel();
-                                onRecvClosed();
-                            }
+                        }
+                        if (in.isOpen() && recvKey.isValid()) {
+                            getIoHub().addInterestRead(recvKey);
+                        } else {
+                            recvKey.cancel();
+                            onRecvClosed();
                         }
                     } catch (ClosedChannelException e) {
                         recvKey.cancel();


jtnord:
**Note**:  still running the benchmark on this (tests fail on windows - so need the PR to verify the status of the build)

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
```diff
@@ -147,12 +147,12 @@ public void ready(boolean accept, boolean connect, boolean read, boolean write)
...
                                 onRead(recv);
                             }
-                            if (in.isOpen() && recvKey.isValid()) {
-                                getIoHub().addInterestRead(recvKey);
-                            } else {
-                                recvKey.cancel();
-                                onRecvClosed();
-                            }
+                        }
+                        if (in.isOpen() && recvKey.isValid()) {
```
stephenc:
```
if (count != -1) {
  ...
}
```

and hoist the scope of count outside the while loop... otherwise a count of -1 will trigger two closed events

jtnord: Hmmm... Why does the current [`count == -1`](https://github.com/jenkinsci/remoting/pull/115/files#diff-5bd66617da2d4d7a7625a042f0297afaR136) check not cancel the `recvKey`?

jtnord:
[Stress test results](https://gist.github.com/jtnord/cd82133ecdbf573abde09ea8673cce90)
Fix uses less threads (expected) and is uses less CPU - however uses much more RAM.

oleg-nenashev:
Uhh, not the first such failure.
/me hopes jenkins.io deployment finishes soon. Can do it on my own if @rtyler grants me permissions there

oleg-nenashev:
Yes, should be an INFRA ticket

oleg-nenashev:
https://issues.jenkins-ci.org/browse/INFRA-946

stephenc:
After one load test run (`--warmup 60 --collect 90 --connect JNLP4-plaintext`)

OS-X java 1.8.0_102

CPU before: 1.53
CPU after: 1.48

Avg Mem before: 195592±43263
Avg Mem after: 199295±40312

stephenc:
Statistically speaking, no significant difference in either memory or CPU usage

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
stephenc: I think because -1 means the channel is closed and the key has been cancelled already... but that is just a guess

oleg-nenashev:
Would be happy if you follow the reviewbybees policy next time

oleg-nenashev:
It looks good to me, but I do not feel good about releasing it without some soak testing.
Remoting 3 release will be delayed for a week.

PR #116, opened 2016-10-06
[JENKINS-38692] Add extra logging to help diagnosing Thread spike.

jtnord:
[JENKINS-38692](https://issues.jenkins-ci.org/browse/JENKINS-38692) Add extra logging statements to try and get to the root cause
of the IOHub thread spike.

@reviewbybees esp @stephenc

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -292,6 +292,10 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      * @param key the key.
      */
     public final void addInterestRead(SelectionKey key) {
+        if (LOGGER.isLoggable(Level.FINEST)) {
+            // TODO probably want some more info about the key here...
+            LOGGER.log(Level.FINEST, "Scheduling adding OP_READ to {0}", key);
+        }
         interestOps.add(new InterestOps(key, SelectionKey.OP_READ, 0));
         selector.wakeup();
     }
@@ -302,6 +306,10 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      * @param key the key.
      */
     public final void removeInterestRead(SelectionKey key) {
+        if (LOGGER.isLoggable(Level.FINEST)) {
+            // TODO probably want some more info about the key here...
+            LOGGER.log(Level.FINEST, "Scheduling removing OP_READ to {0}", key);
+        }
         interestOps.add(new InterestOps(key, 0, SelectionKey.OP_READ));
         selector.wakeup();
     }
@@ -312,6 +320,10 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      * @param key the key.
      */
     public final void addInterestWrite(SelectionKey key) {
+        if (LOGGER.isLoggable(Level.FINEST)) {
+            // TODO probably want some more info about the key here...
+            LOGGER.log(Level.FINEST, "Scheduling adding OP_WRITE to {0}", key);
+        }
         interestOps.add(new InterestOps(key, SelectionKey.OP_WRITE, 0));
         selector.wakeup();
     }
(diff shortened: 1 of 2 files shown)


jtnord:
pending running through the stress test benchmark to check for performance regressions.

jtnord:
[Stress test results](https://gist.github.com/jtnord/cd82133ecdbf573abde09ea8673cce90)
Fix uses more RAM, but has a lower VM load.

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -292,6 +292,10 @@ public final void removeInterestConnect(SelectionKey key) {
      * @param key the key.
      */
     public final void addInterestRead(SelectionKey key) {
+        if (LOGGER.isLoggable(Level.FINEST)) {
```
stephenc: I thought you were going to pull this out to a class level constant initialized on classloading so that hotspot can reduce to no-op

stephenc:
Single measurement results

`agent-load-test --protocol JNLP4-plaintext --warmup 60 --collect 180` on OS-X, java version "1.8.0_102"

CPU load before: 1.53
CPU load after: 1.49

Avg Memory before: 195592±43263
Avg Memory after: 190434±41279

No significant difference observed... I'll need to collect a few more runs to get some measure on CPU load effects

oleg-nenashev:
I think this PR is fine. More diagnostics rarely hurts, and additional condition check should not cause a significant performance degradation. So 👍 form me. BTW, it's going into remoting 3.1+ IMHO

amuniz:
:+1:

oleg-nenashev:
Retriggering

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
batmat: I wouldn't have added the if here. It's adding verbosity for no reason in that case. I suppose you added that because of the TODO below...

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -480,7 +496,11 @@ public final void run() {
      * Process the scheduled tasks list.
      */
     private void processScheduledTasks() {
-        if (scheduledTasks.size() > 4) {
+        final int tasksWaiting = scheduledTasks.size();
+        if (LOGGER.isLoggable(Level.FINEST)) {
```
batmat: same

stephenc:
@batmat those `if` statements are critical for performance as this is hot code and they avoid constructing the `new Object[]`

batmat:
@stephenc my comment wasn't blindly targeting all ifs. I'm aware of what you say. The one I commented on weren't constructing objects.

stephenc:
Still having to set up the stack frame to pass through... the testing I did on a few of these showed a negative impact where the if was removed... there was negligible difference vs where I moved the `LOGGER.isLoggable(Level.FINEST)` to a class level final boolean though

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
jtnord: @stephenc so the performance without inlining didn;t seem to bad - and if we leave it as is then you can use the standard Jenkins UI (or JMX for agents) to change the logging level

jtnord:
@reviewbybees done - @oleg-nenashev over to you.

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
jtnord: test saves auto-boxing and GC churn

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
jtnord: >  I suppose you added that because of the TODO below...

yes - easier to do it now and know the performance hit - also prevents someone forgetting to do it later.

PR #118, opened 2016-10-07
[JENKINS-37566] - Always run setAccessible and classloader constructors in the privileged blocks

oleg-nenashev:
- [x] - Always run setAccessible() from an privileged block (via Util class)
- [x] - Create Classloaders inside privileged blocks

Just fixes about 7 issues reported by FindBugs.


src/main/java/hudson/remoting/Which.java
```diff
@@ -141,19 +144,25 @@ public static File jarFile(Class clazz) throws IOException {
                 Object delegate = is;
                 while (delegate.getClass().getEnclosingClass()!=ZipFile.class) {
                     Field f = delegate.getClass().getDeclaredField("delegate");
-                    f.setAccessible(true);
+                    if (!ReflectionUtils.makeAccessibleOrLog(f, LOGGER, Level.FINE)) {
+                        continue;
```
oleg-nenashev: Self-:bug: causes infinite cycle

src/main/java/hudson/remoting/Which.java
```diff
@@ -141,19 +144,25 @@ public static File jarFile(Class clazz) throws IOException {
...
+                    }
+
                     delegate = f.get(delegate);
                     //JENKINS-5922 - workaround for CertificateReaderInputStream; JBoss 5.0.0, EAP 5.0 and EAP 5.1
                     if(delegate.getClass().getName().equals("java.util.jar.JarVerifier$VerifierStream")){
                         f = delegate.getClass().getDeclaredField("is");
-                        f.setAccessible(true);
+                        if (!ReflectionUtils.makeAccessibleOrLog(f, LOGGER, Level.FINE)) {
+                            continue;
```
oleg-nenashev: Maybe better to stop execution as well

src/main/java/org/jenkinsci/remoting/nio/SelectableFileChannelFactory.java
```diff
@@ -128,7 +131,7 @@ public SocketChannel create(FileDescriptor fd) {
         } catch (ClassNotFoundException e) {
```
oleg-nenashev: Multi-catch around should be cleaned up during the refactoring

src/test/java/org/jenkinsci/remoting/engine/HandlerLoopbackLoadStress.java
```diff
@@ -45,14 +45,7 @@
...
 import java.nio.channels.SocketChannel;
-import java.security.KeyManagementException;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
-import java.security.KeyStore;
-import java.security.KeyStoreException;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.security.UnrecoverableKeyException;
+import java.security.*;
```
oleg-nenashev: Self-:bug:
Tried editing something in IDEA, going back to my caves

oleg-nenashev:
@reviewbybees since the test failure is unrelated

src/main/java/org/jenkinsci/remoting/util/ReflectionUtils.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Makes a method accessible using reflection.
+     * There is no automatic accessibility status recovery.
+     * @param object Object to be modified
+     * @throws PrivilegedActionException Missing permissions or other modification error.
+     *      This exception cannot happen if the object is accessible when the method is invoked.
+     */
+    public static void makeAccessible(final @Nonnull AccessibleObject object) throws PrivilegedActionException {
```
kohsuke: :bug:  This is a wrong usage of `AccessController.doPrivileged()` that creates a security hole. It allows untrusted caller to use this public method to elevate privilege and make any method/field accessible.

It's best not to fiddle with `AcessController` IMO. The only situation where it makes a difference is when remoting is in trusted code source and the client app code is not trusted. Remoting deals too much with classloader and other potentially dangerous code. It's best just not to let the untrusted caller use remoting at all.

oleg-nenashev: Trusted caller can always do it via the reflection API, no?

oleg-nenashev: Anyway, restriction of this Utility class access is reasonable

kohsuke: > Trusted caller can always do it via the reflection API, no?

I'm not sure how to interpret that.

A decade ago, I worked on JAXB, which eventually became a part of JavaSE. A part of that integration work involves making JAXB safely usable from untrusted code (think of applet/JavaFX, for example), and that experience taught me a lot about the danger and difficulty of correctly using privilege escalation.

The findbugs explanation of `DP_DO_INSIDE_DO_PRIVILEGED` does not discuss any implication of privilege escalations and down right dangerous.

IIUC, the motivation here is to make this module FindBugs clean, and in that case I recommend simply disabling this detector.

I can tell you more about this if you really care about, but it's much too long to type it in.

oleg-nenashev: That's why I submitted it as a separate PR. I went through some docs/discussions on this topic and decided it may be reasonable to perform such change. 

The motivation is to really make FindBugs clean, but I agree it requires more investigation. Putting it on hold

oleg-nenashev:
We may hit it in Java9, but I have no plan to work on it anytime soon

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -105,14 +109,29 @@
...
             // as ClassLoaderProxy. This means, the so-called remote classloader here is
             // actually our classloader that we exported to the other side.
             return ((ClassLoaderProxy)proxy).cl;
         }
-        return new RemoteClassLoader(parent, proxy);
+
+        // Create new RemoteClassLoader otherwise
+        final RemoteClassLoader created;
+        try {
+            created = AccessController.doPrivileged(
```
jglick, line 123: You should only ever need to use this in case you are running with a security manager enabled, which for Remoting we are not. So I do not see the point of this PR. If it is just to get SpotBugs to shut up, use the warning suppression annotation, or just configure a SpotBugs exclude for that warning in the POM.

jglick (changes requested):

src/main/java/hudson/remoting/RemoteClassLoader.java
jglick: And there is a Java JEP proposing to deprecate the security manager, and associated methods like this.

jtnord:
Usb tbtime to close this PR given comments from KK and Jesse?

jeffret-b:
@oleg-nenashev , I'm going to close this, based on the various comments and how outdated it is. If you still want to do something with it you can reopen it.

PR #119, opened 2016-10-18
Improve the caller/callee correlation diagnostics

kohsuke:
I was trying to diagnose a remoting call that bridges two VMs that is hanging. This can be made easier if thread dump allows us to correlate the caller/callee.

This change adds the request ID to thread name so that thread dump will show this clearly

Also fixes one incorrect use of `toString()` in diagnosis, which should be `getName()` --- see all the other uses of thread names.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Request.java
+++ b/src/main/java/hudson/remoting/Request.java
@@ -139,7 +139,7 @@ abstract class Request<RSP extends Serializable,EXC extends Throwable> extends C
                 final String name = t.getName();
                 try {
                     // wait until the response arrives
-                    t.setName(name+" / waiting for "+channel);
+                    t.setName(name+" / waiting for "+channel.getName()+" id="+id);
                     while(response==null && !channel.isInClosed())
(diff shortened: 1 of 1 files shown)


recena:
LGTM, 🐝

stephenc:
:bee:

src/main/java/hudson/remoting/Request.java
```diff
@@ -233,6 +233,8 @@ public boolean isDone() {
 
             public RSP get() throws InterruptedException, ExecutionException {
                 synchronized(Request.this) {
+                    String oldThreadName = Thread.currentThread().getName();
```
kohsuke: This is the same variable name used everywhere. I think it's `old (thread name)` not `(old thread) name`

oleg-nenashev:
Post-:bee:

oleg-nenashev:
I will add the post-factum JIRA issue since this stabilization branch is a subject for careful reviews

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-39543

PR #120, opened 2016-10-20
[JENKINS-39150] expose diagnostics across all the channels

kohsuke:
To be used by support-core, we need to be able to enumerate all active channels. We do this via WeakHashMap so that references get automatically garbage collected.

Unclosed channel will remain in memory forever, which also helps us find those leaks.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -45,6 +45,7 @@ import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.util.Collections;
+import java.util.Date;
 import java.util.Hashtable;
 import java.util.Locale;
 import java.util.Map;
@@ -204,7 +205,31 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     private final Vector<Listener> listeners = new Vector<Listener>();
     private int gcCounter;
+
+    /**
+     * Number of {@link Command} objects sent to the other side.
+     */
     private int commandsSent;
+
+    /**
+     * Number of {@link Command} objects received from the other side.
+     *
+     * When a transport is functioning correctly, {@link #commandsSent} of one side
+     * and {@link #commandsReceived} of the other side should closely match.
+     */
+    private int commandsReceived;
+
+    /**
+     * Timestamp of the last {@link Command} object sent/received, in
+     * {@link System#currentTimeMillis()} format.
+     */
+    private long lastCommandSent, lastCommandReceived;
+
+    /**
+     * Timestamp of when this channel was connected/created, in
+     * {@link System#currentTimeMillis()} format.
+     */
+    private final long createdAt = System.currentTimeMillis();
 
     /**
      * Total number of nanoseconds spent for remote class loading.
@@ -494,6 +519,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
 
         transport.setup(this, new CommandReceiver() {
             public void handle(Command cmd) {
+                commandsReceived++;
+                lastCommandReceived = System.currentTimeMillis();
                 updateLastHeard();
                 if (logger.isLoggable(Level.FINE))
                     logger.fine("Received " + cmd);
@@ -509,6 +536,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
                 Channel.this.terminate(e);
             }
         });
+        ACTIVE_CHANNELS.put(this,ref());
     }
 
     /**
@@ -581,6 +609,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
 
         transport.write(cmd, cmd instanceof CloseCommand);
         commandsSent++;
+        lastCommandSent = System.currentTimeMillis();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -204,9 +205,33 @@
...
+     */
     private int commandsSent;
 
     /**
+     * Number of {@link Command} objects received from the other side.
+     *
+     * When a transport is functioning correctly, {@link #commandsSent} of one side
+     * and {@link #commandsReceived} of the other side should closely match.
+     */
+    private int commandsReceived;
```
oleg-nenashev: Should be AtomicLong 🐜

src/main/java/hudson/remoting/Channel.java
```diff
@@ -494,6 +519,8 @@ protected Channel(ChannelBuilder settings, CommandTransport transport) throws IO
 
         transport.setup(this, new CommandReceiver() {
             public void handle(Command cmd) {
+                commandsReceived++;
```
oleg-nenashev: 🐛 since handling may potentially happen in different threads

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1149,6 +1178,20 @@ public void dumpPerformanceCounters(PrintWriter w) throws IOException {
     }
 
     /**
+     * Print the diagnostic information.
+     */
+    public void dumpDiagnostics(PrintWriter w) throws IOException {
```
oleg-nenashev: 🐛 Unsynchronized access to <code> pendingCalls</code> container

src/test/java/hudson/remoting/ChannelTest.java
```diff
@@ -163,4 +165,14 @@ public T call() throws RuntimeException {
             return t;
         }
     }
+
+    public void testDiagnostics() throws Exception {
```
oleg-nenashev: Nice2have: reference the issue

src/main/java/hudson/remoting/Channel.java
```diff
@@ -204,9 +205,33 @@
...
+     * When a transport is functioning correctly, {@link #commandsSent} of one side
+     * and {@link #commandsReceived} of the other side should closely match.
+     */
+    private int commandsReceived;
+
+    /**
+     * Timestamp of the last {@link Command} object sent/received, in
+     * {@link System#currentTimeMillis()} format.
+     */
+    private long lastCommandSent, lastCommandReceived;
```
batmat: Also, isn't `lastCommandReceived` redundant with `lastHeard`? https://github.com/jenkinsci/remoting/blob/522a022ae961d31b7b4346f90d04c3d08115e7f7/src/main/java/hudson/remoting/Channel.java#L305

src/main/java/hudson/remoting/Channel.java
kohsuke: No, commands are received in serial order, so it is not possible for this to happen in different threads that do not have memory barrier established.

src/main/java/hudson/remoting/Channel.java
kohsuke: You are right! I removed the duplication

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1499,6 +1594,18 @@ public static Channel current() {
     }
 
     /**
+     * Calls {@link #dumpDiagnostics(PrintWriter)} across all the active channels in this system.
+     * Used for diagnostics.
+     */
+    public static void dumpDiagnosticsForAll(PrintWriter w) throws IOException {
+        for (Ref ref : ACTIVE_CHANNELS.values()) {
```
stephenc: I think it would be better to use `ACTIVE_CHANNELS.values().toArray(new Ref[0])` so that we do not need the lock. You risk CME iterating, e.g. see the javadoc for `Collections.synchronizedMap` which explicitly states:

> It is imperative that the user manually synchronize on the returned map when iterating over any of its collection views

The `toArray` will give you the shortest lock time

src/main/java/hudson/remoting/Channel.java
kohsuke: Perhaps you meant reader reading incorrect value when write to normal non-volatile long variable can go in two 32bit writes ([JLS](http://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7)). I've added volatile.

oleg-nenashev: Yes. It's what I've meant

src/main/java/hudson/remoting/Channel.java
kohsuke: resolved

kohsuke:
All review comments addressed

stephenc:
:bee:

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: makes sense

oleg-nenashev:
Would be great to address the last comment from @stephenc and to add <code>@since</code> definitions to Javadoc, but generally it's ready to go. 🐝

oleg-nenashev:
@kohsuke By the way, do you need backporting to stable-2.x? Diagnosability improvements IMHO qualify if they are pretty small and  get soaked enough.

kohsuke:
Crap, I meant to target this to `stable-2.x`

kohsuke:
Taken over by #122. See you there.

PR #122, opened 2016-10-21
[JENKINS-39150] expose diagnostics across all the channels

kohsuke:
Continuing from https://github.com/jenkinsci/remoting/pull/120

No code change since then, just re-targeting stable-2.x

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -45,6 +45,7 @@ import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.util.Collections;
+import java.util.Date;
 import java.util.Hashtable;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


kohsuke:
Merging based on :bee: given in #120

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1149,6 +1173,73 @@ public void dumpPerformanceCounters(PrintWriter w) throws IOException {
...
+     * it is request or response that's in flight.
+     *
+     * <dt>Pending incoming calls
+     * <dd>
+     * The reverse of "pending outgoing calls".
+     * Number of RPC calls (e.g., method call through a {@linkplain RemoteInvocationHandler proxy})
+     * that the other side has made to us but not yet returned yet.
+     *
+     * @since 2.26.3
+     */
```
amuniz: Thanks for the Javadocs @kohsuke 👍

oleg-nenashev:
Well, the Javadoc for the *stable* branch is not ideal. We are not supposed to provide new APIs in the stabilization branch, because it may cause binary incompatibilities  for people upgrading from Jenkins LTS to later weekly releases. I'll fix it later before the release

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: Doe not work in such way

src/main/java/hudson/remoting/Channel.java
```diff
@@ -494,7 +515,8 @@ protected Channel(ChannelBuilder settings, CommandTransport transport) throws IO
 
         transport.setup(this, new CommandReceiver() {
             public void handle(Command cmd) {
-                updateLastHeard();
+                commandsReceived++;
```
rsandell: Risk of overflow?

rsandell:
Would it be worth adding JMX beans for this?

oleg-nenashev:
https://github.com/jenkinsci/remoting/pull/120#discussion_r84303313 in the original PR has been also addressed improperly. Needs fix

PR #125, opened 2016-10-24
[JENKINS-39150] - API stabilization && compliance with the compatibility policy

oleg-nenashev:
In https://github.com/jenkinsci/remoting/pull/122 @kohsuke merged the pull request without waiting for the feedback from reviewers. It effectively introduced new public API in the stabilization branch.

I also used my chance to review the code deeply and realised that <code>IOException</code> handling is far from ideal for the production code. Hence I propose to change it as well

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1615,7 +1615,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
                         throw (Error)ex;
                     }
                     logger.log(Level.WARNING, 
-                            String.format("Cannot domp diagmostics for the channel %s", ch.getName()), ex);
+                            String.format("Cannot dump diagnostics for the channel %s", ch.getName()), ex);
                 }
             }
         }
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1229,11 +1229,14 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * Number of RPC calls (e.g., method call through a {@linkplain RemoteInvocationHandler proxy})
      * that the other side has made to us but not yet returned yet.
      *
+     * @param w Output destination
+     * @throws IOException Error while creating or writing the channel information
+     * 
      * @since 2.62.3 - stable 2.x (restricted)
      * @since TODO 3.x - public version 
      */
     @Restricted(NoExternalUse.class)
-    public void dumpDiagnostics(PrintWriter w) throws IOException {
+    public void dumpDiagnostics(@Nonnull PrintWriter w) throws IOException {
         w.printf("Channel %s%n",name);
         w.printf("  Created=%s%n", new Date(createdAt));
         w.printf("  Commands sent=%d%n", commandsSent);
@@ -1595,15 +1598,25 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * Calls {@link #dumpDiagnostics(PrintWriter)} across all the active channels in this system.
      * Used for diagnostics.
      *
+     * @param w Output destination
+     * 
      * @since 2.62.3 - stable 2.x (restricted)
      * @since TODO 3.x - public version 
      */
     @Restricted(NoExternalUse.class)
-    public static void dumpDiagnosticsForAll(PrintWriter w) throws IOException {
+    public static void dumpDiagnosticsForAll(@Nonnull PrintWriter w) {
         for (Ref ref : ACTIVE_CHANNELS.values().toArray(new Ref[0])) {
             Channel ch = ref.channel();
-            if (ch!=null) {
-                ch.dumpDiagnostics(w);
+            if (ch != null) {
+                try {
+                    ch.dumpDiagnostics(w);
+                } catch (Throwable ex) {
+                    if (ex instanceof Error) {
+                        throw (Error)ex;
+                    }
+                    logger.log(Level.WARNING, 
+                            String.format("Cannot domp diagmostics for the channel %s", ch.getName()), ex);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -1226,9 +1229,14 @@ public void dumpPerformanceCounters(PrintWriter w) throws IOException {
...
-     * @since 2.26.3
+     * @param w Output destination
+     * @throws IOException Error while creating or writing the channel information
+     * 
+     * @since 2.62.3 - stable 2.x (restricted)
+     * @since TODO 3.x - public version 
      */
-    public void dumpDiagnostics(PrintWriter w) throws IOException {
+    @Restricted(NoExternalUse.class)
+    public void dumpDiagnostics(@Nonnull PrintWriter w) throws IOException {
```
oleg-nenashev: Maybe IOException is YAGNI, but I would like to keep it since one may override Channel implementation and start throwing something. The current implementation does not throw exceptions

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1226,9 +1229,14 @@ public void dumpPerformanceCounters(PrintWriter w) throws IOException {
...
      *
-     * @since 2.26.3
+     * @param w Output destination
+     * @throws IOException Error while creating or writing the channel information
+     * 
+     * @since 2.62.3 - stable 2.x (restricted)
+     * @since TODO 3.x - public version 
      */
-    public void dumpDiagnostics(PrintWriter w) throws IOException {
+    @Restricted(NoExternalUse.class)
```
oleg-nenashev: It does not effectively prevent from usage in libraries, which do not invoke `accmod` checks in the build process. Is it fine enough?

abayer:
:bee:

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1584,17 +1592,32 @@ public static Channel current() {
...
-                ch.dumpDiagnostics(w);
+            if (ch != null) {
+                try {
+                    ch.dumpDiagnostics(w);
+                } catch (Throwable ex) {
+                    if (ex instanceof Error) {
+                        throw (Error)ex;
+                    }
+                    logger.log(Level.WARNING, 
+                            String.format("Cannot domp diagmostics for the channel %s", ch.getName()), ex);
```
rsandell: domp -> dump

rsandell:
one :ant:

oleg-nenashev:
And also some additional FUN to handle in the merged PR.
@rsandell Should I address your comments in #122 here?

rsandell:
@oleg-nenashev my comments where more like questions as I don't know if an `int` is enough for long lived connections etc.
And the JMX this is more a nice to have since you then could use something like collectd et.al. for nice supervision and alarm handling. But it is probably a future addon in that case.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1584,17 +1592,32 @@ public static Channel current() {
...
-    public static void dumpDiagnosticsForAll(PrintWriter w) throws IOException {
+    @Restricted(NoExternalUse.class)
+    public static void dumpDiagnosticsForAll(@Nonnull PrintWriter w) {
         for (Ref ref : ACTIVE_CHANNELS.values().toArray(new Ref[0])) {
             Channel ch = ref.channel();
-            if (ch!=null)
-                ch.dumpDiagnostics(w);
+            if (ch != null) {
+                try {
+                    ch.dumpDiagnostics(w);
```
stephenc: If we are performance concerned, I would recommend adding 

```
if (w.checkError()) {
  return; 
}
```

as no sense continuing if PrintWriter has already suppressed the IOE

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1584,17 +1592,32 @@ public static Channel current() {
...
+                } catch (Throwable ex) {
+                    if (ex instanceof Error) {
+                        throw (Error)ex;
+                    }
+                    logger.log(Level.WARNING, 
+                            String.format("Cannot dump diagnostics for the channel %s", ch.getName()), ex);
```
stephenc: should probably send this to the print writer too

stephenc:
:bee:

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1584,17 +1594,48 @@ public static Channel current() {
...
+     */
+    @Restricted(NoExternalUse.class)
+    public static void dumpDiagnosticsForAll(@Nonnull PrintWriter w) {
+        final Ref[] channels = ACTIVE_CHANNELS.values().toArray(new Ref[0]);
+        int processedCount = 0;
+        for (Ref ref : channels) {
+            // Check if we can still write the output
+            if (w.checkError()) {
+                logger.log(Level.WARNING, 
+                        String.format("Cannot dump diagnostics for all channels, because output stream encontered an error. "
```
jglick: typo

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1584,17 +1594,48 @@ public static Channel current() {
...
+            }
+            
+            // Dump channel info if the reference still points to it
+            final Channel ch = ref.channel();
+            if (ch != null) {
+                try {
+                    ch.dumpDiagnostics(w);
+                } catch (Throwable ex) {
+                    if (ex instanceof Error) {
+                        throw (Error)ex;
```
jglick: Not necessarily helpful. (For `ThreadDeath`, yes; for `LinkageError`, generally no; for `OutOfMemoryError`, probably irrelevant.)

oleg-nenashev:
@reviewbybees done.
I'll fix the typo during the merge

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: Seems spelling in my NetBeans does not work anymore :(

PR #126, opened 2016-10-24
Maven command should use `--batch-mode`

jtnord:
The maven command should use batch mode - otherwise the logs are an unintelligible mess containing output of the amount of KB downloaded when downloading dependencies.

@oleg-nenashev @rtyler


oleg-nenashev:
👍 as a maintainer

PR #127, opened 2016-10-27
[FIXED JENKINS-39289] When a proxy fails, report what caused the channel to…

kohsuke:
… go down

Today, this requires out of bound knowledge about which connection the proxy was representing, then use other means to figure out why it has failed.

This exception chaining shortens that step and makes it easy to find the cause as to why the channel was shut down.


kohsuke:
Wrong target branch

PR #128, opened 2016-10-27
[FIXED JENKINS-39289] When a proxy fails, report what caused the channel to…

kohsuke:
… go down

Today, this requires out of bound knowledge about which connection the proxy was representing, then use other means to figure out why it has failed.

This exception chaining shortens that step and makes it easy to find the cause as to why the channel was shut down.

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -201,7 +201,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * termination and simplify the circles into chains which can then be collected easily by the garbage collector.
      * @since FIXME after merge
      */
-    private final Ref reference = new Ref(this);
+    private final Ref reference;
 
     /**
      * Registered listeners. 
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -1725,11 +1738,27 @@ public Channel channel() {
...
+            return name;
+        }
+
+        /**
          * Clears the {@link #channel} to signify that the {@link Channel} has been closed and break any complex
          * object cycles that might prevent the full garbage collection of the channel's associated object tree.
          */
-        public void clear() {
-            channel = null;
+        public void clear(Exception cause) {
```
oleg-nenashev: API breakage. Please keep the original method as a Deprecated(?) one with null cause

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1725,11 +1738,27 @@ public Channel channel() {
         }
 
         /**
+         * If the channel is null, return the cause of the channel termination.
+         */
+        public Exception cause() {
```
oleg-nenashev: Nice2Have: CheckForNull

src/main/java/hudson/remoting/Channel.java
kohsuke: This is not a method on a public class, therefore no external code could have called it.

oleg-nenashev:
@kohsuke please do not release without this fix

oleg-nenashev:
Agreed with comments. Do we keep it in Jenkins 3 only? IMHO it can be backported if you restrict the new public API

PR #130, opened 2016-11-07
[JENKINS-39547] - Corrupt jar cache

olivergondza:
[JENKINS-39547](https://issues.jenkins-ci.org/browse/JENKINS-39547)

- Verify checksum after loaded from slave jar cache on agent side.
- Extract checksum caching from `Checksum` class to `JarLoaderImpl` and `FileSystemJarCache`.
  - It was done twice on master side and made the system more fragile on both ends.
  - I wonder if we need checksum cashing on agent side at all as we should rarely call `#retrieve` on same jar more than once between agent restarts. Getting rid of it would improve this fix as it would detect jar cache manipulation without restart.

### DIFF
--- a/src/main/java/hudson/remoting/FileSystemJarCache.java
+++ b/src/main/java/hudson/remoting/FileSystemJarCache.java
@@ -60,13 +60,21 @@ public class FileSystemJarCache extends JarCacheSupport {
 
     @Override
     protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException, InterruptedException {
+        Checksum expected = new Checksum(sum1, sum2);
         File target = map(sum1, sum2);
 
         if (target.exists()) {
-            // Assume its already been fetched correctly before. ie. We are not going to validate
-            // the checksum.
-            LOGGER.fine(String.format("Jar file already exists: %16X%16X", sum1, sum2));
-            return target.toURI().toURL();
+            Checksum actual = Checksum.forFile(target);
+            if (expected.equals(actual)) {
+                LOGGER.fine(String.format("Jar file already exists: %s", expected));
+                return target.toURI().toURL();
+            } else {
+                LOGGER.warning(String.format(
+                        "Cached file checksum mismatch: %s%nExpected: %s%n Actual: %s",
+                        target.getAbsolutePath(), expected, actual
+                ));
+                target.delete();
+            }
         }
 
         try {
@@ -81,7 +89,6 @@ public class FileSystemJarCache extends JarCacheSupport {
                 }
 
                 // Verify the checksum of the download.
-                Checksum expected = new Checksum(sum1, sum2);
                 Checksum actual = Checksum.forFile(tmp);
                 if (!expected.equals(actual)) {
                     throw new IOException(String.format(
@@ -93,7 +100,6 @@ public class FileSystemJarCache extends JarCacheSupport {
                     if (!target.exists()) {
                         throw new IOException("Unable to create " + target + " from " + tmp);
                     }
-
                     // Even if we fail to rename, we are OK as long as the target actually exists at
                     // this point. This can happen if two FileSystemJarCache instances share the
                     // same cache dir.
@@ -105,8 +111,12 @@ public class FileSystemJarCache extends JarCacheSupport {
                                 "Incorrect checksum of previous jar: %s%nExpected: %s%nActual: %s",
                                 target.getAbsolutePath(), expected, actual));
                     }
+                } else {
+                    Checksum ts = Checksum.forFile(target);
+                    if (!expected.equals(ts)) {
+                        throw new Error(expected + " != " + ts);
+                    }
                 }
-
 
                 return target.toURI().toURL();
             } finally {
--- a/src/main/java/hudson/remoting/FileSystemJarCache.java
+++ b/src/main/java/hudson/remoting/FileSystemJarCache.java
@@ -1,15 +1,16 @@
 package hudson.remoting;
 
 import javax.annotation.Nonnull;
+import javax.annotation.concurrent.GuardedBy;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -28,6 +29,12 @@ public class FileSystemJarCache extends JarCacheSupport {
      * We've reported these checksums as present on this side.
      */
     private final Set<Checksum> notified = Collections.synchronizedSet(new HashSet<Checksum>());
+
+    /**
+     * Cache of computer checksums for cached jars.
+     */
+    @GuardedBy("itself")
+    private final Map<String, Checksum> checksumsByPath = new HashMap<>();
 
     /**
      * @param touch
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -117,6 +125,32 @@ protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException
...
 
+    /**
+     * Get file checksum calculating it or retrieving from cache.
+     */
+    private Checksum fileChecksum(File file) throws IOException {
+        // When callers all request the checksum of a large jar the calls to
+        // forURL will all fall through to this method since the first caller's
+        // calculation may take a while. Hence re-check the cache at the start.
+        String location = file.getCanonicalPath();
+        if (checksumsByPath.containsKey(location)) {
```
oleg-nenashev: 🐛 `checksumsByPath ` is not correctly synchronized here. `contains()` and `get()` go as separate calls, and the item may be removed between these two calls

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -117,6 +125,32 @@ protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException
...
+        Checksum cs;
+        // Synchronize the calculation to throttle the load on master side in
+        // case many slaves are connecting at the same time.
+        // Previously when a large number of slaves connected at the same time
+        // the master would experience a spike in CPU and probably I/O. By caching
+        // the results and synchronizing the calculation of the results this issue addressed.
+        synchronized (this) {
+            cs = Checksum.forFile(file);
+        }
+        checksumsByPath.put(location, cs);
```
oleg-nenashev: Again, in the case of unlikely parallel download of different JARs, we may put wrong checksum here.

* Download A 
* Thread 1. Thread interrupt/scheduling
* Thread 1.  Calculate checksum
* Thread 2. Download B, replace A
* Thread 2. Calculate B checksum
* Thread 2. Put B checksum to the cache
* Thread 2. Thread interrupt/scheduling
* Thread 1. Put A checksum to the cache (but the file is B)

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -18,10 +18,9 @@
  */
 @edu.umd.cs.findbugs.annotations.SuppressWarnings("SE_BAD_FIELD")
 class JarLoaderImpl implements JarLoader, Serializable {
-    private final ConcurrentMap<Checksum,URL> knownJars = new ConcurrentHashMap<Checksum,URL>();
+    private final ConcurrentMap<Checksum,URL> knownJars = new ConcurrentHashMap<>();
```
oleg-nenashev: Modification of the serializable class. Would it cause compatibility issues if we send it over the channel with different remoting versions on sides. I'd guess so. 🐜

src/test/java/hudson/remoting/FileSystemJarCacheTest.java
```diff
@@ -110,24 +107,36 @@ public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
     }
 
     @Test
+    public void retrieveInvalidChecksum() throws Exception {
```
oleg-nenashev: Reference issue?

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -18,10 +18,9 @@
...
 
-    @edu.umd.cs.findbugs.annotations.SuppressWarnings("DMI_COLLECTION_OF_URLS") // TODO: fix this
-    private final ConcurrentMap<URL,Checksum> checksums = new ConcurrentHashMap<URL,Checksum>();
+    private final ConcurrentMap<String, Checksum> checksums = new ConcurrentHashMap<>();
```
olivergondza: @oleg-nenashev, I did not considered this to be remotable as this is master-side abstraction provided the classloading is not by-directional. This does not seem to be generally true, though.

oleg-nenashev: JarLoader is being used as a Channel property, which is accessible on both sides. So it may be really serialized over the channel if the `Channel#remoteChannel` implementation passes it back as a seruializable result

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -60,13 +69,24 @@ protected URL lookInCache(Channel channel, long sum1, long sum2) throws IOExcept
...
-            // the checksum.
-            LOGGER.fine(String.format("Jar file already exists: %16X%16X", sum1, sum2));
-            return target.toURI().toURL();
+            Checksum actual = fileChecksum(target);
+            if (expected.equals(actual)) {
+                LOGGER.fine(String.format("Jar file already exists: %s", expected));
+                return target.toURI().toURL();
+            }
+
+            LOGGER.warning(String.format(
```
oleg-nenashev: Is it really a warning? I suppose it's a valid behavior when somebody updates jars on the remote side (e.g. plugin update)

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -60,13 +69,24 @@ protected URL lookInCache(Channel channel, long sum1, long sum2) throws IOExcept
...
     protected URL retrieve(Channel channel, long sum1, long sum2) throws IOException, InterruptedException {
+        Checksum expected = new Checksum(sum1, sum2);
         File target = map(sum1, sum2);
 
         if (target.exists()) {
-            // Assume its already been fetched correctly before. ie. We are not going to validate
-            // the checksum.
-            LOGGER.fine(String.format("Jar file already exists: %16X%16X", sum1, sum2));
-            return target.toURI().toURL();
+            Checksum actual = fileChecksum(target);
```
oleg-nenashev: Improvement follow-up proposal: Perform checksum calculation for JAR cache during the remoting startup (with option to disable it) before establishing the connection. Should help with performance issues during first job execution due to classloading.

src/main/java/hudson/remoting/JarLoaderImpl.java
oleg-nenashev: Just ignore. :facepalm:

oleg-nenashev:
@olivergondza Maybe it makes sense to integrate the patch into stable-2.x once it's ready

src/main/java/hudson/remoting/FileSystemJarCache.java
olivergondza: It is, when you update a jar, it should have different checksum and therefore be stored on different location so this code path will not be executed. This warning is a signal that what we found cached as `ABC` (by location) is in fact `DEF` (by checksum). This can be caused by manipulating the cache dir (what this PR addresses), bug in remoting that cached jar on wrong location or change of checksum algorithm. In all these cases it is undesirable to load the jar blindly.

olivergondza:
I do not consider it all that critical after we agreed it is not security vulnerability. No hurry as long as I am concerned.

src/main/java/hudson/remoting/FileSystemJarCache.java
oleg-nenashev: Agreed

oleg-nenashev:
Retriggering the CI

olivergondza:
Maven invocation on ci.jenkins.io fails badly as maven terminate prematurely there. I am not able to reproduce that locally with:

```
Apache Maven 3.3.9 (NON-CANONICAL_2015-11-23T13:17:27+03:00_root; 2015-11-23T11:17:27+01:00)
Maven home: /opt/maven
Java version: 1.8.0_112, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-jdk/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.31-1-lts", arch: "amd64", family: "unix"
```

PR #133, opened 2016-11-17
[JENKINS-39835] - Be extra defensive about Errors and Exceptions

jtnord:
[JENKINS-39835](https://issues.jenkins-ci.org/browse/JENKINS-39835) Be even more defensive then against leaving connections dangling.

NB: no testing performed of this fix - it is a direct port from the internal code that was submitted to make the JNLP v4 protocol.
@reviewbybees esp @stephenc @oleg-nenashev

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
@@ -177,15 +177,19 @@ public class NIONetworkLayer extends NetworkLayer implements IOHubReadyListener 
                         }
                         recvKey.cancel();
                         onRecvClosed();
-                    } catch (RuntimeException e) {
+                    } catch (Throwable t) {
                         // this should *never* happen... but just in case it does we will log & close connection
-                        if (LOGGER.isLoggable(Level.WARNING)) {
-                            LogRecord record = new LogRecord(Level.WARNING, "[{0}] Uncaught {1}");
-                            record.setThrown(e);
-                            record.setParameters(new Object[]{stack().name(), e.getClass().getSimpleName()});
+                        try {
+                            if (LOGGER.isLoggable(Level.SEVERE)) {
+                                LogRecord record = new LogRecord(Level.SEVERE, "[{0}] Uncaught {1}");
+                                record.setThrown(t);
+                                record.setParameters(new Object[]{stack().name(), t.getClass().getSimpleName()});
+                            }
+                        } finally {
+                            // incase this was an OOMErr and logging caused another OOMErr
+                            recvKey.cancel();
+                            onRecvClosed();
                         }
-                        recvKey.cancel();
-                        onRecvClosed();
                     } finally {
                         release(recv);
                     }


stephenc:
:bee:

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
```diff
@@ -177,15 +177,19 @@ public void ready(boolean accept, boolean connect, boolean read, boolean write)
...
+                        try {
+                            if (LOGGER.isLoggable(Level.SEVERE)) {
+                                LogRecord record = new LogRecord(Level.SEVERE, "[{0}] Uncaught {1}");
+                                record.setThrown(t);
+                                record.setParameters(new Object[]{stack().name(), t.getClass().getSimpleName()});
+                            }
+                        } finally {
+                            // incase this was an OOMErr and logging caused another OOMErr
+                            recvKey.cancel();
+                            onRecvClosed();
```
oleg-nenashev: I suppose this code should rethrow errors afterwards

amuniz: It was not doing it before the change.

amuniz:
:bee:

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
oleg-nenashev: It was not catching `Error`s before the change

oleg-nenashev:
@jtnord Do you think we need a similar fix in stable-2.x?

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
jtnord: if it does that you could kill the thread / leave the system in a bad state.

jtnord:
@oleg-nenashev if you throw the error will likely kill the thread and you may not end up in the state you think you want to end up in.

I'm not comfortable re-throwing Errors, so if you would like that then I will leave it up to yourself, cloudbees oss-team or someone else to provide.

jtnord:
> Do you think we need a similar fix in stable-2.x?

This does not contain the JNLP4v protocol (and hence this code) does it?

stephenc:
W.r.t. Killing the thread. Thinking about this more... You would only ever kill a worker thread, which should be ok as they are just assigned to this operation and the threadpool will replace if they die

oleg-nenashev:
@jtnord @stephenc Should we proceed with this PR?

jtnord:
Yes.. ops read is no longer registered and only adds it later on, leaving hanging network.  Seen at a customer site.

oleg-nenashev:
I still do not like the handling of some fatal errors, but I agree to merge it taking the potential impact into account.

oleg-nenashev:
BTW I also think that killing the worker thread is fine

jtnord:
@oleg-nenashev if the worker thread dies handling a read then whatever is connected is dead as far as OP_READ.  as OP_READ is removed when we start the thread and only removed when we have read all of the data available in the callback (ie the thread that dies).

See line 147 in [this block](https://github.com/jenkinsci/remoting/blob/ec9b5c13b879f44c04fa28ee6c8b113a165c9e57/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java#L138-L165)

Did you see my proposal on the Jenkins-dev list for handling uncaught exceptions :-)

oleg-nenashev:
@jtnord 
> Did you see my proposal on the Jenkins-dev list for handling uncaught exceptions 

Show me the link

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
```diff
@@ -177,15 +177,19 @@ public void ready(boolean accept, boolean connect, boolean read, boolean write)
...
-                        if (LOGGER.isLoggable(Level.WARNING)) {
-                            LogRecord record = new LogRecord(Level.WARNING, "[{0}] Uncaught {1}");
-                            record.setThrown(e);
-                            record.setParameters(new Object[]{stack().name(), e.getClass().getSimpleName()});
+                        try {
+                            if (LOGGER.isLoggable(Level.SEVERE)) {
+                                LogRecord record = new LogRecord(Level.SEVERE, "[{0}] Uncaught {1}");
+                                record.setThrown(t);
+                                record.setParameters(new Object[]{stack().name(), t.getClass().getSimpleName()});
+                            }
```
jglick: The whole block could be written more simply as

```java
LOGGER.log(Level.SEVERE, "Uncaught error in " + stack().name(), t);
```

without any real loss.

src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java
```diff
@@ -177,15 +177,19 @@ public void ready(boolean accept, boolean connect, boolean read, boolean write)
...
+                        } finally {
+                            // incase this was an OOMErr and logging caused another OOMErr
```
jglick: Easier and probably safer to set a `boolean` flag at the top of the method indicating whether it completed normally; wrap everything in a `try`-`finally` block that closes the channel if not.

oleg-nenashev:
OK, let's go forward with this PR as is.
@jtnord , my plan is to have remoting-3.2.1 with this fix only if @olivergondza agrees. Actually #138 may also deserve backporting due to the quite wide exposure of the issue. @olivergondza WDYT?

olivergondza:
I am ok with the code as it is - logging the errors is fine. It seems reasonable safe to have it backported into .2.

oleg-nenashev:
OK, going forward with the merge. Will be in this weekly if @kohsuke is available

PR #134, opened 2016-11-18
Docs - Rephrasing JNLP4-plaintext a bit to clarify

batmat:
Seems like "like for like" might exist, but IMO it makes the sentence a bit more complicated that required.
And as a non-native person, it felt surprising, so maybe without it it'll be a bit easier to read.

@reviewbybees


oleg-nenashev:
Just a Flaky Test

oleg-nenashev:
🐝 and merging as a maintainer

stephenc:
So what you have lost is that JNLP4-plaintext is the closest comparison with JNLP2-connect. It's an "apples to apples" comparison if you want another phrasing of it.

PR #135, opened 2016-11-25
Investigate failed build on ci.jenkins.io

olivergondza:


olivergondza:
It still fail the same way

PR #138, opened 2016-12-15
[JENKINS-40491] - Improve diagnostincs of the preliminary FifoBuffer termination

oleg-nenashev:
When NioChannelHub suffers from the preliminary buffer closure, it will print a SEVERE log to the Agent log. This change should improve diagnostics of issues like [JENKINS-31050](https://issues.jenkins-ci.org/browse/JENKINS-31050), which has been likely somehow addressed/fixed in #100.

https://issues.jenkins-ci.org/browse/JENKINS-40491

@reviewbybees


oleg-nenashev:
@stephenc 

> Is there an analogous change for the new JNLP4 protocols?

Do you mean equivalent of #100 or of this change?

stephenc:
Well #100 does not make sense for the new protocols as they all use ByteBufferQueue to ensure sequencial contract.

Thinking some more, the new protocols should be immune from this class of issue... let's go

PR #140, opened 2017-01-10
[FIX JENKINS-40710, JENKINS-41513] Add null check, use correct header for map key

cgag:
I just tested this fix (I filed the bug) and ran into an NPE (even when there are headers, it appears the keyset contains null), and then realized it was still returning null even after matching because it was indexing into the map with the incorrectly cased header.

* https://issues.jenkins-ci.org/browse/JENKINS-41513
* https://issues.jenkins-ci.org/browse/JENKINS-40710


oleg-nenashev:
CC @daniel-beck

oleg-nenashev:
@cgag Integrtaed the fix, thanks!

olivergondza:
FTR, it seems that `HttpURLConnection` stored the first line (like `HTTP/1.1 301 Moved Permanently`) as the null key.

PR #136, opened 2016-11-29
ProxyException should also retain any suppressed exceptions from the original

jglick:
Potentially relevant for Pipeline, especially after https://github.com/jenkinsci/workflow-api-plugin/pull/18.

@reviewbybees

PR #141, opened 2017-01-13
Replacing MimicException with the older ProxyException

jglick:
`ProxyException` dates back to 2007 in 2515c1cddf732d5af6120060a2055b6795addf92. Yet for some reason @kohsuke added the nearly identical `MimicException` in 2012 in d084f5095f11a8bcb80913266d664b9d1f17dc3b. There is no reason to have both, especially when #136 improved the former.

The capability is unnecessary since `ProxyException` predates the introduction of capabilities themselves, in b310dd5443799a755b19631b88cef5c5a985dc2d in 2009.

Possibly `MimicException` could be simply deleted, not just deprecated, though it is not obvious to me whether it might be serialized somewhere on disk.

@reviewbybees esp. @oleg-nenashev

### DIFF
--- a/src/main/java/hudson/remoting/Capability.java
+++ b/src/main/java/hudson/remoting/Capability.java
@@ -55,6 +55,8 @@ public final class Capability implements Serializable {
         return (mask& MASK_PIPE_THROTTLING)!=0;
     }
 
+    /** @deprecated no longer used */
+    @Deprecated
     public boolean hasMimicException() {
         return (mask&MASK_MIMIC_EXCEPTION)!=0;
     }
@@ -172,6 +174,7 @@ public final class Capability implements Serializable {
     /**
      * Supports {@link MimicException}.
      */
+    @Deprecated
     private static final long MASK_MIMIC_EXCEPTION = 1L << 3;
 
     /**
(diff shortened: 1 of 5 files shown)


jglick:
Hmm, seems like there are still test failures that I need to look at.

jglick:
Nope, tests just seem really flaky.

oleg-nenashev:
Retriggering the PR builder after #142

src/main/java/hudson/remoting/Capability.java
```diff
@@ -55,6 +55,8 @@ public boolean supportsPipeThrottling() {
         return (mask& MASK_PIPE_THROTTLING)!=0;
     }
 
+    /** @deprecated no longer used */
+    @Deprecated
```
oleg-nenashev: One may still use the Mimic exception if he writes code in the hudson.remoting package. Not sure if it's sane enough BTW. I agree with such comment in Javadoc anyway

jglick:
> the MimicException masking logic, which is not equivalent to ProxyException behavior

How so?

oleg-nenashev:
Yes, I'd say that there is no real impact 🐝

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
@jglick IMHO it is not a backporting candidate. Do you agree?

jglick:
Agreed.

PR #142, opened 2017-01-18
Try to handle flaky tests

jglick:
Hoping this will deal with errors like in #141.

Was unable to run a build on Windows, since it fails inside VirtualBox when trying to get the owner of a file during `assembly`. Probably a VBox shared filesystem bug.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@ THE SOFTWARE.
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
     <version>1.37</version>
+    <relativePath/>
   </parent>
 
   <groupId>org.jenkins-ci.main</groupId>
@@ -82,12 +83,6 @@ THE SOFTWARE.
   </repositories>
 
   <dependencies>
-    <dependency>
-      <groupId>org.kohsuke</groupId>
-      <artifactId>access-modifier-annotation</artifactId>
-      <version>1.7</version>
-      <scope>provided</scope>
-    </dependency>
     <dependency>
       <groupId>org.codehaus.mojo</groupId>
       <artifactId>animal-sniffer-annotations</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -29,6 +29,7 @@ THE SOFTWARE.
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
     <version>1.37</version>
+    <relativePath/>
```
jglick: general POM cleanup

pom.xml
```diff
@@ -83,12 +84,6 @@ THE SOFTWARE.
 
   <dependencies>
     <dependency>
-      <groupId>org.kohsuke</groupId>
-      <artifactId>access-modifier-annotation</artifactId>
-      <version>1.7</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
```
jglick: duplicate dependency, overridden below

pom.xml
```diff
@@ -469,6 +464,14 @@ THE SOFTWARE.
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.19.1</version>
+        <configuration>
+          <trimStackTrace>false</trimStackTrace> <!-- SUREFIRE-1226 workaround -->
+          <rerunFailingTestsCount>4</rerunFailingTestsCount>
+        </configuration>
```
jglick: Also tried

```xml
<forkedProcessTimeoutInSeconds>180</forkedProcessTimeoutInSeconds>
<reuseForks>false</reuseForks>
```

but got errors so I have up on that.

jglick:
Not sure I follow what the errors on ci.jenkins.io are about, but at least tests pass on the old builder…

oleg-nenashev:
:bee:

oleg-nenashev:
@reviewbybees done

PR #129, opened 2016-10-31
[JENKINS-39370,JENKINS-39369] - Support of work directories in Remoting

oleg-nenashev:
This pull requests adds a basic support of workspaces to Remoting. It offers options, which allow enabling workspaces for JNLP and SSH agents in Jenkins. 
* In the original implementation we had no way to determine working directory on startup. Remoting itself didn't know about workspaces even if it was a Jenkins agent/slave
* With previous parameters the remoting behavior does not change
* The change will require additional configuration on the Jenkins side to enable workspaces by default

Scope of changes:
- [x] Introduce concept of working directory in the remoting (should be compatible with Jenkins agents) - [JENKINS-39370](https://issues.jenkins-ci.org/browse/JENKINS-39370)
- [x] Automatically create STDERR and STDOUT outputs if the workspace is enabled - [JENKINS-39369](https://issues.jenkins-ci.org/browse/JENKINS-39369)
- [x] Use alternate JAR cache location if workspace is specified [JENKINS-18578](https://issues.jenkins-ci.org/browse/JENKINS-18578)
- [x] Add option to fail the agent startup if the workdir does not exist [JENKINS-39130](https://issues.jenkins-ci.org/browse/JENKINS-39130)
- [x] Functional tests
- [x] Automatic Logrotate for logs within workdir. Should be manageable by JUL settings 
- [x] Documentation
- [x] Adjust logging to make it compliant with docs, add support of JUL property files

Potential improvements:
- Postponed: specifying workdir after the agent startup (let Jenkins enable it?)
- TBD: other magic to automagically determine Workdir in a compatible way

PR #143, opened 2017-01-27
Print the FS Jar Cache directory in exception when it is not writable

carlossg:


oleg-nenashev:
🐝 . CC @reviewbybees. It will be pretty important once I do the rest of the changes in #129

oleg-nenashev:
It was not originally submitted with reviewbybees + the change is trivial. Hence I feel I can just merge it

PR #145, opened 2017-01-31
stable-2.x should be two digit versions

kohsuke:
In order to avoid confusion with three digit releases and their
respective branch names, let's keep stable-2.x for two digits.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@ THE SOFTWARE.
 
   <groupId>org.jenkins-ci.main</groupId>
   <artifactId>remoting</artifactId>
-  <version>2.62.5-SNAPSHOT</version>
+  <version>2.63-SNAPSHOT</version>
 
   <name>Jenkins remoting layer</name>
   <description>


kohsuke:
`2.62.5-SNAPSHOT` is for the `2.26.x` branch

jglick:
> for the 2.26.x branch

2.62.x I hope you mean.

kohsuke:
I'd just point out the fact that this confusion has already happened; `2.62.x` branch was tracking 2.62.x release and the effort went in there didn't recognize that there's also `stable-2.x` branch that was tracking 2.62.x. The entire CERT team missed that. I think that's a clear enough sign of confusion.

The security fix has to deal with so many branches. having `2.62.x` branch own 2.62.x makes it more consistent with all the other 3 digit branches. It'll also help downstream integrators understand that 3 digit remoting release is a security release.

If you insist on never producing 2.63 and instead stick with 2.62.x forever, I don't understand it but I respect that. In that case, please remove `stable-2.x` branch, so that there won't be two active branches that are tracking the same 2.62.x releases.

oleg-nenashev:
@kohsuke I would rather disagree that the fact of the `stable-2.x` branch was unknown to the SECURITY team:
* `stable-2.x` naming is what we've discussed with you during the discussion of the Remoting 3 release. This naming has been chosen, because it follows the LTS branc naming approach in the core
* There were also plenty of pull requests featuring this branch (including the documentation release). 
* Previously you executed a couple of requests regarding the releases from `stable-2.x`
* There were pull requests from SECURITY team members to this branch, namely @kohsuke , @stephenc and @abayer 

So I do not buy the argument that `stable-2.62.x` is something unknown. Anyway, this branch has been created earlier that `2.62.x`

I also do not know who has created `2.62.x`. From what I see, @kohsuke, it was you or @jglick during the release of the `2.62.3` security fix. According to Git, I have merged the branch after this release to `stable-2.x`. Not sure if I have notified you about the collision in November.

On the other hand, I admit that having 3-digit names is something, which may be more preferable to the SECURITY team. My proposals would be to...

1) Add more explicit documentation of the current branching strategy for baselines
2) Discuss the preferred approach for next releases with the SECURITY team

stephenc:
enjoying the show

oleg-nenashev:
@stephenc If you want to comment, I would rather prefer to see your opinion instead of the meme GIF.

@kohsuke We have a meeting on Friday anyway, let's discuss it there and agree on further naming

stephenc:
@oleg-nenashev I have no opinion, but it is a very enjoyable scrap to watch... the 🍿 is tasty too!

jglick:
Probably I created the branch, in the course of merging up aed8272898db2cd7779c8e52e62777c3f66ed587 & 94bdbbde848065aa623cd46853e1255ab0e490dc into other release lines (since these started in `2.53.x`). Since all other x.y.z versions referred to security releases, or more generally to targeted fixes to particular x.y versions, I maintained the same pattern, creating branches where necessary (definitely I created `3.4.x` to ensure that only 3.4 plus security fixes were shipped, and not other stuff in 3.5-SNAPSHOT). I had no knowledge of the existence of a `stable-2.x` branch or of this special-cased versioning strategy for the 2.62.x releases. If someone on the CERT team knew otherwise, they were not watching commits; the problem only arose when actually trying to cut a release.

I would request a consistent naming convention. Probably there should be no `stable-2.x` branch name as it is inconsistent with the others. `2.x` would be appropriate for commits targeting 2.63, if that ever exists. If you plan to create more 2.62.x releases with non-API, non-security changes, use `2.62.x`.

jglick:
@oleg-nenashev & I agreed to just drop this branch and use `2.62.x` etc. instead.

jglick:
Cf. #160.

PR #146, opened 2017-02-05
[JENKINS-41730] - Add option to ignore the X-Jenkins-Agent-Protocols header

oleg-nenashev:
In particular cases the `X-Jenkins-Agent-Protocols` response header may change (e.g. due to the improperly configured proxy). This change adds diagnostics for such case and also adds a system property, which allows working around the issue by sacrificing the performance.

- [x] Add the `org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver.ignoreJenkinsAgentProtocolsHeader` property, which disables the cache
- [x] Print warning in the case if the list of supported protocols is empty
- [x] Add documentation

https://issues.jenkins-ci.org/browse/JENKINS-41730

@reviewbybees, esp. @stephenc

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -191,9 +191,9 @@ public class JnlpAgentEndpointResolver {
                                 "In the case of the header corruption as a workaround you can use the " +
                                 "'org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver.protocolNamesToTry' system property " +
                                 "to define the supported protocols.");
+                    } else {
+                        LOGGER.log(Level.INFO, "Remoting server accepts the following protocols: {0}", agentProtocolNames);
                     }
-                    
-                    LOGGER.log(Level.INFO, "Remoting server accepts the following protocols: {0}", agentProtocolNames);
                 }
                 
                 if (PROTOCOL_NAMES_TO_TRY != null) {
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -77,15 +77,14 @@ public class JnlpAgentEndpointResolver {
     private String tunnel;
 
     /**
-     * If enabled, remoting will ignore the list of supported protocols returned by the server.
-     * This option can be used in order to workaround issues when the headers cannot be delivered
+     * If specified, only the protocols from the list will be tried during the connection.
+     * The option provides protocol names, but the order of the check is defined internally and cannot be changed.
+     * This option can be also used in order to workaround issues when the headers cannot be delivered
      * from the server due to whatever reason (e.g. JENKINS-41730).
-     * Ignoring of the supported protocol headers leads to higher connection times because remoting agent 
-     * has to try connecting protocols till it finds the supported one.
      * @since TODO
      */
-    private static final boolean IGNORE_SUPPORTED_PROTOCOL_NAMES_HEADER =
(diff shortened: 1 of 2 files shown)


stephenc:
:bug: if the server says do not try JNLP2 then do not try JNLP2 as there could be a MiM. Better to have the system property take the list of protocols rather than being a boolean

That way for systems needing the assurance of JNLP4 they can keep that assurance

oleg-nenashev:
@stephenc OK, will update. Though I feel it causes some fun with order of the protocols. It should be listed at least

stephenc:
>  causes some fun with order of the protocols

Just say it is the *set* of protocols to try. the order of protocol attempts is hard-coded

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -173,7 +184,31 @@ public JnlpAgentEndpoint resolve() throws IOException {
                             }
                         }
                     }
+                    
+                    if (agentProtocolNames.isEmpty()) {
+                        LOGGER.log(Level.WARNING, "Received the empty list of supported protocols from the server. " +
+                                "All protocols are disabled on the master side OR the 'X-Jenkins-Agent-Protocols' header is corrupted (JENKINS-41730). " +
```
daniel-beck: So the problem is that the header is set, but empty? Who the hell does that!?

Shouldn't this also apply to the case that the header isn't set, or do we handle that better anyway?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -173,7 +184,31 @@ public JnlpAgentEndpoint resolve() throws IOException {
...
+                                "In the case of the header corruption as a workaround you can use the " +
+                                "'org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver.protocolNamesToTry' system property " +
+                                "to define the supported protocols.");
+                    }
+                    
+                    LOGGER.log(Level.INFO, "Remoting server accepts the following protocols: {0}", agentProtocolNames);
```
daniel-beck: Should this not be in an else block?

oleg-nenashev:
@daniel-beck. "No header" is being already handled, because it is a valid
case when new agent connects to the master with old core version.

Regarding the header replacement, this is a HTTPS wrapper being used by one
of CB customers. Cannot say more here, but you can find the support ticket.
I have confirmed that my Nginx-powered HTTPS wrapper returns headers
correctly. So it is likely a configuration issue. But workaround is
reasonable in ang case IMHO (e.g. when a custom agent admin wants to
enforce JNLP4)

daniel-beck:
> Regarding the header replacement, this is a HTTPS wrapper

Right, just wondered where we draw the line between what's reasonable to support, and what's just too crazy…

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: yes, I agree

oleg-nenashev:
@stephenc ping, wdyt?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
stephenc: Were you going to address this?

oleg-nenashev:
Yes, but there was no sense to fix it before the major blocker got solved

oleg-nenashev:
May be a candidate for backporting

oleg-nenashev:
I doubt we really want to backport it.

oleg-nenashev:
@reviewbybees done

PR #148, opened 2017-02-10
[FIXED JENKINS-41852] Tweak pinning logic

stephenc:
See [JENKINS-41852](https://issues.jenkins-ci.org/browse/JENKINS-41852)

@reviewbybees


oleg-nenashev:
🐝 though a direct Unit test would be useful. I also think that reference count tweaking for `pin()` is a hack, which may deserve removal (though I cannot imagine a sane case when the object is being released MAX_INT/2 times)

oleg-nenashev:
@reviewbybees done

PR #150, opened 2017-02-12
Document and annotate object Export/Unexport logic

oleg-nenashev:
During last releases we have fixed several issues related to the object export/unexport operations (JENKINS-23271, JENKINS-41852). I decided that it makes sense to improve the method documentation just to make Javadocs more explicit.

The change also adds missing documentation of `hudson.remoting.ExportTable.unexportLogSize`

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -73,14 +73,17 @@ final class ExportTable {
          * {@code object.getClass().getName()} kept around so that we can see the type even after it
          * gets deallocated.
          */
+        @Nonnull
         private final String objectType;
         /**
          * Where was this object first exported?
          */
+        @Nonnull
         final CreatedAt allocationTrace;
         /**
          * Where was this object unexported?
          */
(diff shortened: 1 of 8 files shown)


src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -347,8 +373,8 @@ synchronized Object getOrNull(int oid) {
         return null;
     }
     
-    synchronized @Nonnull
-    Class[] type(int id) throws ExecutionException {
+    @Nonnull
+    synchronized Class[] type(int id) throws ExecutionException {
```
jglick: gratuitous hunk

oleg-nenashev: I can revert it if you prefer. Just wanted to align the syntax in the file

jglick: Just a general preference to keep diffs small.

oleg-nenashev:
@reviewbybees done

PR #152, opened 2017-02-27
[FIXED JENKINS-42371] - Call disconnect() on HttpURLConnection object

nehaljwani:
Release the network resource(s) held by the underlying connection(s?):

 - during retry attempts
 - after obtaining the JNLP connection arguments


oleg-nenashev:
As we discussed in IRC, it is a backporting candidate

nehaljwani:
Related issue: https://issues.jenkins-ci.org/browse/JENKINS-42371

PR #153, opened 2017-03-06
Cleanup hudson.remoting.Channel.Ref - NPE et al.

oleg-nenashev:
This is an aggregated PR, cleans up the class implementation

- [x] Fix the NPE risk introduced by @kohsuke in https://github.com/jenkinsci/remoting/commit/0d8a2af7cffa6aa5a6f2675e810b286a984af04e
- [x] Add name to toString() in order to display names for terminated channels
- [x] Update Javadoc and annotations

https://github.com/jenkinsci/remoting/commit/0d8a2af7cffa6aa5a6f2675e810b286a984af04e also introduces the incompatible change of Ref#clear(), but it's fine this the class is package-internal

@reviewbybees, esp. @stephenc as a creator of this class

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -876,7 +876,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     @java.lang.SuppressWarnings("ToArrayCallWithZeroLengthArrayArgument")
     @SuppressWarnings("ITA_INEFFICIENT_TO_ARRAY") // intentionally; race condition on listeners otherwise
-    public void terminate(IOException e) {
+    public void terminate(@Nonnull IOException e) {
         try {
             synchronized (this) {
                 if (e == null) throw new IllegalArgumentException();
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@reviewbybees done

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1734,7 +1736,7 @@ public static void dumpDiagnosticsForAll(@Nonnull PrintWriter w) {
          * @param channel the {@link Channel}.
          */
         private Ref(@CheckForNull Channel channel) {
-            this.name = channel.getName();
+            this.name = channel != null ? channel.getName() : "unknown (null reference)";
```
jglick: Where was FindBugs before?

oleg-nenashev: FindBugs prints warnings, but it does not fail the build now. I am handling it as https://issues.jenkins-ci.org/browse/JENKINS-37566 . One year ago there were more than 100 failures, once I merge the pending PRs there should be only 10 left

PR #154, opened 2017-03-06
FindBugs - Unravel the classloading clew coming from the PreloadJarTask

oleg-nenashev:
It appears that the target classloader in PreloadJarTask may be actually null. And FindBugs is grumbling about it. So I made an attempt to process the dependencies and to properly annotate methods and to catch possible edge cases.

The only real problem is that UserRequest#perform() may fail horribly if the request has been constructed with null classloader ref. So I have added explicit handling for this case on the invoking side.

@reviewbybees, esp. @kohsuke and @stephenc who are involved into this logic

### DIFF
--- a/src/main/java/hudson/remoting/DelegatingCallable.java
+++ b/src/main/java/hudson/remoting/DelegatingCallable.java
@@ -23,6 +23,8 @@
  */
 package hudson.remoting;
 
+import javax.annotation.CheckForNull;
+
 /**
  * {@link Callable} that nominates another claassloader for serialization.
  *
@@ -37,10 +39,12 @@ package hudson.remoting;
  * In such a case, implement this interface, instead of plain {@link Callable} and
  * return a classloader that can see all the classes.
  *
- * In case of Hudson, {@code PluginManager.uberClassLoader} is a good candidate.  
+ * In case of Jenkins, {@code PluginManager.uberClassLoader} is a good candidate.  
  *
  * @author Kohsuke Kawaguchi
  */
 public interface DelegatingCallable<V,T extends Throwable> extends Callable<V,T> {
+    
+    @CheckForNull
     ClassLoader getClassLoader();
 }
--- a/src/main/java/hudson/remoting/DelegatingCallable.java
+++ b/src/main/java/hudson/remoting/DelegatingCallable.java
@@ -45,6 +45,14 @@ import javax.annotation.CheckForNull;
  */
 public interface DelegatingCallable<V,T extends Throwable> extends Callable<V,T> {
     
+    /**
+     * Returns the class loader to be used for the callable.
+     * 
+     * @return {@link ClassLoader} to be used.
+     *         The value may be {@code null} if the classloader is not being propagated to the remote side.
+     *         If all classes in the call are primitives or {@code Void}, the value may be also {@code null}. 
+     *         In such cased  the handling code should try other possible classloaders.
+     */
     @CheckForNull
     ClassLoader getClassLoader();
 }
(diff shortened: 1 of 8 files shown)


src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -58,26 +62,50 @@
...
+     */
+    @CheckForNull
+    /*package*/ static ClassLoader getClassLoader(@Nonnull Callable<?,?> c) {
     	ClassLoader result = null;
         
     	if(c instanceof DelegatingCallable) {
         	result =((DelegatingCallable)c).getClassLoader();
         }
-        else {
+        if (result == null) {
```
oleg-nenashev: I suspect the original code was not correct here, since we do not give classes a chance to contribute in the case of `DelegatingCallable `. But I would appreciate feedback

src/main/java/hudson/remoting/DelegatingCallable.java
```diff
@@ -37,10 +39,12 @@
...
  * return a classloader that can see all the classes.
  *
- * In case of Hudson, {@code PluginManager.uberClassLoader} is a good candidate.  
+ * In case of Jenkins, {@code PluginManager.uberClassLoader} is a good candidate.  
  *
  * @author Kohsuke Kawaguchi
  */
 public interface DelegatingCallable<V,T extends Throwable> extends Callable<V,T> {
+    
+    @CheckForNull
```
stephenc: It would be good to document what a null value means here

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -105,7 +105,14 @@
      */
     private final Map<String,ClassReference> prefetchedClasses = Collections.synchronizedMap(new HashMap<String,ClassReference>());
 
-    public static ClassLoader create(ClassLoader parent, IClassLoader proxy) {
+    /**
+     * Creates a remotable classloader
+     * @param parent Parent classloader. Can be {@code null} if there is no delegating classloader
```
stephenc: what does `null` mean... does it mean that the parent is supposed to be the receiving channel's base classloader?

src/main/java/hudson/remoting/UserRequest.java
stephenc: need to see what you think the documentation of `((DelegatingCallable)c).getClassLoader()` is w.r.t. the meaning of a null value

oleg-nenashev:
> need to see what you think the documentation of ((DelegatingCallable)c).getClassLoader() is w.r.t. the meaning of a null value

As you may guess, there was no documentation. From the `Class.getClassLoader()` documentation, I assume it is rather a case for primitive types and `Void`. It seems to be consistent with this implementation: https://github.com/jenkinsci/remoting/blob/master/src/main/java/hudson/remoting/RemoteInvocationHandler.java#L878-L883

oleg-nenashev:
Another justification for supporting `null` classloaders is different behavior on different sides of the channels. E.g. in the core there is such comment made by @kohsuke : https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/RemotingDiagnostics.java#L118-L129

src/main/java/hudson/remoting/RemoteClassLoader.java
oleg-nenashev: This null goes to the JVM internals. Javadoc for ClassLoader#parent:

```
// The parent class loader for delegation
    // Note: VM hardcoded the offset of this field, thus all new fields
    // must be added *after* it.
```

In the implementation of the `ClassLoader` I see the execution is being passed to the BootStrap Classloader then, e.g.:

```java
  protected Class<?> loadClass(String name, boolean resolve)
        throws ClassNotFoundException
    {
        synchronized (getClassLoadingLock(name)) {
            // First, check if the class has already been loaded
            Class<?> c = findLoadedClass(name);
            if (c == null) {
                long t0 = System.nanoTime();
                try {
                    if (parent != null) {
                        c = parent.loadClass(name, false);
                    } else {
                        c = findBootstrapClassOrNull(name);
                    }
                } catch (ClassNotFoundException e) {
                    // ClassNotFoundException thrown if class not found
                    // from the non-null parent class loader
                }
...
```

oleg-nenashev:
@reviewbybees done

PR #156, opened 2017-03-07
[JENKINS-48130] - Improve handling and diagnostics of RejectedExecutionException in the code

oleg-nenashev:
Background: I was analysing JIRA issues related to the NIOHub fatal channel termination causing massive disconnection of agents. It appears that the `SingleLaneExecutor` is not completely correctly used there...

TL;DR: A single packet sent to the channel with pending shutdown may cause the termination of all remoting channels in `JNLP1`, `JNLP2`, `CLI`, and `CLI2` protocols. `JNLP4` does not seem to be affected.

* When we receive the command in a particular `NioTransport`, message parts are are being submitted to its `SingleLaneExecutor` 
* If the executor service rejects the task (e.g. due to the pending shutdown), a runtime `RejectedExecutionException` is being thrown. E.g. [here](https://github.com/jenkinsci/remoting/blob/master/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java#L595-L599)
* This exception is not being caught on the `NioTransport` level and gets proparated to the top level of `NioChannelHub#run()`
* `NioChannelHub#run()` catches the unhandled `RuntimeException`... and terminates the entire `NioHub`. [code is here](https://github.com/jenkinsci/remoting/blob/master/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java#L655-L660)

https://issues.jenkins-ci.org/browse/JENKINS-48130

### Applied changes

- [x] - Introduced the `ExecutorServiceUtils` class, which encapsulates the unsafe task submission and its runtime exception 
 ** FindBugs was raising the flag about the method handling BTW. [Here is also a discussion](https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2010-November/003223.html) about this feature in FindBugs
- [x] - Also introduced distinguishing of  `RejectedExecutionException` and FATAL `RejectedExecutionException` in the logic
- [x] - Modified `NioChannelHub` to handle the task submission errors without BOOM
- [x] - Modified `JarCacheSupport` to handle the task submission errors (e.g. when remoting gets shutdown in parallel with classloading) - same cause
- [x] - Modified `SingleLaneExecutorService` to properly handle the issues happening in its proxied Executor Service

### Related issues

* I would expect the messages like "Unexpected shutdown of the selector thread" or "The executor service is shutting down" to be reported in system logs for the root cause, but there is no such JIRA tickets
* I see no issues with such pattern in JIRA, so the log messages are not being reported on the master side. But it still may happen on the agent side. Or my analysis may be wrong
* I have seen many issues regarding spontaneous failures reported to JNLP2, which have been fixed by migrating to JNLP4. It could be one of the rootcases

### DIFF
--- a/src/main/java/hudson/remoting/JarCacheSupport.java
+++ b/src/main/java/hudson/remoting/JarCacheSupport.java
@@ -65,9 +65,16 @@ public abstract class JarCacheSupport extends JarCache {
                     // if nobody else caches it before.
                     inprogress.putIfAbsent(key, promise);
                 } catch (ExecutorServiceUtils.ExecutionRejectedException ex) {
-                    LOGGER.log(Level.SEVERE, "Downloader executor service has rejected the download command for checksum " + key, ex);
-                    // Prevent infinite loop, try again in 100ms to let the shyutdown logic kill this operation correctly
-                    Thread.sleep(1000);
+                    final String message = "Downloader executor service has rejected the download command for checksum " + key;
+                    LOGGER.log(Level.SEVERE, message, ex);
+                    // Retry the submission after 100 ms if the error is not fatal
+                    if (ex.isFatal()) {
+                        // downloader won't accept anything else, do not even try
+                        throw new IOException(message, ex);
+                    } else {
+                        //TODO: should we just fail? unrealistic case for the current AtmostOneThreadExecutor implementation anyway
+                        Thread.sleep(100);
+                    }
                 }
             }
         }
--- a/src/main/java/hudson/remoting/JarCacheSupport.java
+++ b/src/main/java/hudson/remoting/JarCacheSupport.java
@@ -7,6 +7,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.jenkinsci.remoting.util.ExecutorServiceUtils;
 
 /**
  * Default partial implementation of {@link JarCache}.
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
@reviewbybees, and esp. @stephenc and @kohsuke since they have insights regarding the NioHub behavior.

oleg-nenashev:
Various timeouts on both CI instances :(

src/main/java/hudson/remoting/JarCacheSupport.java
```diff
@@ -50,50 +51,92 @@
...
+            this.sum1 = sum1;
+            this.sum2 = sum2;
+            this.key = key;
+            this.promise = promise;
+        }
+        
+        @Override
+        public void run() {
+            try {
+                // Deduplication: There is a risk that multiple downloadables get scheduled, hence we check if
```
oleg-nenashev: The logic has been moved to a nested class, The deduplication section below is the only difference

src/main/java/hudson/remoting/JarCacheSupport.java
```diff
@@ -50,50 +51,92 @@
...
-                            bailout(e);
-                        } catch (InterruptedException e) {
-                            // we are bailing out, but we need to allow another thread to retry later.
-                            bailout(e);
+                try {
+                    final AsyncFutureImpl<URL> promise = new AsyncFutureImpl<URL>();
+                    ExecutorServiceUtils.submitAsync(downloader, new  DownloadRunnable(channel, sum1, sum2, key, promise));
+                    // Now we are sure that the task has been accepted to the queue, hence we cache the promise
+                    // if nobody else caches it before.
+                    inprogress.putIfAbsent(key, promise);
```
oleg-nenashev: With this logic I am not fully convinced that `ConcurrentMap` is still a good idea here, maybe a common lock would be preferable since it allows to avoid duplication on race conditions and then the deduplication logic.

In this case there is also a risk that `DownloadRunnable` gets executed before putting the promise into the cache. 
And yes, putIfAbsent here should return the value if something is returned.

oleg-nenashev:
So the change in NioHub and SingleLaneExecutorService somehow makes `PipeTest` to hang randomly

oleg-nenashev:
After the patch `hudson.remoting.ChannelTest.testExportCallerDeallocation` just hangs. Attached the threaddump from the laptop. 
[hangs.txt](https://github.com/jenkinsci/remoting/files/1456611/hangs.txt)

oleg-nenashev:
@reviewbybees I have finally discovered why the tests hang. 🤦‍♂️ , but now it works. needs review

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/JENKINS-48130 for this issue

oleg-nenashev:
`JnlpProtocolHandlerTest ` still hangs consistently on CI with the recent patch

oleg-nenashev:
@reviewbybees a second review would be useful

oleg-nenashev:
I am feeling lucky. 🚢 🇮🇹

PR #157, opened 2017-03-07
[JENKINS-37566] - FindBugs - Cleanup issues in ExportTable implementation

oleg-nenashev:
This pull request just suppresses two issues in `ExportTable` implementation.

- [x] Caching of stacktrace for heapdumps. No warnings anymore
- [x] Ignore Serialization warnings for `ExportList` since this class is a part of internal API && is not being actually serialized

For discussion about caching or `Source` stacktraces:

* It is a legacy behavior introduced by @kohsuke in https://github.com/jenkinsci/remoting/commit/5db46de05db55262e75755ba47a99c4b12965d9b (remoting-2.40). 
* For every `object Deallocation` and `Reference Counter Increment` Remoting calculates the stacktrace and then also clones it. Implementation in `Throwable` is provided below
* The current behavior causes calculation and memory overheads, just to make heapdumps to provide more info
* I doubt we want this behavior to be enabled by default

Throwable implemetation:

```java
    public StackTraceElement[] getStackTrace() {
        return getOurStackTrace().clone();
    }

    private synchronized StackTraceElement[] getOurStackTrace() {
        // Initialize stack trace field with information from
        // backtrace if this is the first call to this method
        if (stackTrace == UNASSIGNED_STACK ||
            (stackTrace == null && backtrace != null) /* Out of protocol state */) {
            int depth = getStackTraceDepth();
            stackTrace = new StackTraceElement[depth];
            for (int i=0; i < depth; i++)
                stackTrace[i] = getStackTraceElement(i);
        } else if (stackTrace == null) {
            return UNASSIGNED_STACK;
        }
        return stackTrace;
    }
```

[JENKINS-37566](https://issues.jenkins-ci.org/browse/JENKINS-37566)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -23,6 +23,7 @@
  */
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -41,6 +42,8 @@ import static java.util.logging.Level.*;
 import javax.annotation.CheckForNull;
 import javax.annotation.CheckReturnValue;
 import javax.annotation.meta.When;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 /**
  * Manages unique ID for exported objects, and allows look-up from IDs.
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -225,11 +230,26 @@ synchronized void addInterface(Class<? super T> clazz) {
...
+        // Approach #1: Maybe a manual implementation of getOurStackTrace() and local storage is preferable.
+        // Approach #2: Consider disabling this logic by default
+        /**
+         * Update the internal stacktrace cache.
+         * Forces the computation of the stack trace in a Java friendly data structure,
+         * so that the call stack can be seen from the heap dump after the fact.
+         * @return Cloned version of the inner cache.
+         */
+        @CheckReturnValue(when = When.NEVER)
+        protected final StackTraceElement[] updateOurStackTraceCache() {
```
jglick: Just inline again and use `@SuppressFBWarnings`.

oleg-nenashev:
@reviewbybees done

PR #161, opened 2017-05-16
[JENKINS-44290] - Prevent crash when starting Remoting agents in the default mode

oleg-nenashev:
The class was not covered by tests at all, so the issue has sneaked into the Remoting 3.8 release. `JNLPLauncher` test caught it in CI though. Added test + one-liner fix

https://issues.jenkins-ci.org/browse/JENKINS-44290

@reviewbybees


oleg-nenashev:
@reviewbybees done

PR #162, opened 2017-05-19
Fix Javadoc and enforce its checks during the build

oleg-nenashev:
subj

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -483,6 +483,22 @@ THE SOFTWARE.
           <rerunFailingTestsCount>4</rerunFailingTestsCount>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>run-javadoc</id>
+            <goals>
+              <goal>jar</goal>
+              <goal>javadoc</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <failOnError>true</failOnError>
+          <quiet>true</quiet> <!--Do not report every checked file-->
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -495,7 +495,6 @@ THE SOFTWARE.
           </execution>
         </executions>
         <configuration>
-          <failOnError>true</failOnError>
           <quiet>true</quiet> <!--Do not report every checked file-->
         </configuration>
       </plugin>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -483,6 +483,22 @@ THE SOFTWARE.
...
+          <execution>
+            <id>run-javadoc</id>
+            <goals>
+              <goal>jar</goal>
+              <goal>javadoc</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <failOnError>true</failOnError>
```
jglick: Is this not the default?

pom.xml
```diff
@@ -483,6 +483,22 @@ THE SOFTWARE.
           <rerunFailingTestsCount>4</rerunFailingTestsCount>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>run-javadoc</id>
+            <goals>
+              <goal>jar</goal>
```
jglick: Unnecessary.

pom.xml
```diff
@@ -483,6 +483,22 @@ THE SOFTWARE.
           <rerunFailingTestsCount>4</rerunFailingTestsCount>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>run-javadoc</id>
```
jglick: You could do this only in a profile if `!skipTests` to speed up developer builds, as we do in core & plugin builds.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1225,7 +1225,7 @@ public void dumpPerformanceCounters(PrintWriter w) throws IOException {
      * Number of RPC calls (e.g., method call through a {@linkplain RemoteInvocationHandler proxy})
      * that are made but not returned yet. If you have the remoting diagnostics of the other side, you
      * can compare this number with "pending incoming calls" on the other side to see how many RPC
-     * calls are executing vs in flight. "one side's incoming calls" < "the other side's outgoing calls"
+     * calls are executing vs in flight. "one side's incoming calls" &lt; "the other side's outgoing calls"
```
jglick: or `@code`

pom.xml
oleg-nenashev: I see no need in it. Javadoc check is pretty fast. And it is no safe to do it, because I may need to release with -skipTests on Windows (really bad practice, yeah)

PR #168, opened 2017-06-20
[JENKINS-37566] - Annotate methods which may explicitly return `null`

oleg-nenashev:
Just a bulk change of annotations + some Javadoc, no behavior changes. Will submit a couple of other cases independently since they change to logic.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/MimicException.java
+++ b/src/main/java/hudson/remoting/MimicException.java
@@ -1,5 +1,9 @@
 package hudson.remoting;
 
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
 /**
  * Exception that prints like the specified exception.
  *
@@ -29,7 +33,8 @@ class MimicException extends Exception {
         return (message != null) ? (s + ": " + message) : s;
     }
 
-    public static Throwable make(Channel ch, Throwable cause) {
+    @Nullable
+    public static Throwable make(@Nonnull Channel ch, @CheckForNull Throwable cause) {
         if (cause == null)  return null;
 
         // make sure the remoting layer of the other end supports this
--- a/src/main/java/hudson/remoting/MimicException.java
+++ b/src/main/java/hudson/remoting/MimicException.java
@@ -3,6 +3,8 @@ package hudson.remoting;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
 
 /**
  * Exception that prints like the specified exception.
@@ -15,6 +17,7 @@ import javax.annotation.Nullable;
  * @deprecated Use {@link ProxyException} instead.
  */
 @Deprecated
+@Restricted(DoNotUse.class)
 class MimicException extends Exception {
     private final String className;
     MimicException(Throwable cause) {
--- a/src/main/java/hudson/remoting/MimicException.java
+++ b/src/main/java/hudson/remoting/MimicException.java
@@ -37,7 +37,7 @@ class MimicException extends Exception {
     }
 
     @Nullable
-    public static Throwable make(@Nonnull Channel ch, @CheckForNull Throwable cause) {
+    public static Throwable make(@Nonnull Channel ch, @Nullable Throwable cause) {
         if (cause == null)  return null;
 
         // make sure the remoting layer of the other end supports this
--- a/src/main/java/hudson/remoting/MultiClassLoaderSerializer.java
+++ b/src/main/java/hudson/remoting/MultiClassLoaderSerializer.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 
 /**
  * {@link ObjectInputStream}/{@link ObjectOutputStream} pair that can handle object graph that spans across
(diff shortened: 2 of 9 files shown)


src/main/java/hudson/remoting/MultiClassLoaderSerializer.java
```diff
@@ -122,6 +124,7 @@ private ClassLoader readClassLoader() throws IOException, ClassNotFoundException
             channel.classFilter.check(name);
             try {
                 ClassLoader cl = readClassLoader();
+                // TODO: handle null classloader as a System one?
```
oleg-nenashev: 8-year old change in `readClassLoader()`: https://github.com/jenkinsci/remoting/commit/6254e9a2e4e7fbd34bf2f75391768155f838163a
I cannot find any related issues

oleg-nenashev: FindBugs will make me to do something later anyway

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -224,6 +226,7 @@ public static Channel unwrap(Object proxy) {
         return null;
     }
 
+    @Nullable
```
stephenc: This is the default, I assume you are adding it here to remind yourself not to go over-zealous and add `@CheckForNull` ;-)

Also I note that in this specific case, the assumption of `@Nullable` is entirely correct (if effectively redundant), we cannot know what the contract of the invoked method is w.r.t. nullability, rather the caller of the proxied interface should know the contract of the method they are requesting an invocation of.

src/main/java/hudson/remoting/MimicException.java
```diff
@@ -29,7 +33,8 @@ public String toString() {
         return (message != null) ? (s + ": " + message) : s;
     }
 
-    public static Throwable make(Channel ch, Throwable cause) {
+    @Nullable
```
stephenc: `@Nullable` can cause confusion for a lot of people, any reason you wanted to avoid `@CheckForNull`... the contract of this method is that the nullability of the result follows the nullability of the provided parameter... so callers should be aware of whether they pass in a null... not unreasonable to ask them to be aware of the potential to get a null back when providing one... smells like a :bug:

oleg-nenashev: @stephenc This is how we handle such use-cases in other places (IIRC I was also requested to do so at some point). Here are two examples in the core:

* https://github.com/jenkinsci/jenkins/blob/461472fc009dfe313009acd8ce23cff22ee66b91/core/src/main/java/jenkins/SoloFilePathFilter.java#L24
* https://github.com/jenkinsci/jenkins/blob/c3b1bff6477e05050a40c0e20816bea9cf396534/core/src/main/java/jenkins/model/Jenkins.java#L1997

Although I agree it would better to have a `CheckForNull` (or even `Nonnull`) method, we need to agree on the approach

oleg-nenashev:
@stephenc `MimicException` class is deprecated, and it is not being used anywhere else. I have added an explicit annotation. Does it address your concern?

src/main/java/hudson/remoting/MimicException.java
stephenc: The `SoloFilePathFilter` that you cite as an example is a different case, the supplied parameter is `@Nullable` and the nullability of the result of that method is entirely determined by the nullability of the supplied parameter. Hence for `SoloFilePathFilter` the returned value being `@Nullable` (i.e. the caller is responsible for knowing when it might be `null`) is entirely correct.

The `@Nullable Jenkins.getLabelAtom(@CheckForNull String name)` is incorrect. It should probably be fixed by making the supplied parameter `@Nullable` as the only case where it could return `null` is where the `name` is `null`.

If you want to keep `@Nullable` here then the `cause` parameter should be `@Nullable` and not `@CheckForNull` otherwise you are weakening the contract of the annotations.

src/main/java/hudson/remoting/MimicException.java
```diff
@@ -11,6 +17,7 @@
  * @deprecated Use {@link ProxyException} instead.
  */
 @Deprecated
+@Restricted(DoNotUse.class)
```
jglick: Why? Sounds gratuitous. Already deprecated.

src/main/java/hudson/remoting/MimicException.java
jglick: I see no problem with the current code. `cause` is `@CheckForNull` because the code reading this parameter—the method body—does indeed need to check for null. It would be `@Nullable` only if the implementation could assume it were non-null under certain conditions, such as based on the values of other parameters. The method return is `@Nullable` since the caller may or may not need to check for null, depending on whether or not they might have passed in a null `cause`. All the Javadoc says is:

> developers will have to read the documentation to determine when a null value is acceptable and whether it is necessary to check for a null value

There is no mention of the `ElementType.METHOD` annotation being restricted to cases where nullability of a method return value is determinable by nullability of method arguments.

And FWIW the `SoloFilePathFilter` and `getLabelAtom` examples above seem exactly the same as each other, and this case: the return value is null iff the parameter is null.

src/main/java/hudson/remoting/MimicException.java
oleg-nenashev: Yes. `@Restricted` will just make a whatever external dependency to fail. But yes, it does not make much difference

oleg-nenashev:
@stephenc So do you still block this change due to the `MimicException` concern?

src/main/java/hudson/remoting/MimicException.java
stephenc: `@CheckForNull` does not force you to check the return value for null, rather it says that the return value is likely to be `null`... whereas `@Nullable` says that the caller of the method is supposed to know whether they need to check for null or not.

Either mark the method `@CheckForNull` or mark the parameter `@Nullable` if you want my blessing... Ignore if  you want to go ahead with maintainer privilidges

oleg-nenashev:
@reviewbybees done

PR #169, opened 2017-06-20
Prevent NullPointerException in ResourceImageBoth if cannot initialize JAR retrieval

oleg-nenashev:
I firstly started from annotating the methods and throwing IOException, but then decided to do a proper error propagation instead. 

It smells like a real issue though there is no relevant bugtracker entries.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ResourceImageBoth.java
+++ b/src/main/java/hudson/remoting/ResourceImageBoth.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nonnull;
 
 /**
  * @author Kohsuke Kawaguchi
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/ResourceImageBoth.java
```diff
@@ -32,19 +33,27 @@ public ResourceImageBoth(URL resource, Checksum sum) throws IOException {
...
             return c.resolve(channel, sum1, sum2);
         } catch (IOException e) {
             LOGGER.log(Level.WARNING, "Failed to initiate retrieval", e);
+            throw e;
         } catch (InterruptedException e) {
             LOGGER.log(Level.WARNING, "Failed to initiate retrieval", e);
             Thread.currentThread().interrupt(); // process the interrupt later
+            throw e;
         }
-        return null;
```
oleg-nenashev: NPE will happen in `f.isDone()` (line 30)

oleg-nenashev:
@reviewbybees done

PR #172, opened 2017-06-25
[CID-156932] - Fix the imaginary resource leak in NullOutputStream

oleg-nenashev:
`DigestOutputStream` may change in the future, hence it worth to just refactor the code. I have not found annotations which would indicate that no closing it required for NullOutputStream. There are such warnings in the core as well OTOH

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Checksum.java
+++ b/src/main/java/hudson/remoting/Checksum.java
@@ -74,8 +74,10 @@ final class Checksum {
     static Checksum forURL(URL url) throws IOException {
         try {
             MessageDigest md = MessageDigest.getInstance(JarLoaderImpl.DIGEST_ALGORITHM);
-            Util.copy(url.openStream(), new DigestOutputStream(new NullOutputStream(), md));
-            return new Checksum(md.digest(), md.getDigestLength() / 8);
+            try(OutputStream ostream = new DigestOutputStream(new NullOutputStream(), md)) {
+                Util.copy(url.openStream(), ostream);
+                return new Checksum(md.digest(), md.getDigestLength() / 8);
+            }
         } catch (NoSuchAlgorithmException e) {
             throw new AssertionError(e);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/Checksum.java
```diff
@@ -74,8 +74,10 @@ static Checksum forFile(File file) throws IOException {
     static Checksum forURL(URL url) throws IOException {
         try {
             MessageDigest md = MessageDigest.getInstance(JarLoaderImpl.DIGEST_ALGORITHM);
-            Util.copy(url.openStream(), new DigestOutputStream(new NullOutputStream(), md));
-            return new Checksum(md.digest(), md.getDigestLength() / 8);
+            try(OutputStream ostream = new DigestOutputStream(new NullOutputStream(), md)) {
```
jglick: If you actually were concerned about `NullOutputStream` leaking something (impossible), you would use

```java
try (NullOutputStream nos = new NullOutputStream();
     OutputStream ostream = new DigestOutputStream(nos)) {
```

src/main/java/hudson/remoting/Checksum.java
```diff
@@ -74,8 +74,10 @@ static Checksum forFile(File file) throws IOException {
...
+                Util.copy(url.openStream(), ostream);
```
jglick: But here you are failing to provably close the one stream here which _might_ actually hold open a socket or file handle! So you need to include this `InputStream` amongst the closeables being bound.

src/main/java/hudson/remoting/Checksum.java
oleg-nenashev: Not really required, because DigestOutputStream closes the nested one

jglick: Unless `DigestOutputStream.<init>` throws an exception. That is why as a rule you should declare every `Closeable` object you are constructing as a separate variable in the block.

PR #174, opened 2017-06-29
[JENKINS-44785] - Built-in request timeout support

oleg-nenashev:
This is a PoC implementation of the timeout support in Remoting API, See one of the use-cases in [JENKINS-44785](https://issues.jenkins-ci.org/browse/JENKINS-44785)

- [x] - Implement working PoC
- [ ] - TBD make decision about switching to Java 8. The PoC needs it for the `java.time.Duration` class, but it can be replaced if we stay on Java 7
- [ ] - If Java 8 is selected, implement new `call()` default implementations in interfaces (with `TimeoutException`s) to address the [JENKINS-44785](https://issues.jenkins-ci.org/browse/JENKINS-44785) request from @jglick 
- [ ] - `channel#callAsync()` should also support timeouts on the remote side
- [ ] - TBD: If we stay on Java 7, stick to `InterruptedException`?
- [ ] - Weaponize it! Set some default timeouts for user requests and RPC calls

@reviewbybees @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <java.level>7</java.level>
+    <java.level>8</java.level>
     <build.type>private</build.type>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
@@ -418,7 +418,7 @@ THE SOFTWARE.
         <configuration>
           <signature>
             <groupId>org.codehaus.mojo.signature</groupId>
-            <artifactId>java17</artifactId>
+            <artifactId>java18</artifactId>
             <version>1.0</version>
           </signature>
           <ignores>
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -45,6 +45,7 @@ import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Hashtable;
@@ -61,6 +62,7 @@ import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
+import org.jenkinsci.remoting.util.Timeout;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
@@ -530,8 +532,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
                 commandsReceived++;
                 lastCommandReceivedAt = System.currentTimeMillis();
                 if (logger.isLoggable(Level.FINE))
-                    logger.fine("Received " + cmd);
-                try {
+                    logger.fine("Received " + cmd);   
+                try(Timeout t = Timeout.optLimit(cmd.getExecutionTimeout())) {
                     cmd.execute(Channel.this);
                 } catch (Throwable t) {
                     logger.log(Level.SEVERE, "Failed to execute command " + cmd + " (channel " + Channel.this.name + ")", t);
@@ -832,9 +834,15 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     public <V,T extends Throwable>
     V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
+        return call(callable, UserRequest.DEFAULT_PERFORM_TIMEOUT, UserRequest.DEFAULT_EXECUTION_TIMEOUT);
+    }
+    
+    public <V,T extends Throwable>
+    V call(@Nonnull Callable<V,T> callable, @CheckForNull Duration performTimeout, @CheckForNull Duration executionTimeout) 
+            throws IOException, T, InterruptedException {
         UserRequest<V,T> request=null;
-        try {
-            request = new UserRequest<V, T>(this, callable);
+        try(Timeout t = Timeout.optLimit(executionTimeout)) {
+            request = new UserRequest<V, T>(this, callable, performTimeout, executionTimeout);
             UserResponse<V,T> r = request.call(this);
             return r.retrieve(this, UserRequest.getClassLoader(callable));
 
--- a/src/main/java/org/jenkinsci/remoting/util/Timeout.java
+++ b/src/main/java/org/jenkinsci/remoting/util/Timeout.java
@@ -1,0 +1,108 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2017 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.remoting.util;
+
+import hudson.remoting.VirtualChannel;
+import java.time.Duration;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * Allows operations to be limited in execution time.
+ * For example, {@link VirtualChannel#call} could otherwise may hang forever.
+ * Use in a {@code try}-with-resources block.
+ * @author Jesse Glick
+ * @author Oleg Nenashev
+ * @since TODO
+ */
+@Restricted(NoExternalUse.class)
+public class Timeout implements AutoCloseable {
+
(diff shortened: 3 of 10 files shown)


src/main/java/org/jenkinsci/remoting/util/Timeout.java
```diff
@@ -0,0 +1,108 @@
+/*
```
oleg-nenashev: This class is an adapted version of https://github.com/jenkinsci/workflow-support-plugin/commit/c810b3874134f60be670d1205b6673fde5003c14

jglick:
> `channel#callAsync()` should also support timeouts on the remote side

Why? I see no need for it.

pom.xml
```diff
@@ -418,7 +418,7 @@ THE SOFTWARE.
         <configuration>
           <signature>
             <groupId>org.codehaus.mojo.signature</groupId>
-            <artifactId>java17</artifactId>
+            <artifactId>java18</artifactId>
```
jglick:
```xml
<artifactId>java1${java.level}</artifactId>
```

src/main/java/hudson/remoting/Channel.java
```diff
@@ -832,9 +834,15 @@ public void setJarCache(@Nonnull JarCache jarCache) {
      */
     public <V,T extends Throwable>
     V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
+        return call(callable, UserRequest.DEFAULT_PERFORM_TIMEOUT, UserRequest.DEFAULT_EXECUTION_TIMEOUT);
+    }
+    
+    public <V,T extends Throwable>
+    V call(@Nonnull Callable<V,T> callable, @CheckForNull Duration performTimeout, @CheckForNull Duration executionTimeout) 
+            throws IOException, T, InterruptedException {
```
jglick: What, no `TimeoutException`?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -832,9 +834,15 @@ public void setJarCache(@Nonnull JarCache jarCache) {
      */
     public <V,T extends Throwable>
     V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
+        return call(callable, UserRequest.DEFAULT_PERFORM_TIMEOUT, UserRequest.DEFAULT_EXECUTION_TIMEOUT);
+    }
+    
+    public <V,T extends Throwable>
+    V call(@Nonnull Callable<V,T> callable, @CheckForNull Duration performTimeout, @CheckForNull Duration executionTimeout)
```
jglick: I do not think `executionTimeout` is necessary. Just keep it to `performTimeout`. No one cares about the difference. A caller simply wants to ensure that a network call does not block forever.

src/main/java/org/jenkinsci/remoting/util/Timeout.java
```diff
@@ -0,0 +1,108 @@
...
+    public static Timeout limit(final long delay, final @Nonnull TimeUnit unit) {
+        final Thread thread = Thread.currentThread();
+        return new Timeout(Timer.get().schedule(new Runnable() {
+            @Override public void run() {
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    Throwable t = new Throwable();
+                    t.setStackTrace(thread.getStackTrace());
+                    LOGGER.log(Level.FINE, "Interrupting " + thread + " after " + delay + " " + unit, t);
+                }
+                thread.interrupt();
```
jglick: I think this utility is inappropriate here.

I introduced it in Pipeline code (later elsewhere) because I was calling predefined APIs (`Channel.call`, ultimately) which offered no timeout (à la `Future.get(long, TimeUnit)`), and from reading the source code I knew that the implementation would _typically_ be inside an `Object.wait()` loop in `Request.call`, which I had no control over and which waited with no bound except a response, channel closing, or a thread interruption. So this was certainly a workaround.

But if you are implementing timeouts in `remoting` itself, there is no reason to resort to that. You can simply make `Request.call` not wait forever. It can `wait` for the configured timeout, and not use the `while` loop.

src/main/java/org/jenkinsci/remoting/util/Timer.java
```diff
@@ -0,0 +1,152 @@
...
+ * Provides a minimal abstraction for locating the ScheduledExecutorService so that we
+ * can modify it's behavior going forward. For instance, to add manageability/monitoring.
+ *
+ * @author Ryan Campbell
+ * @author Jesse Glick
+ * @author Oleg Nenashev
+ * @since TODO
+ */
+@Restricted(NoExternalUse.class)
+public class Timer {
```
jglick: As above, unnecessary if you implement timeouts in the more natural way.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -45,6 +45,7 @@
 import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
+import java.time.Duration;
```
jglick: Could use `long` + `TimeUnit` if you wanted to keep Java 7 compatibility. (But I see no reason not to move to 8 immediately.)

oleg-nenashev: Yes, no reasons from the Jenkins community PoV since the LTS is on Java 8

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: The call gets interrupted now instead of timeout. See the PR TODO list

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: Yes and no. We want to prevent hanging of requests on both sides.

src/main/java/org/jenkinsci/remoting/util/Timeout.java
oleg-nenashev: > But if you are implementing timeouts in remoting itself, there is no reason to resort to that. You can simply make Request.call not wait forever. It can wait for the configured timeout, and not use the while loop.

It will solve hanging of the calls only on one side of the channel. I want to address it on both

jtnord:
>> channel#callAsync() should also support timeouts on the remote side

> Why? I see no need for it.

One reason is that you ask for something to happen on the remote side and it takes a while to process and if it can not return within the default you do not want to waste resources continuing to do it.  
This may not be say an Agent where you may decide resources are cheap, but where remoting is used inter master.
For example - executing a groovy script on a bunch of inter-connected masters - but the groovy script accidentally uses "`while (true)`" without a break clause, using the timeout on the remote side the script can be interrupted whilst it is still running.

oleg-nenashev:
I suppose @jglick just expected the timeout on one side. In this PR I have intentionally added timeouts on both sides. As @jtnord says, such timeout may be useful if the request hangs due to whatever reason, especially since we have a limited threadpool for them.

src/main/java/hudson/remoting/Channel.java
jglick: I interpreted the TODO in the PR as referring to methods added to `VirtualChannel` with `default` implementations. This is just being added to a `class` so there is no Java 8 dependency. You chose the `throws` clause for the method and I am arguing that it should be throwing `TimeoutException`. See signatures of `Future.get` and so on.

src/main/java/hudson/remoting/Channel.java
jglick: I do not see any need for special infrastructure to control delays on the remote side, i.e., in the body of the callable. That is just up to the discretion of whoever is writing that callable—if it is in fact doing something which might block indefinitely, it may instead select a method variant with an appropriate timeout. But from the perspective of an RPC caller, what is important is just that the calling thread is not blocked for too long _in networking_.

src/main/java/org/jenkinsci/remoting/util/Timeout.java
jglick: See above. I think this is not the right approach.

jglick:
w.r.t. `callAsync`

> you ask for something to happen on the remote side and it takes a while to process and if it can not return within the default you do not want to waste resources continuing to do it

So the body of the callable is free to impose any time limit it sees appropriate. For that matter a poorly written callable might be consuming hundreds of megs of heap for no good reason. This is no different from local method calls—not a concern of Remoting.

oleg-nenashev:
> This is no different from local method calls—not a concern of Remoting.

As a maintainer of Remoting, I think having a timeout for such calls is important. Do you block the PR due to that? Or just "IMHO YAGNI"?

stephenc:
Imho I agree with @jglick

Timingboutvthe remote operation should not be a concern of the remoting api... (or at best it should be something that the caller opts-in e.g. By using a wrapper channel)

I am inclined to object, but at this point I will just give a stern look and ask you to critically self-review ;-)

oleg-nenashev:
> or at best it should be something that the caller opts-in

Currently it's opt-in since the default timeout is "no timeout"

stephenc:
> Currently it's opt-in since the default timeout is "no timeout"

But you are littering the API by multiplying methods.

When somebody needs a timeout on the remote execution they probably just want you to provide them with a wrapping callable that does the timeout for them.

That would simplify the API and simplify the implementation.

IMHO less methods to choose from is better. I could probably go so far as to provide a wrapper to the callable for the caller timeout too so that, in effect, the caller just adds the timeouts to their callables

```
ch.call(localTimeout(remoteTimeout(new Callable<...>(...) {...})));
```

but it is fine to keep the local timeout as a parameter

```
ch.call(withTimeout(new Callable<...>(...){...}), localTimeout);
```

as that way the `withTimeout` static callable decorator can also be reused locally without confusion

jglick:
> provide a wrapper to the callable for the caller timeout too

That is essentially what I attempted to do with the `Timeout` utility in Pipeline code, but it is not a good solution here—hence my RFE (which is overshot by this PR). More: https://github.com/jenkinsci/remoting/pull/174#discussion_r124925887

> Do you block the PR due to that?

Well, -0.5. I think this PR is solving problems it did not need to solve, and solving the problems it _did_ need to solve the wrong way.

To reiterate, my original complaint in JENKINS-44785 was that, say,

```java
FilePath f;
try {
    if (f.isDirectory()) {
        // …
    }
} catch (IOException | InterruptedException x) {
    // …
}
```

could block indefinitely due to problems in the network layer, being an example of one of the fundamental fallacies of distributed computing; and my request was for some variant like

```java
FilePath f;
try {
    if (f.isDirectoryInterruptible()) {
        // …
    }
} catch (IOException | InterruptedException | TimeoutException x) {
    // …
}
```

or

```java
FilePath f;
try {
    if (f.isDirectory(30, TimeUnit.SECONDS)) {
        // …
    }
} catch (IOException | InterruptedException | TimeoutException x) {
    // …
}
```

which would make explicit the fact that the network operation might hang for various arbitrary reasons and the caller must be prepared to deal with it.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -553,10 +555,10 @@ public void handle(Command cmd) {
                 lastCommandReceivedAt = receivedAt;
                 if (logger.isLoggable(Level.FINE)) {
                     logger.fine("Received " + cmd);
-                } else if (logger.isLoggable(Level.FINER)) {
-                    logger.log(Level.FINER, "Received command " + cmd, cmd.createdAt);
```
oleg-nenashev: Merge defect To be fixed

everops-alex:
👋 hello I bumped the JIRA ticket, this might be causing downstream issues when running in k8s:
https://issues.jenkins-ci.org/browse/JENKINS-44785?focusedCommentId=384005&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-384005

timja:
close?

jglick:
Maybe, so long as `stale-pr` is added to the corresponding issue.

jeffret-b:
I've been leaving it around as a reference for some possible ideas, thought the actual value is small.

PR #175, opened 2017-06-29
[JENKINS-45023] - Prevent execution of commands on closed or beingClosed channels

oleg-nenashev:
This is a major update of Request execution logic in Remoting Channels, which should improve stability of the channel and prevent hanging of commands if the channel gets closed.

- [x] - `Channel#close()` does not always wait for synchronization to happen. There is a sender status check before the lock gets acquired. TODO: find an issue for that
- [x] - `Channel#isClosingOrClosed()` now returns `null` once the first `Channel#close()` command arrives, we do not even wait till it acquires the instance lock. The API is [used outside Remoting](https://github.com/search?q=org%3Ajenkinsci+isClosingOrClosed&type=Code), but it seems that the change is correct in that cases
- [x] - `Channel#call()` and `Channel#callAsync()` now fail if the channel `isClosingOrClosed()`. These calls implement `UserRequest`, and I do not think there is a valid case for even trying any user-space request
- [x] - Offer new API in `hudson.remoting.Request`, which allows checking the channel state before invoking a call. By default it just checks if the channel is closed (just “fail fast” without command initialization)
- [x] - Implement the new API in `UserRequest`, to prevent low-level API calls on a channel, which `isClosingOrClosed()`

TODO:
- [x] - Automatic Tests
- [ ] - TBD: Create `SystemRequest` class, which allows running requests on channels being closed. Not sure if there is a valid non-YOLO use-case though
- [ ] - TBD: Also fail `RPCRequest`s when shutdown is pending? Otherwise there is a risk of similar hanging in Jenkins (e.g. System script invocation)

Direct issue - [JENKINS-45023](https://issues.jenkins-ci.org/browse/JENKINS-45023). Many other issues should be addressed by it, I'd guess.

@reviewbybees @jglick @stephenc

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -911,6 +911,9 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      *
      * This is an SPI for {@link CommandTransport} implementation to notify
      * {@link Channel} when the underlying connection is severed.
+     * 
+     * Once the call is called {@link #closeRequested} will be set immediately to prevent further executions
+     * of {@link UserRequest}s.
      *
      * @param e
      *      The error that caused the connection to be aborted. Never null.
@@ -918,9 +921,14 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     @java.lang.SuppressWarnings("ToArrayCallWithZeroLengthArrayArgument")
     @SuppressWarnings("ITA_INEFFICIENT_TO_ARRAY") // intentionally; race condition on listeners otherwise
     public void terminate(@Nonnull IOException e) {
+        
+        if (e == null) {
+            throw new IllegalArgumentException("Cause is null. Channel cannot be closed properly in such case");
+        }
+        closeRequested = true;
+        
         try {
-            synchronized (this) {
-                if (e == null) throw new IllegalArgumentException();
+            synchronized (this) {  
                 outClosed = inClosed = e;
                 // we need to clear these out early in order to ensure that a GC operation while
                 // proceding with the close does not result in a batch of UnexportCommand instances
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -327,6 +327,12 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     /*package*/ @Nonnull final ClassFilter classFilter;
 
     /**
+     * Indicates that close of the channel has been requested.
+     * After it it does not make sense to execute new user-space commands.
+     */
+    private boolean closeRequested = false;
+    
+    /**
      * Communication mode used in conjunction with {@link ClassicCommandTransport}.
      * 
      * @since 1.161
@@ -580,11 +586,27 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     }
     
     /**
+     * Get why the sender side of the channel has been closed.
+     * @return Close cause or {@code null} if the sender side is active.
+     *         {@code null} result does not guarantee that the channel is actually operational.
+     * @since TODO
+     */
+    @CheckForNull
+    public final Throwable getSenderCloseCause() {
+        return outClosed;
+    }
+    
+    /**
      * Returns {@code true} if the channel is either in the process of closing down or has closed down.
+     * 
+     * If the result is {@code true}, it means that the channel will be closed at some point by Remoting,
+     * and that it makes no sense to send any new {@link UserRequest}s to the remote side.
+     * Invocations like {@link #call(hudson.remoting.Callable)} and {@link #callAsync(hudson.remoting.Callable)} 
+     * will just fail as well.
      * @since 2.33
      */
     public boolean isClosingOrClosed() {
-        return inClosed != null || outClosed != null;
+        return closeRequested || inClosed != null || outClosed != null;
     }
 
     /**
@@ -832,6 +854,11 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     public <V,T extends Throwable>
     V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
+        if (isClosingOrClosed()) {
+            // No reason to even try performing a user request
+            throw new IOException("Remote call on " + name + " failed. The channel is closing down or has closed down");
+        }
+        
         UserRequest<V,T> request=null;
         try {
             request = new UserRequest<V, T>(this, callable);
@@ -862,6 +889,11 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     public <V,T extends Throwable>
     Future<V> callAsync(final Callable<V,T> callable) throws IOException {
+        if (isClosingOrClosed()) {
+            // No reason to even try performing a user request
+            throw new IOException("Remote call on " + name + " failed. The channel is closing down or has closed down");
+        }
+        
         final Future<UserResponse<V,T>> f = new UserRequest<V,T>(this, callable).callAsync(this);
         return new FutureAdapter<V,UserResponse<V,T>>(f) {
             protected V adapt(UserResponse<V,T> r) throws ExecutionException {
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -327,6 +327,12 @@
     /*package*/ @Nonnull final ClassFilter classFilter;
 
     /**
+     * Indicates that close of the channel has been requested.
+     * After it it does not make sense to execute new user-space commands.
```
jglick: `After that`?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -327,6 +327,12 @@
...
+     */
+    private boolean closeRequested = false;
```
jglick: Could we drop this field and just check `outClosed != null`, since `close` sets both? Or is there a race condition in doing that?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -832,6 +854,11 @@ public void setJarCache(@Nonnull JarCache jarCache) {
      */
     public <V,T extends Throwable>
     V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
+        if (isClosingOrClosed()) {
+            // No reason to even try performing a user request
+            throw new IOException("Remote call on " + name + " failed. The channel is closing down or has closed down");
```
jglick: I would expect this to be `ChannelClosedException`. That is what `Request.call` would have thrown a bit later without this patch, right?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -862,6 +889,11 @@ V call(Callable<V,T> callable) throws IOException, T, InterruptedException {
      */
     public <V,T extends Throwable>
     Future<V> callAsync(final Callable<V,T> callable) throws IOException {
+        if (isClosingOrClosed()) {
+            // No reason to even try performing a user request
+            throw new IOException("Remote call on " + name + " failed. The channel is closing down or has closed down");
```
jglick: Ditto.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1268,12 +1300,14 @@ public void dumpDiagnostics(@Nonnull PrintWriter w) throws IOException {
...
      */
-    public synchronized void close() throws IOException {
+    public void close() throws IOException {
         close(null);
     }
 
     /**
      * Closes the channel.
+     * 
+     * If the channel is not closed, the
```
jglick: Huh?

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1283,32 +1317,40 @@ public synchronized void close() throws IOException {
      *
      * @since 2.8
      */
-    public synchronized void close(@CheckForNull Throwable diagnosis) throws IOException {
+    public void close(@CheckForNull Throwable diagnosis) throws IOException {
         if(outClosed!=null)  return;  // already closed
-
-        try {
```
jglick: [diff for reviewers](https://github.com/jenkinsci/remoting/pull/175/files?w=1#diff-8a0e2be489e004b87cc21e1fc8d17d3dR1320)

src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -90,6 +90,17 @@ public UserRequest(Channel local, Callable<?,EXC> c) throws IOException {
...
     }
 
+    @Override
+    public void checkIfCanBeExecutedOnChannel(Channel channel) throws IOException {
+        // Default check for all requests
+        super.checkIfCanBeExecutedOnChannel(channel);
+        
+        // We also do not want to run UserRequests when the channel is being closed
+        if (channel.isClosingOrClosed()) {
+            throw new IOException("The request cannot be executed on channel " + channel + ". The channel is closing down or has closed down");
```
jglick: Does this not just duplicate the check in `Channel.call` (as well as the check later in `Channel.send`)? I am unclear on why we are doing the same check three times in the course of one call.

(And again this should be `ChannelClosedException`.)

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: Not exactly. `ChannelClosedException` would have been thrown if the channel is closed, but not "is closing". Probably there is no difference from user perspective though

src/main/java/hudson/remoting/UserRequest.java
oleg-nenashev: Unfortunately this low-level method is being called not only within `Channel#call`, but also from `Request#call()` directly (also a public API). Hence I had to add a check there.

I could just remove checks in `Channel#call()`, but in such case some logic in requests will be invoked (e.g. UserRequest constructor, export object recorder, etc.). I would prefer to fail fast there

src/main/java/hudson/remoting/Channel.java
oleg-nenashev: Yes, race condition. `outClosed` is set within the `synchronize(this)` block. If there is a deadlock or delay, the commands will be accepted even if the `close()` command has been called somewhere

src/main/java/hudson/remoting/Channel.java
jglick: I think from the user (~ plugin author) perspective either the channel is open for business, or it is closed and the request cannot be satisfied. There is no reason to distinguish a channel in the process of closing.

src/main/java/hudson/remoting/UserRequest.java
jglick: > from `Request#call()` directly (also a public API)

While it might have been `public` by some accident, clearly it was only intended to be called from `Channel.call`, and any other call site should be considered unsupported and the behavior undefined.

oleg-nenashev: @jglick Yes and No. `Channel#call()` is a wrapper method for calling `Callable`s by wrapping them into `UserRequest`s. All other Request types and user-created requests can be executed only via the low level API. So I can imagine use-cases when somebody needs to call them directly. from outside Remoting 

Withing Remoting these methods are being used to invoke RPCRequests from RemoteInvocationHandler. Ans such calls should be definitely protected as well.

oleg-nenashev:
After some consideration I want to do the same for `RPCRequest`s in `RemoteInvocationHandler`s (all RPC requests), at least for synchronous calls. Otherwise things like CommandLaunchers in Jenkins core may still get into the issue. But I am not sure it is perfectly safe since some calls may actually run on channels being closed. @reviewbybees @jglick WDYT?

src/test/java/hudson/remoting/ChannelTest.java
```diff
@@ -176,4 +183,156 @@ public void testDiagnostics() throws Exception {
...
+     * @throws Exception Test Error
+     */
+    @Bug(45023)
+    public void testShouldNotAcceptUserRequestsWhenIsBeingClosed() throws Exception {
+        
+        // Create a sample request to the channel
+        final Callable<Void, Exception> testPayload = new NeverEverCallable();
+        UserRequest<Void, Exception> delayedRequest = new UserRequest<>(channel, testPayload);
+        
+        try (ChannelCloseLock lock = new ChannelCloseLock(channel)) {
```
oleg-nenashev: This auto-closeable lock is a bit tricky, but we will need it in other tests.

jglick:
> I want to do the same for `RPCRequest`s in `RemoteInvocationHandler`s (all RPC requests), at least for synchronous calls.

Seems reasonable. IIRC this would apply to any methods called on exported objects.

oleg-nenashev:
> Seems reasonable. IIRC this would apply to any methods called on exported objects.

Exactly. So I am quite aware about doing by default - high risk of collateral damage on wrapped objects like `ExportTable` or classloaders. On the other hand, any RPC calls against the closing channel are subjects to race conditions anyway. 

The tests pass when I enable checks for `RPCRequest`s and also `ChunkRequest`s

oleg-nenashev:
@reviewbybees done
Created https://issues.jenkins-ci.org/browse/JENKINS-45294 for `RPCRequest`s

PR #176, opened 2017-06-30
Add IOHub introduction version

batmat:
Very small improvement


oleg-nenashev:
That's why @stephenc should consider using TODO instead of FIXME. But yes, I like the latter comment annotation as well.

oleg-nenashev:
🐝 and @reviewbybees done

PR #177, opened 2017-06-30
[JENKINS-45233] - Log errors when Response message cannot be delivered due to the closed channel

oleg-nenashev:
Just noticed some issues in command execution logic...

- [x] [JENKINS-45233](https://issues.jenkins-ci.org/browse/JENKINS-45233) - Command processing does not print errors when the Response cannot be delivered due to the closed channel
- [x] Add FINE/FINER logging for better diagnostics of incoming command processing

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -528,11 +528,18 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
         transport.setup(this, new CommandReceiver() {
             public void handle(Command cmd) {
                 commandsReceived++;
-                lastCommandReceivedAt = System.currentTimeMillis();
-                if (logger.isLoggable(Level.FINE))
+                long receivedAt = System.currentTimeMillis();
+                lastCommandReceivedAt = receivedAt;
+                if (logger.isLoggable(Level.FINE)) {
                     logger.fine("Received " + cmd);
+                } else if (logger.isLoggable(Level.FINER)) {
+                    logger.log(Level.FINER, "Received command " + cmd, cmd.createdAt);
+                }
                 try {
                     cmd.execute(Channel.this);
+                    if (logger.isLoggable(Level.FINE)) {
+                        logger.log(Level.FINE, "Completed command {0}. It took {1}ms", new Object[] {cmd, System.currentTimeMillis() - receivedAt});
+                    }
                 } catch (Throwable t) {
                     logger.log(Level.SEVERE, "Failed to execute command " + cmd + " (channel " + Channel.this.name + ")", t);
                     logger.log(Level.SEVERE, "This command is created here", cmd.createdAt);
--- a/src/main/java/hudson/remoting/Request.java
+++ b/src/main/java/hudson/remoting/Request.java
@@ -25,6 +25,7 @@ package hudson.remoting;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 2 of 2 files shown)


src/main/java/hudson/remoting/Request.java
```diff
@@ -346,13 +347,17 @@ public void run() {
                         rsp.createdAt.initCause(createdAt);
 
                     synchronized (channel) {// expand the synchronization block of the send() method to a check
-                        if(!channel.isOutClosed())
+                        if(!channel.isOutClosed()) {
                             channel.send(rsp);
+                        } else {
+                            // Propagate exception to the catch clause instead of supressing it
```
jglick: typo

src/main/java/hudson/remoting/Request.java
```diff
@@ -346,13 +347,17 @@ public void run() {
                         rsp.createdAt.initCause(createdAt);
 
                     synchronized (channel) {// expand the synchronization block of the send() method to a check
-                        if(!channel.isOutClosed())
+                        if(!channel.isOutClosed()) {
```
jglick: arguably easier to read if test is inverted

src/main/java/hudson/remoting/Request.java
```diff
@@ -346,13 +347,17 @@ public void run() {
...
                             channel.send(rsp);
+                        } else {
+                            // Propagate exception to the catch clause instead of supressing it
+                            throw new ClosedChannelException();
```
jglick: Is there a known `cause` here?

src/main/java/hudson/remoting/Request.java
```diff
@@ -346,13 +347,17 @@ public void run() {
...
+                        }
                     }
                 } catch (IOException e) {
                     // communication error.
                     // this means the caller will block forever
-                    logger.log(Level.SEVERE, "Failed to send back a reply",e);
+                    logger.log(Level.SEVERE, "Failed to send back a reply to the request " + this, e);
```
jglick: BTW `SEVERE` is probably excessive.

oleg-nenashev: Yeah, I would also drop it to WARNING

src/main/java/hudson/remoting/Channel.java
```diff
@@ -528,11 +528,18 @@ protected Channel(@Nonnull ChannelBuilder settings, @Nonnull CommandTransport tr
         transport.setup(this, new CommandReceiver() {
             public void handle(Command cmd) {
                 commandsReceived++;
-                lastCommandReceivedAt = System.currentTimeMillis();
-                if (logger.isLoggable(Level.FINE))
+                long receivedAt = System.currentTimeMillis();
+                lastCommandReceivedAt = receivedAt;
+                if (logger.isLoggable(Level.FINE)) {
                     logger.fine("Received " + cmd);
+                } else if (logger.isLoggable(Level.FINER)) {
```
jtnord: http://javaadventure.blogspot.co.uk/2016/12/the-costs-of-java-utils-logging.html

src/main/java/hudson/remoting/Request.java
oleg-nenashev: Actually the check is not required at all. `Channel#send()` does the check

src/main/java/hudson/remoting/Channel.java
```diff
@@ -528,11 +528,18 @@ protected Channel(@Nonnull ChannelBuilder settings, @Nonnull CommandTransport tr
...
+                    logger.log(Level.FINER, "Received command " + cmd, cmd.createdAt);
```
jtnord: unless cmd.createdAt is an Exception

jtnord:
`Command` does not implement `toString()` - yet it is used in lots of logging - are you intentionally wanting to print the hashCode - in which case I would suggest you replace it's usage with [`System.identityHashCode(it)`](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object-)
if you are expecting somethign else then shouldn't you make toString() abstract and force its implementation according to a spec (defined in the classes javadoc)?

src/main/java/hudson/remoting/Request.java
```diff
@@ -346,13 +347,12 @@ public void run() {
                         rsp.createdAt.initCause(createdAt);
 
                     synchronized (channel) {// expand the synchronization block of the send() method to a check
```
jtnord: comment implies this should be removed.

oleg-nenashev: right

oleg-nenashev:
> Command does not implement toString() - yet it is used in lots of logging - are you intentionally wanting to print the hashCode - in which case I would suggest you replace it's usage with System.identityHashCode(it)
if you are expecting somethign else then shouldn't you make toString() abstract and force its implementation according to a spec (defined in the classes javadoc)?

Implementations like `UserRequest` and `RPCRequest` have `toString()` implementations, hence I do not consider it as a big problem. But yes, it will be printing class names and hashes for system calls. Not a big deal

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
Errors are unrelated

PR #178, opened 2017-07-05
[JENKINS-45522] Introduced ClassFilter.appendDefaultFilter

jglick:
[JENKINS-45522](https://issues.jenkins-ci.org/browse/JENKINS-45522)

Allows us to get rid of some awful reflection in core.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -7,10 +7,10 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
@@ -87,6 +87,17 @@ public abstract class ClassFilter {
      * unless the context guarantees the trust between two channels.
      */
     public static final ClassFilter DEFAULT = createDefaultInstance();
+
+    /**
+     * Adds an additional exclusion to {@link #DEFAULT}.
+     * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
+     */
+    public static void appendDefaultFilter(Pattern filter) {
+        if (System.getProperty(FILE_OVERRIDE_LOCATION_PROPERTY) == null) {
+            ((RegExpClassFilter) DEFAULT).add(filter.toString());
+        }
+    }
 
     /**
      * No filtering whatsoever.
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -89,6 +89,17 @@ public final Class check(Class c) {
     public static final ClassFilter DEFAULT = createDefaultInstance();
 
     /**
+     * Adds an additional exclusion to {@link #DEFAULT}.
+     * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
+     */
+    public static void appendDefaultFilter(Pattern filter) {
+        if (System.getProperty(FILE_OVERRIDE_LOCATION_PROPERTY) == null) {
+            ((RegExpClassFilter) DEFAULT).add(filter.toString());
```
oleg-nenashev: seems to be a leaky API presumption. FindBugs will grumble once enabled. Add class check and checked exception at least 🐜

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -89,6 +89,17 @@ public final Class check(Class c) {
     public static final ClassFilter DEFAULT = createDefaultInstance();
 
     /**
+     * Adds an additional exclusion to {@link #DEFAULT}.
+     * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
+     */
+    public static void appendDefaultFilter(Pattern filter) {
```
oleg-nenashev: Throw checked exceptions to be able to propagate errors if we need to add some (like the class assumption failure below)

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -89,6 +89,17 @@ public final Class check(Class c) {
     public static final ClassFilter DEFAULT = createDefaultInstance();
 
     /**
+     * Adds an additional exclusion to {@link #DEFAULT}.
+     * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
+     */
```
oleg-nenashev: since TODO

src/main/java/hudson/remoting/ClassFilter.java
jglick: I do not believe there is any need to check the cast. `DEFAULT` cannot be anything other than a `RegExpClassFilter`. Could adjust return type of `createDefaultInstance` to make this more apparent.

src/main/java/hudson/remoting/ClassFilter.java
jglick: What checked exceptions? I see no need for any.

jglick:
I have responded to all comments and do not believe they are real issues.

oleg-nenashev:
@jglick IIRC we have discussed it during one of the dailys ~1 month ago. I believe we need a way to propagate errors from `appendDefaultFilter()` if we need to harden the implementation (e.g. to add blacklists). If you do not want to change the implementation, let me know. But I keep 🐛 for the current state

jglick:
> if we need to harden the implementation (e.g. to add blacklists)

What does this mean? The API’s sole purpose is to blacklist entries.

> If you do not want to change the implementation

I do not follow what you are asking for or why.

oleg-nenashev:
Created #185 with my vision of this change

PR #179, opened 2017-07-07
[JENKINS-38696] - Add File Leak Detector + Allow Disabling File cache in JARUrlConnection

oleg-nenashev:
This change adds functionality, which allows fixing some failing tests in Windows due to the file leaks. See [JENKINS-38696](https://issues.jenkins-ci.org/browse/JENKINS-38696)

- [x] - Add the `surefire-with-fileleak-detector` profile, which runs tests with File Leak Detector (and its Web UI)
- [x] - Add system property, which disables File caching in `ResourceImageInJar`. I tried to do it in a generic way (ChannelBuilder): https://github.com/oleg-nenashev/remoting/commit/144922e5de65244729fbd8de2a610daa90ea13e1 . It adds too much boilerplate code for a feature, which is not supposed to be used in production
- [x] PrefetchingTest now disables file caching on Windows and fails fast if there are leaked file descriptors in `tearDown()`
- [ ] - Document changes and the new property in the developer documentation 

In the future I consider creating FileLeakDetector test rule for JUnit (with automatic checks), but it is a long shot.

This change fixes only one test, [JENKINS-45349](https://issues.jenkins-ci.org/browse/JENKINS-45349) has to be also fixed for others

@reviewbybees @jtnord

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -581,9 +581,6 @@ THE SOFTWARE.
     </profile>
     <profile>
       <id>surefire-with-fileleak-detector</id>
-      <activation>
-        <activeByDefault>true</activeByDefault>
-      </activation>
       <properties>
         <maven-surefire-plugin.agent>-javaagent:${project.build.directory}\agents\file-leak-detector.jar=http=20000</maven-surefire-plugin.agent>
       </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -494,6 +494,7 @@ THE SOFTWARE.
         <configuration>
           <trimStackTrace>false</trimStackTrace> <!-- SUREFIRE-1226 workaround -->
           <rerunFailingTestsCount>4</rerunFailingTestsCount>
+          <argLine>${maven-surefire-plugin.agent}</argLine>
         </configuration>
       </plugin>
       <plugin>
@@ -574,6 +575,43 @@ THE SOFTWARE.
                 <argument>http://timestamp.comodoca.com/rfc3161</argument>
               </arguments>
             </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>surefire-with-fileleak-detector</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <properties>
+        <maven-surefire-plugin.agent>-javaagent:${project.build.directory}\agents\file-leak-detector.jar=http=20000</maven-surefire-plugin.agent>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>copy-fileleak-detector</id>
+                <phase>process-test-classes</phase>
+                <goals>
+                  <goal>copy</goal>
+                </goals>
+                <configuration>
+                  <artifactItems>
+                    <artifactItem>
+                      <groupId>org.kohsuke</groupId>
+                      <artifactId>file-leak-detector</artifactId>
+                      <version>1.8</version>
+                      <classifier>jar-with-dependencies</classifier>
+                      <outputDirectory>${project.build.directory}/agents</outputDirectory>
+                      <destFileName>file-leak-detector.jar</destFileName>
+                    </artifactItem>
+                  </artifactItems>
+                </configuration>
+              </execution>
+            </executions>
           </plugin>
         </plugins>
       </build>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -578,5 +579,42 @@ THE SOFTWARE.
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>surefire-with-fileleak-detector</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
oleg-nenashev: Leftover

pom.xml
```diff
@@ -494,6 +494,7 @@ THE SOFTWARE.
         <configuration>
           <trimStackTrace>false</trimStackTrace> <!-- SUREFIRE-1226 workaround -->
           <rerunFailingTestsCount>4</rerunFailingTestsCount>
+          <argLine>${maven-surefire-plugin.agent}</argLine>
```
jtnord: should add `${argLine}` to allow passing of arguments from the command line still (such as running with less memory than 2GB :-)

pom.xml
```diff
@@ -578,5 +579,39 @@ THE SOFTWARE.
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>surefire-with-fileleak-detector</id>
+      <properties>
+        <maven-surefire-plugin.agent>-javaagent:${project.build.directory}\agents\file-leak-detector.jar=http=20000</maven-surefire-plugin.agent>
```
jtnord: do you not want to use `/` rather than `\` otherwise this will blow up on unixy type systems

src/main/java/hudson/remoting/ResourceImageInJar.java
```diff
@@ -43,11 +53,24 @@
         return new FutureAdapter<byte[],URL>(_resolveJarURL(channel)) {
             @Override
             protected byte[] adapt(URL jar) throws ExecutionException {
+                final URLConnection c;
                 try {
-                    return Util.readFully(toResourceURL(jar,resourcePath).openStream());
+                    URL url = toResourceURL(jar,resourcePath);
+                    c = url.openConnection();
+                    if (DISABLE_FILE_CACHING_IN_JAR_CONNECTION) {
```
jtnord: Why would we not just always disable this?
WIthin the same JVM we should be be called multiple times to adapt the same JAR should we - if we are then would we not get classloading issues?

oleg-nenashev: > if we are then would we not get classloading issues?

We won't get classloading issues, but multiple resource loading will be extremely ineffective from the performance PoV. So it's a test-only hack

src/test/java/hudson/remoting/PrefetchingTest.java
```diff
@@ -65,20 +73,12 @@ private File toFile(URL url) {
...
-                FileUtils.deleteDirectory(dir);
-                return;
-            } catch (IOException e) {
-                if (i==3)   throw e;
-                Thread.sleep(1000);
-            }
-        }
+        
+        // Cleanup the temporary cache and ensure we have no locked files left
+        FileUtils.deleteDirectory(dir);
```
jtnord: why not use 
`@Rule public TemproaryFolder temp = new TemppraryFolder` ?

oleg-nenashev: In order to use Rule, the Class and the Entire RMITestBase need to be reworked to JUnit4. I definitely +1 regarding doing it, but even in such case we will have to alter the test behavior

src/main/java/hudson/remoting/ResourceImageInJar.java
jtnord: ack

src/test/java/hudson/remoting/PrefetchingTest.java
jtnord: ack

pom.xml
oleg-nenashev: crap, I have forgotten to push it

oleg-nenashev:
thanks @jtnord . I realized I have not pushed the POM changes

oleg-nenashev:
@jtnord Fixed POM

oleg-nenashev:
Tests now fail...
* Error: Could not find or load main class `${maven-surefire-plugin.agent}`

oleg-nenashev:
Does not happen on my machine, probably inclusion of ${args} changed the behavior

oleg-nenashev:
On hold till I get the PR builder running

jeffret-b:
JENKINS-38696 has been resolved with changes submitted. @oleg-nenashev, is there any need to keep this PR open? It seems to have been superseded by a number of events since it was opened.

jeffret-b:
No response and this doesn't look current.

PR #181, opened 2017-07-20
[JENKINS-45349] - Test Client: Allow disabling file caching in HelloGetResource and HelloGetResources.

oleg-nenashev:
This change adds an optional parameter, which disables file caching. My plan is to reuse it in Remoting tests in order to suppress some issues in Windows tests.

https://issues.jenkins-ci.org/browse/JENKINS-45349

@reviewbybees @jtnord

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.26</version>
+    <version>1.37</version>
   </parent>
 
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -39,6 +39,10 @@ THE SOFTWARE.
     This module is separate because some of the tests
     require code in a jar file not available in the default classpath.
   </description>
+  
+  <properties>
+    <java.level>7</java.level>
```
jtnord: 7?

oleg-nenashev: 7 for now, Remoting is on Java 7, will update later to 8

PR #185, opened 2017-08-07
[JENKINS-45522] - Introduce ClassFilter.appendDefaultFilter

oleg-nenashev:
It is a modified version of #178 from @jglick with better error propagation

https://issues.jenkins-ci.org/browse/JENKINS-45522

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -93,13 +93,15 @@ public abstract class ClassFilter {
             DEFAULT = createDefaultInstance();
         } catch (ClassFilterException ex) {
             LOGGER.log(Level.SEVERE, "Default class filter cannot be initialized. Remoting will not start", ex);
-            throw new IllegalStateException("Default class filter cannot be initialized", ex);
+            throw new ExceptionInInitializerError(ex);
         }
     }
 
     /**
      * Adds an additional exclusion to {@link #DEFAULT}.
+     * 
      * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * This API is not supposed to be used anywhere outside Jenkins core, calls for other sources may be rejected later.
      * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
      * @throws ClassFilterException Filter pattern cannot be applied.
      *                              It means either unexpected processing error or rejection by the internal logic.
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -86,14 +86,26 @@ public abstract class ClassFilter {
      * A set of sensible default filtering rules to apply,
      * unless the context guarantees the trust between two channels.
      */
-    public static final ClassFilter DEFAULT = createDefaultInstance();
+    public static final ClassFilter DEFAULT;
+
+    static {
+        try {
+            DEFAULT = createDefaultInstance();
+        } catch (ClassFilterException ex) {
+            LOGGER.log(Level.SEVERE, "Default class filter cannot be initialized. Remoting will not start", ex);
+            throw new IllegalStateException("Default class filter cannot be initialized", ex);
+        }
+    }
 
     /**
      * Adds an additional exclusion to {@link #DEFAULT}.
      * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
      * @param filter a regular expression for {@link Class#getName} which, if matched according to {@link Matcher#matches}, will blacklist the class
+     * @throws ClassFilterException Filter pattern cannot be applied.
+     *                              It means either unexpected processing error or rejection by the internal logic.
+     * @since TODO
      */
-    public static void appendDefaultFilter(Pattern filter) {
+    public static void appendDefaultFilter(Pattern filter) throws ClassFilterException {
         if (System.getProperty(FILE_OVERRIDE_LOCATION_PROPERTY) == null) {
             ((RegExpClassFilter) DEFAULT).add(filter.toString());
         }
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -174,31 +197,36 @@ public final Class check(Class c) {
...
-                } else {
-                    blacklistPatterns[i] = Pattern.compile(patterns[i]);
+        void add(String pattern) throws ClassFilterException {
+
+            if (OPTIMIZE1.matcher(pattern).matches()) {
+                // this is a simple startsWith test, no need to slow things down with a regex
+                blacklistPatterns.add(pattern.substring(1, pattern.length() - 2).replace("[.]", "."));
+            } else if (OPTIMIZE2.matcher(pattern).matches()) {
+                // this is a simple startsWith test, no need to slow things down with a regex
+                blacklistPatterns.add(pattern.substring(3, pattern.length() - 4));
```
oleg-nenashev: Perhaps it also makes sense to handle out of the bounds exception in this PR. Not a new issue though

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -86,14 +86,26 @@ public final Class check(Class c) {
...
      * unless the context guarantees the trust between two channels.
      */
-    public static final ClassFilter DEFAULT = createDefaultInstance();
+    public static final ClassFilter DEFAULT;
+
+    static {
+        try {
+            DEFAULT = createDefaultInstance();
+        } catch (ClassFilterException ex) {
+            LOGGER.log(Level.SEVERE, "Default class filter cannot be initialized. Remoting will not start", ex);
```
jglick: Suffices to throw `ExceptionInInitializerError`. (Why Java does not let you throw checked exceptions in a `static` block to begin with, I do not know.)

src/test/java/hudson/remoting/DefaultClassFilterTest.java
```diff
@@ -146,10 +146,12 @@ public void describeMismatch(Object item, Description description) {
             return new BlackListMatcher();
         }
 
-        public boolean matches(Object item) {
+        public boolean matches(Object item)  {
```
jglick: ?

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -187,21 +199,27 @@ public static void appendDefaultFilter(Pattern filter) {
...
             if (OPTIMIZE1.matcher(pattern).matches()) {
                 // this is a simple startsWith test, no need to slow things down with a regex
                 blacklistPatterns.add(pattern.substring(1, pattern.length() - 2).replace("[.]", "."));
             } else if (OPTIMIZE2.matcher(pattern).matches()) {
                 // this is a simple startsWith test, no need to slow things down with a regex
                 blacklistPatterns.add(pattern.substring(3, pattern.length() - 4));
             } else {
+                try {
+                    Pattern regex = Pattern.compile(pattern);
+                } catch (PatternSyntaxException ex) {
```
jglick: Impossible, because we are either in `createDefaultInstance`, for which the patterns are statically defined and known to be valid; or in `appendDefaultFilter`, which takes a `Pattern` which is by definition well-formed.

oleg-nenashev:
@jglick 

> Introduces an exception type into the API which cannot be thrown.

I disagree, this case is valid as we discussed in the previous ticket. The exception may be thrown with the newly introduced API, e.g. when there is a decision to blacklist some patterns (e.g. `.*`). There is also a case for Groovy calls invoking private APIs (similar to reflection calls in the core)

src/main/java/hudson/remoting/ClassFilter.java
oleg-nenashev: I throw the unchecked exception. `ExceptionInInitializerError ` is definitely better though

src/main/java/hudson/remoting/ClassFilter.java
oleg-nenashev: It is still a good protective measure against reflection calls and Groovy hackers

jglick:
I do not see any reason to write code to block, say, addition of `.*` as a blacklist pattern because there is at any given time expected to be at most one call to this API, from Jenkins core, and that can be expected to be correct; if you want to spend time fine-tuning it anyway, the unchecked `IllegalArgumentException` would be appropriate. Similarly, there seems to be no purpose in spending time writing code to block other call sites (suffices to note its intended usage in Javadoc), but if you care to write a call stack check, the unchecked `IllegalStateException` would be appropriate (or `@Restricted` could be enhanced for this purpose in the future).

oleg-nenashev:
@jglick Added usage warning to Javadoc. Later we could add the call site prevention if we really want to. `IllegalStateException` would be enough to break the caller code, but I want to enforce checks in the API caller code.

rysteboe:
@oleg-nenashev LGTM overall, but CI failures make me hesitant to approve.

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -86,7 +86,32 @@ public final Class check(Class c) {
...
+            LOGGER.log(Level.SEVERE, "Default class filter cannot be initialized. Remoting will not start", ex);
+            throw new ExceptionInInitializerError(ex);
+        }
+    }
+
+    /**
+     * Adds an additional exclusion to {@link #DEFAULT}.
+     * 
+     * Does nothing if the default list has already been customized via {@link #FILE_OVERRIDE_LOCATION_PROPERTY}.
+     * This API is not supposed to be used anywhere outside Jenkins core, calls for other sources may be rejected later.
```
jglick: I suppose it would be legitimate to call from a plugin if we found that there was a dangerous class in a library bundled with a plugin (but not in Jenkins core).

jglick:
What is the status here?

oleg-nenashev:
> Maybe instead of an API method we should just declare some resource name as per ClassLoader.getResources(String), like META-INF/hudson.remoting.ClassFilter.DEFAULT, containing one entry per line (excluding blanks and #)? This would solve signature issues and be clearer to use. The tricky part is listening to changes in uberClassLoader after plugins are loaded and after dynamic plugin installation.

I think it's a good idea, but having binary API would not hurt as well. E.g. for System Groovy Scripting , etc., etc. So I would probably merge it and create a follow-up ticket

src/main/java/hudson/remoting/ClassFilter.java
oleg-nenashev: Maybe. I just want to avoid advertising it for now.
If we hit a real use-case, we can reconsider it.

oleg-nenashev:
Let's keep the current change as is, I will add more features later if required

src/main/java/hudson/remoting/ClassFilter.java
oleg-nenashev: fixed

oleg-nenashev:
Merging without squash since there are 2 authors. It is not a backporting candidate anyway

PR #186, opened 2017-08-11
[JENKINS-46259] - Log all linkage errors during user requests

olivergondza:
Linkage errors are a symptom of subtle classloading problems that are hart to investigate on its own. Propagating the exception upwards and over the channel only does not help as the stacktraces are dispersed all over the place (build logs, server log, discarded by caller, etc.) making the problem discovery and chain-of-the-events reconstruction harder.

I presume this is something of an interest for instance administrator so let's log it at one place.

### DIFF
--- a/src/main/java/hudson/remoting/UserRequest.java
+++ b/src/main/java/hudson/remoting/UserRequest.java
@@ -51,6 +51,8 @@ import javax.annotation.Nonnull;
  */
 final class UserRequest<RSP,EXC extends Throwable> extends Request<UserResponse<RSP,EXC>,EXC> {
 
+    private static final Logger LOGGER = Logger.getLogger(UserRequest.class.getName());
+
     private final byte[] request;
     
     @Nonnull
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -204,6 +206,9 @@ public void checkIfCanBeExecutedOnChannel(Channel channel) throws IOException {
                 } finally {
                     Thread.currentThread().setContextClassLoader(old);
                 }
+            } catch (LinkageError e) {
+                LOGGER.log(Level.INFO, "LinkageError while performing " + toString(), e);
```
oleg-nenashev: Maybe warning?

olivergondza: Sure, why not.

olivergondza:
Tests passing locally, not sure what is wrong with CI builds.

oleg-nenashev:
Yes, I should finally fix builders

PR #187, opened 2017-08-14
[JENKINS-38711] Add UncaughtExceptionHandlers to the various remoting threads

rysteboe:
@reviewbybees

https://issues.jenkins-ci.org/browse/JENKINS-38711

### DIFF
--- a/src/main/java/hudson/remoting/forward/CopyThread.java
+++ b/src/main/java/hudson/remoting/forward/CopyThread.java
@@ -42,7 +42,6 @@ final class CopyThread extends Thread {
                 int len;
                 while ((len = in.read(buf)) > 0)
                     out.write(buf, 0, len);
-                throw new IllegalArgumentException();
             } finally {
                 in.close();
                 out.close();
--- a/src/main/java/hudson/remoting/forward/CopyThread.java
+++ b/src/main/java/hudson/remoting/forward/CopyThread.java
@@ -20,17 +20,17 @@ final class CopyThread extends Thread {
      * Callers are responsible for closing the input and output streams.
      */
     public CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination) {
-        this(threadName, in, out, termination, 0);
+        this(threadName, in, out, termination, 5);
     }
 
-    private CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination, int previousTries) {
+    private CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination, int remainingTries) {
         super(threadName);
         this.in = in;
         this.out = out;
         setUncaughtExceptionHandler((t, e) -> {
-            if (previousTries < 5) {
-                LOGGER.log(Level.FINE, "Uncaught exception in CopyThread " + t + ", retrying copy", e);
-                new CopyThread(threadName, in, out, termination, previousTries + 1).start();
+            if (remainingTries > 0) {
+                LOGGER.log(Level.WARNING, "Uncaught exception in CopyThread " + t + ", retrying copy", e);
+                new CopyThread(threadName, in, out, termination, remainingTries - 1).start();
             } else {
                 LOGGER.log(Level.SEVERE, "Uncaught exception in CopyThread " + t + ", out of retries", e);
                 termination.run();
--- a/src/main/java/hudson/remoting/forward/CopyThread.java
+++ b/src/main/java/hudson/remoting/forward/CopyThread.java
@@ -27,7 +27,7 @@ final class CopyThread extends Thread {
         setUncaughtExceptionHandler((t, e) -> {
             if (previousTries < 5) {
                 LOGGER.log(Level.FINE, "Uncaught exception in CopyThread " + t + ", retrying copy", e);
-                new CopyThread(threadName, in, out, termination, previousTries + 1).run();
+                new CopyThread(threadName, in, out, termination, previousTries + 1).start();
             } else {
                 LOGGER.log(Level.SEVERE, "Uncaught exception in CopyThread " + t + ", out of retries", e);
                 termination.run();
@@ -42,6 +42,7 @@ final class CopyThread extends Thread {
                 int len;
                 while ((len = in.read(buf)) > 0)
                     out.write(buf, 0, len);
+                throw new IllegalArgumentException();
             } finally {
                 in.close();
                 out.close();
--- a/src/main/java/hudson/remoting/forward/CopyThread.java
+++ b/src/main/java/hudson/remoting/forward/CopyThread.java
@@ -20,6 +20,9 @@ final class CopyThread extends Thread {
         super(threadName);
         this.in = in;
         this.out = out;
+        setUncaughtExceptionHandler((t, e) -> {
+            LOGGER.log(Level.WARNING, "Uncaught exception in CopyThread " + t, e);
+        });
     }
 
     public void run() {
--- a/src/main/java/hudson/remoting/forward/CopyThread.java
+++ b/src/main/java/hudson/remoting/forward/CopyThread.java
@@ -16,12 +16,22 @@ final class CopyThread extends Thread {
     private final InputStream in;
     private final OutputStream out;
 
-    public CopyThread(String threadName, InputStream in, OutputStream out) {
+    public CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination) {
+        this(threadName, in, out, termination, 0);
+    }
+
+    private CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination, int previousTries) {
         super(threadName);
         this.in = in;
         this.out = out;
         setUncaughtExceptionHandler((t, e) -> {
-            LOGGER.log(Level.WARNING, "Uncaught exception in CopyThread " + t, e);
+            if (previousTries < 5) {
+                LOGGER.log(Level.FINE, "Uncaught exception in CopyThread " + t + ", retrying copy", e);
+                new CopyThread(threadName, in, out, termination, previousTries + 1).run();
+            } else {
+                LOGGER.log(Level.SEVERE, "Uncaught exception in CopyThread " + t + ", out of retries", e);
+                termination.run();
+            }
         });
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/hudson/remoting/forward/CopyThread.java
```diff
@@ -20,6 +20,9 @@ public CopyThread(String threadName, InputStream in, OutputStream out) {
         super(threadName);
         this.in = in;
         this.out = out;
+        setUncaughtExceptionHandler((t, e) -> {
+            LOGGER.log(Level.WARNING, "Uncaught exception in CopyThread " + t, e);
+        });
```
oleg-nenashev: I think the thread needs to be restarted in such case

oleg-nenashev: Ideally creators of this thread should pass additional termination handler:

* PortForwarder - do handling and close the socket ?
* ForwarderFactory#connect - Interrupt RemoteOutputStream ?

src/main/java/hudson/remoting/forward/PortForwarder.java
```diff
@@ -57,6 +57,14 @@ public PortForwarder(int localPort, Forwarder forwarder) throws IOException {
         // mark as a daemon thread by default.
         // the caller can explicitly cancel this by doing "setDaemon(false)"
         setDaemon(true);
+        setUncaughtExceptionHandler((t, e) -> {
```
oleg-nenashev: Note that `run()` below creates a nested `Port forwarding session from... ` thread. Maybe it also needs a handler

src/main/java/hudson/remoting/forward/CopyThread.java
rysteboe: If we're restarting the thread, this should probably also used a buffered input and output stream, no?  Otherwise data will be lost or mangled.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -270,6 +270,7 @@ public boolean verify(String s, SSLSession sslSession) {
     public boolean failIfWorkDirIsMissing = WorkDirManager.DEFAULT_FAIL_IF_WORKDIR_IS_MISSING;
 
     public static void main(String... args) throws Exception {
+        Thread.setDefaultUncaughtExceptionHandler((t, e) -> LOGGER.log(Level.SEVERE, "Unhandled exception in thread " + t, e));
```
oleg-nenashev: 🐛 with such approach logging won't be initialized on the Master side.

src/main/java/hudson/remoting/forward/CopyThread.java
rysteboe: Thinking about this - how would that restart work?  The `finally` block in the previous `run` would close the input and output streams, meaning that by the time it tries again, it can't use those streams.  Perhaps the input and output stream closures should be moved elsewhere?  Or is there a simpler solution that I'm not thinking of?

src/main/java/hudson/remoting/forward/CopyThread.java
```diff
@@ -29,6 +42,7 @@ public void run() {
                 int len;
                 while ((len = in.read(buf)) > 0)
                     out.write(buf, 0, len);
+                throw new IllegalArgumentException();
```
oleg-nenashev: Needs some comments and maybe exception message around.
It should be likely IllegalStateException instead.

rysteboe: That was just there for testing purposes - pretty sure this shouldn't actually be erroring out?

src/main/java/hudson/remoting/forward/CopyThread.java
```diff
@@ -16,10 +16,23 @@
...
+    }
+
+    private CopyThread(String threadName, InputStream in, OutputStream out, Runnable termination, int previousTries) {
         super(threadName);
         this.in = in;
         this.out = out;
+        setUncaughtExceptionHandler((t, e) -> {
+            if (previousTries < 5) {
+                LOGGER.log(Level.FINE, "Uncaught exception in CopyThread " + t + ", retrying copy", e);
+                new CopyThread(threadName, in, out, termination, previousTries + 1).start();
```
rysteboe: The problem that I see here is that the `finally` block in the previous thread would close the input and output streams, meaning that by the time it tries again, it can't use those streams. Perhaps the input and output stream closures should be moved elsewhere? Or is there a simpler solution that I'm not thinking of?

rysteboe: I did a thing for this.  It might be stupid.

rysteboe:
@oleg-nenashev, can you take a look with the changes?

src/main/java/hudson/remoting/forward/CopyThread.java
oleg-nenashev: Maybe it should be a WARNING even for allowed retries. Even if the copythread recovers, it may be a dataloss if the input stream has been read without writing. Or maybe the thread should be able to send cause to the termination handler? It would be able to process it in a custom way then. It would require some extra API, but it seems to be reasonable.

src/main/java/hudson/remoting/forward/ForwarderFactory.java
```diff
@@ -93,4 +111,6 @@ private Object writeReplace() {
      * Role that's willing to open a socket to arbitrary node nad forward that to the other side.
      */
     public static final Role ROLE = new Role(ForwarderFactory.class);
+
+    private static final Logger LOGGER = Logger.getLogger(ForwarderFactory.class.getName());
```
oleg-nenashev: I would move variables to the top in order to follow the Oracle code style

src/main/java/hudson/remoting/forward/CopyThread.java
rysteboe: I played around here with trying to prevent data loss (or rewriting data) but I couldn't come up with a solid solution.  You can reset the `InputStream` but I could not figure out a way to determine how many bytes were successfully written to the output stream to restart the `InputStream` from there (short of literally writing one byte at a time).  I think that approach can still be used for errors on reading the `InputStream`, but I don't think that any reset will be safe if an exception is thrown while attempting to write.

All of this leads me to question if we should be trying to restart the thread from here if an exception is thrown while writing.  It may be better to do it from the `Runnable` passed into `CopyThread` so that the streams can be reinitialized.  I'm going to try to prototype that approach when I get back home tonight).

oleg-nenashev: Or just propagate the error to the caller thread as I've proposed above.

rysteboe:
Updated.

src/main/java/hudson/remoting/forward/CopyThread.java
rysteboe: I guess I'm confused about what should happen here.  Propagating the error is complicated - it's simply enough to pass the error into the termination handler, of course, but I don't know why the termination handler itself cares about the specific error.  From what I understand, and there's a good chance that my understanding is poor, is that the termination handler needs to clean up resources regardless of the specifics of the error.  Said cleanup will likely also cause other failures in the parent thread, as closed sockets can no longer be read from, etc.  Rethrowing the error in the parent thread is problematic at best because it would require waiting on a future (I think?) and ultimately doesn't seem worthwhile to me since the socket/stream closures will have no trouble causing errors themselves (which are in turn handled).  

What am I missing here?  (Yes, I've spent far too much time thinking about/writing random code around this.)

oleg-nenashev:
One datapoint from CI of another pull request:

```
Running org.jenkinsci.remoting.protocol.IOHubTest
Sep 03, 2017 3:44:34 AM org.jenkinsci.remoting.protocol.IOHub$OnReady run
SEVERE: [IOHub#1: Worker[channel:sun.nio.ch.ServerSocketChannelImpl[/0.0.0.0:42834]] / noReadyCallbackIfInterestRemoved(org.jenkinsci.remoting.protocol.IOHubTest)-3] Listener org.jenkinsci.remoting.protocol.IOHubTest$6@2f3b5c8e propagated an uncaught NullPointerException
java.lang.NullPointerException
	at org.jenkinsci.remoting.protocol.IOHubTest$6.ready(IOHubTest.java:227)
	at org.jenkinsci.remoting.protocol.IOHub$OnReady.run(IOHub.java:721)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)

Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.288 sec - in org.jenkinsci.remoting.protocol.IOHubTest
```

rysteboe:
@oleg-nenashev, I had a question above wrt `CopyThread` that I think got buried.  Copy pasting down here:

You:
> Maybe it should be a WARNING even for allowed retries. Even if the copythread recovers, it may be a dataloss if the input stream has been read without writing. Or maybe the thread should be able to send cause to the termination handler? It would be able to process it in a custom way then. It would require some extra API, but it seems to be reasonable.

Me:

> Propagating the error is complicated - it's simply enough to pass the error into the termination handler, of course, but I don't know why the termination handler itself cares about the specific error. From what I understand, and there's a good chance that my understanding is poor, is that the termination handler needs to clean up resources regardless of the specifics of the error. Said cleanup will likely also cause other failures in the parent thread, as closed sockets can no longer be read from, etc. Rethrowing the error in the parent thread is problematic at best because it would require waiting on a future (I think?) and ultimately doesn't seem worthwhile to me since the socket/stream closures will have no trouble causing errors themselves (which are in turn handled).  What am I missing here?

oleg-nenashev:
I'll try to recover the context and to prototype my proposal. Probably I am wrong

oleg-nenashev:
> Propagating the error is complicated - it's simply enough to pass the error into the termination handler, of course, but I don't know why the termination handler itself cares about the specific error.

My idea was to cause chain termination of threads and then a graceful restart of the logic

> Rethrowing the error in the parent thread is problematic at best because it would require waiting on a future (I think?) and ultimately doesn't seem worthwhile to me since the socket/stream closures will have no trouble causing errors themselves (which are in turn handled).

I agree. At least the behavior does not get worse that the current one

rysteboe:
@reviewbybees done

PR #190, opened 2017-08-18
[JENKINS-37567] - Add option to specify certchain, enforce certificate checks

oleg-nenashev:
It should finalize [JENKINS-37567](https://issues.jenkins-ci.org/browse/JENKINS-37567).

- [x] - Add `certchain` option (requires a custom profile now)
  * See https://github.com/apache/maven-plugins/pull/125 and https://github.com/apache/maven-shared/pull/24
- [x] - Start **REALLY** validating the signing correctness when the release flow is enabled

I have checked the default options, the change does not impact the local dev build and @kohsuke 's release flow. I am quite aware about the `jarsigner -tsa` step in the `release` profile, but it seems to be harmless on my local machine.

@reviewbybees @olamy @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,8 @@ THE SOFTWARE.
     <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <findbugs.failOnError>false</findbugs.failOnError>
+    <!-- TODO: Required to override by a version with certchain support (MJARSIGNER-53) on @oleg-nenashev's machine. Remove once it's in upstream -->
+    <maven-jarsigner-plugin.version>1.4</maven-jarsigner-plugin.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -568,12 +577,39 @@ THE SOFTWARE.
...
+                <phase>verify</phase>
+                <goals>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <processMainArtifact>true</processMainArtifact>
+                  <processAttachedArtifacts>false</processAttachedArtifacts>
+                  <certs>true</certs>
+                  <errorWhenNotSigned>true</errorWhenNotSigned>
+                  <arguments>-strict</arguments> <!--otherwise certificate chains will be ignored-->
```
oleg-nenashev: @olamy Likely I should create a MJARSIGNER issue for adding this parameter. I would argue it should be enabled by default in Java's jarsigner or in at least Maven JarSigner Plugin, but it would be a potentially breaking change

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
Merging in order to proceed with the release

PR #191, opened 2017-08-22
[JENKINS-46341] - Disable Windows build for now

oleg-nenashev:
It just does not make sense while [JENKINS-38696](https://issues.jenkins-ci.org/browse/JENKINS-38696) is not fixed

@reviewbybees


oleg-nenashev:
@reviewbybees done

jglick:
Yet the build still times out…

oleg-nenashev:
Yes, Linux flow was also broken. On my own Docker container it fails with different error

PR #192, opened 2017-08-25
fix jar cache from user.home to user.dir - Fixes official jenkins docker image

clouless:
### Problem

 * :bangbang: `user.home` info IS NOT TAKEN from `$HOME` EnvVar, but from `/etc/passwd` instead.
 * :white_check_mark: `user.dir` IS TAKEN from `$HOME` EnvVar

Due to most people using dockerized builds and setting HOME EnvVar **it would be better to use `user.dir` instead of `user.home` for the JarCache** to avoid errors like this one:

image

### Where does it happen?

In official docker image `jenkinsci/jenkins:2.75` (when working with kubernetes-plugin).

image

### How to test

Test under Ubuntu 16.04

```
docker run -i -t ubuntu:16.04 bash
```

Inside Docker as root:

```
apt-get install openjdk-8-jdk vim
groupadd -g 10000 jenkins
useradd -c "Jenkins user" -d /home/jenkins -u 10000 -g 10000 -m jenkins
```

Change /etc/passwd from `jenkins:x:10000:10000:Jenkins user:/home/jenkins:` to `jenkins:x:10000:10000:Jenkins user:/home/jenkins2:`.

Switch to Jenkins User and create `Test.java`:

```
su - jenkins
cd /home/jenkins
vim /home/jenkins/Test.java
```

```java
import java.io.File;

public class Test {

  public static void main(String[] args){
      System.out.println(System.getProperty("user.home"));
      System.out.println(System.getProperty("user.dir"));
      File file = new File(System.getProperty("user.dir"),".jenkins/cache/jars");
      file.mkdir();
      System.out.println(file.getAbsoluteFile());
    }
}
```

Compile and run:

```
javac Test.java
java Test
```

It Prints out:

```
/home/jenkins2
/home/jenkins
/home/jenkins/.jenkins/cache/jars
```

### DIFF
--- a/src/main/java/hudson/remoting/JarCache.java
+++ b/src/main/java/hudson/remoting/JarCache.java
@@ -29,6 +29,7 @@ public abstract class JarCache {
     
     @Nonnull
     /*package*/ static JarCache getDefault() {
+        DEFAULT_NOWS_JAR_CACHE_LOCATION.getAbsoluteFile()
         return new FileSystemJarCache(DEFAULT_NOWS_JAR_CACHE_LOCATION, true);
     }
     
--- a/src/main/java/hudson/remoting/JarCache.java
+++ b/src/main/java/hudson/remoting/JarCache.java
@@ -25,11 +25,10 @@ public abstract class JarCache {
      * Default JAR cache location for disabled workspace Manager.
      */
     /*package*/ static final File DEFAULT_NOWS_JAR_CACHE_LOCATION = 
-        new File(System.getProperty("user.home"),".jenkins/cache/jars");
+        new File(System.getProperty("user.dir"),".jenkins/cache/jars");
     
     @Nonnull
     /*package*/ static JarCache getDefault() {
-        DEFAULT_NOWS_JAR_CACHE_LOCATION.getAbsoluteFile()
         return new FileSystemJarCache(DEFAULT_NOWS_JAR_CACHE_LOCATION, true);
     }
     


oleg-nenashev:
There was a ticket for that: https://issues.jenkins-ci.org/browse/JENKINS-18578

It was fixed in 3.8 by adding support of [Work directories](https://github.com/jenkinsci/remoting/blob/master/docs/workDir.md). So a correct approach would be to just define `-workDir` or `-jarCache`.

Regarding the change, I am aware about possible regressions in some cases with the default behavior

clouless:
This happens on the Jenkins (Master) side, not the Slave side. So do you suggest to add `-workDir` startup parameters to Jenkins, to have the jarCache default behaviour disabled? I think `-workDir` is just a parameter for the remoting.jar and will not be interpreted by OPTS for Jenkins?

oleg-nenashev:
> This happens on the Jenkins (Master) side, not the Slave side. 

Then yes, I should review it with Master in mind

clouless:
ok cool. Even though the error is unlikely to happen on wide spectrum. It is cleaner that way.
And this only happens when you start the docker image with an UID different than 1000.
I did this with kubernetes and everything else worked fine, but this error popped up, since a user with UID!=1000 has no /etc/passwd entry to point to homeDir.

oleg-nenashev:
Sorry for the delay. I hope to find some time this week

clouless:
no worries :) I find the overall response time of you guys very fast. Regarding the amount of tickets and prs to handle for jenkins as a whole :) I have waited much much longer for a first reaction on issues/prs in other projects. So just take your time, we are in no rush :)

oleg-nenashev:
I have submitted https://github.com/jenkinsci/remoting/pull/197, which probably fixes the root cause of the issue (JENKINS-45755) on the master side.

But maybe this enhancement makes sense anyway. @recena also propose using "/tmp", so we could try several destinations when `-workDir` and `-jarCache` are not set

clouless:
@oleg-nenashev nice. The more directories tried out before giving up the better. But if you want to user userHome then please use `user.dir` because of the `/etc/passwd` problem. You can close this PR if you solved it elsewhere :)

oleg-nenashev:
This is something I would definitely want to integrate, but in more generic way. I will keep it open for now

sschuberth:
> `user.dir` IS TAKEN from `$HOME` EnvVar

That doesn't seem to be correct in general @clouless, as `user.dir` is documented to be the [user working directory](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html) which is [not necessarily within the user's home directory](https://stackoverflow.com/a/16239152/1127485).

As such it seems to be better to use `System.getProperty("user.home")` unless it's blank or `?`, then try `System.getenv("HOME")`, then try `System.getenv("USERPROFILE")` (for Windows).

clouless:
@sschuberth I will close that PR as won't fix since it has been open for more than two years.

jeffret-b:
This dates back to before I took over as the primary Remoting maintainer. If there's something here  still that is valuable to do, please re-open, create a new PR, or something and we can see about moving it forward.

clouless:
I have solved it now for me somehow, cannot remember anymore how but I found a workaround. 
I still think that `user.dir` is the way to go (at least for linux people). But I can currently not invest time into this, so from my side this can stay closed.

PR #195, opened 2017-09-06
Bump to latest jenkinsci/pom

batmat:
Downstream to https://github.com/jenkinsci/pom/pull/16

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.38</version>
+    <version>1.39-20170902.001419-2</version>
   </parent>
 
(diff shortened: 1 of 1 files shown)


batmat:
WUT. IIUC the build fails because the pom can't be found, but this succeeded with the exact same value for https://github.com/jenkinsci/maven-interceptors/pull/14, on the same infra. Possibly some temporary flakiness, going to retrigger the build to see if the error is persistent.

batmat:
@oleg-nenashev this fails because the pom SNAPSHOT cannot be found, but it works on https://github.com/jenkinsci/maven-interceptors/pull/14 

I'm probably missing something obvious, like a typo or so, if you can throw a quick look would be great. Thanks

oleg-nenashev:
@batmat it looks fine. Maybe there is a special symbol or something. Anyway, I am fine with the upstream release since the build is passing

oleg-nenashev:
This change does not make the things worse

PR #196, opened 2017-09-09
Javadoc cleanup

oleg-nenashev:
I have recently added Javadoc publishing for Remoting, so some polishing would not hurt.

- [x] Fix some glitches in Javadoc
- [x] Add `@since` versions to the most of `@since TODO` tags

@reviewbybees


oleg-nenashev:
@reviewbybees done

PR #167, opened 2017-06-19
[JENKINS-44973] - Ensure that JAR Cache management works correctly in Launcher

oleg-nenashev:
It is an unreleased regression caused by me in https://github.com/jenkinsci/remoting/pull/166 . Filed it as [JENKINS-44973](https://issues.jenkins-ci.org/browse/JENKINS-44973).

Discovered it during testing the recent remoting version on SSH. Effectively launches with STDERR/STDOUT (SSH) and socket (tests) are failing now.

Also annotated all the things to prevent it in the future.

@reviewbybees

PR #197, opened 2017-09-13
[JENKINS-45755] - Make JARCache nullable in Channel

oleg-nenashev:
It is a rework of my #167 fix for [JENKINS-44973](https://issues.jenkins-ci.org/browse/JENKINS-44973). During that fix I investigated a codebase and decided that it is better to make `Channel#jarCache` non-null to prevent misbehavior of agent initializers. But then we got a regression in the core, which was also initializing channels...

This fix makes the following adjustments:

- [x] - `Channel` now supports null JAR Cache
- [x] - JAR Loading logic now checks for nulls and fails with IOExceptions instead of NPEs
- [x] - `ChannelBuilder` deprecated `withChannel(null)`, but there are replacement methods
- [x] - Default JAR Cache initialization throws IOException instead of Runtime one where possible

I have not written a unit test for it, because it requires starting Jenkins or Remoting Test Client in an instance with non-writable home. I will do it once I have a better dockerized test framework for Remoting issues.
 
https://issues.jenkins-ci.org/browse/JENKINS-45755

@reviewbybess @recampbell @recena @Vlatombe

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -317,7 +317,12 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      */
     /*package*/ final ClassLoader baseClassLoader;
 
-    @Nonnull
+    /**
+     * JAR Resolution Cache.
+     * Can be {@code null} if caching disabled for this channel.
+     * In such case some classloading operations may be rejected.
+     */
+    @CheckForNull
     private JarCache jarCache;
 
     /*package*/ final JarLoaderImpl jarLoader;
@@ -515,12 +520,10 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
         this.underlyingOutput = transport.getUnderlyingStream();
         
         // JAR Cache resolution
-        JarCache effectiveJarCache = settings.getJarCache();
-        if (effectiveJarCache == null) {
-            effectiveJarCache = JarCache.getDefault();
-            logger.log(Level.CONFIG, "Using the default JAR Cache: {0}", effectiveJarCache);
+        this.jarCache = settings.getJarCache();
+        if (this.jarCache == null) {
+            logger.log(Level.CONFIG, "JAR Cache is not defined for channel {0}", name);
         }
-        this.jarCache = effectiveJarCache;
 
         this.baseClassLoader = settings.getBaseLoader();
         this.classFilter = settings.getClassFilter();
@@ -838,9 +841,12 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
 
     /**
      * If this channel is built with jar file caching, return the object that manages this cache.
+     * @return JAR Cache object. {@code null} if JAR caching is disabled
      * @since 2.24
+     * @since 3.10 JAR Cache is Nonnull
+     * @since 3.12 JAR Cache made nullable again due to <a href="https://issues.jenkins-ci.org/browse/JENKINS-45755">JENKINS-45755</a>
      */
-    @Nonnull
+    @CheckForNull
     public JarCache getJarCache() {
         return jarCache;
     }
@@ -851,6 +857,9 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      *
      * So to best avoid performance loss due to race condition, please set a JarCache in the constructor,
      * unless your call sequence guarantees that you call this method before remote classes are loaded.
+     *
+     * @param jarCache New JAR Cache to be used.
+     *                 Cannot be {@code null}, JAR Cache disabling on a running channel is not supported.
      * @since 2.24
      */
     public void setJarCache(@Nonnull JarCache jarCache) {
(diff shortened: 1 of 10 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -838,9 +841,12 @@ public boolean preloadJar(ClassLoader local, URL... jars) throws IOException, In
 
     /**
      * If this channel is built with jar file caching, return the object that manages this cache.
+     * @return JAR Cache object. {@code null} if JAR caching is disabled
      * @since 2.24
+     * @since 3.10 JAR Cache is Nonnull
+     * @since 3.12 JAR Cache made nullable again due to <a href="https://issues.jenkins-ci.org/browse/JENKINS-45755">JENKINS-45755</a>
      */
-    @Nonnull
+    @CheckForNull
```
oleg-nenashev: It's kinda breaking change in API. I will make sure it is documented in changelogs

src/main/java/hudson/remoting/Engine.java
```diff
@@ -263,7 +263,11 @@ public synchronized void startEngine() throws IOException {
                 throw new IOException("Cannot find the JAR Cache location");
             }
             LOGGER.log(Level.FINE, "Using standard File System JAR Cache. Root Directory is {0}", jarCacheDirectory);
-            jarCache = new FileSystemJarCache(jarCacheDirectory, true);
+            try {
+                jarCache = new FileSystemJarCache(jarCacheDirectory, true);
+            } catch (IllegalArgumentException ex) {
+                throw new IOException("Failed to initialize FileSystem JAR Cache in " + jarCacheDirectory, ex);
```
recena: @oleg-nenashev I don't have experience here, in advance my apologies for my ignorance. Could we use something like `/tmp` by default in case `jarCacheDirectory` has some problem?

When I say `/tmp`, I mean the temporary folder defined in the O.S.

src/test/java/hudson/remoting/PrefetchingTest.java
```diff
@@ -215,7 +215,11 @@ private ForceJarLoad(Checksum sum) {
         public Void call() throws IOException {
             try {
                 Channel ch = Channel.current();
-                ch.getJarCache().resolve(ch,sum1,sum2).get();
+                final JarCache jarCache = ch.getJarCache();
+                if (jarCache == null) {
+                    throw new IOException("Cannot Force JAR load, JAR cache is disabled");
```
recena: 👍

src/main/java/hudson/remoting/Engine.java
oleg-nenashev: @recena We could do that. Default would be a breaking change for Docker images and other things, which rely on the current default location, but it may be applicable as a fallback. 

Good news is that `-workDir` mode changes the location.

src/main/java/hudson/remoting/FileSystemJarCache.java
```diff
@@ -54,7 +55,7 @@ public FileSystemJarCache(@Nonnull File rootDir, boolean touch) {
         try {
             Util.mkdirs(rootDir);
         } catch (IOException ex) {
-            throw new RuntimeException("Root directory not writable: " + rootDir);
+            throw new IllegalArgumentException("Root directory not writable: " + rootDir);
```
olivergondza: Can we append cause so admins has a chance to identify the cause and fix?

olivergondza:
So what would happen when the cache dir is not writeable? Warning at agent startup and classes/jars will be loaded over and over but never cached?

oleg-nenashev:
@olivergondza 

> So what would happen when the cache dir is not writeable? Warning at agent startup and classes/jars will be loaded over and over but never cached?

Agent startup will fail in such case, but it was a default behavior before 3.8 with work dirs.
I have tried to implement no-cache agent a while ago, it just does not work and overloads the channel and master.

olivergondza:
Oh, so the problem is we are not able to create the cache dir on _master_ side which is something we should not be doing on the first place.

oleg-nenashev:
> Oh, so the problem is we are not able to create the cache dir on master side which is something we should not be doing on the first place

@olivergondza It depends. Some Jenkins implementations may require it if a master performs classloading from another instance over Remoting. But it is not a use-case of the Jenkins core, so, yeah, we should not be doing it in Jenkins core itself.

My mistake was that I assumed that any `Channel` instance should have access to the cache, and I have not taken the master-side implementation into account. Unfortunately, the presumption passed local tests with flying colors :(

I am going to create a better framework for testing Remoting with Docker.

oleg-nenashev:
I am going to release it tomorrow if there is no negative feedback
@reviewbybees done

recampbell:
How does this change the user experience when the user home directory is not writable?

oleg-nenashev:
@recampbell The change disables JAR Caching in the Channel by default. It reverts the initialization behavior to what we had in Remoting 3.7 (2.60.x)...

* If a ChannelBuilder does not explicitly require the JarCache, it gets disabled. So the home writeability does not get checked 
* Standard agent types specify the JAR Cache on all paths. As in Remoting 3.7, such agents require home directory to be writeable. Jenkins agent crashes on startup in this case since 2.61 (https://github.com/jenkinsci/remoting/commit/62aad35f64c46d0c41a5d464ccba8c78b784f9d7 from @olivergondza )
* Master's channels and Data I/O Channels do not define JAR cache, hence they do not check writeability
* If a `ChannelBuilder` on a master requires JARCache (e.g. to classload from external instance), `FileSystemJarCache` will check the writeability of the cache and throw Runtime exception. Same as in 2.61+

PR #198, opened 2017-09-14
Build Javadoc only once during the packaging phase.

oleg-nenashev:
Otherwise Javadoc is being generated twice, and then it is being uploaded twice. 
Common users have no package replace permissions in Artifactory, and the release just fails after that.

@reviewbybees @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -283,7 +283,7 @@ THE SOFTWARE.
           <providerClass>${hudson.sign.providerClass}</providerClass>
           <providerArg>${hudson.sign.providerArg}</providerArg>
           <tsa>${hudson.sign.tsa}</tsa>
-          <!-- 
+          <!--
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -283,7 +283,7 @@ THE SOFTWARE.
           <providerClass>${hudson.sign.providerClass}</providerClass>
           <providerArg>${hudson.sign.providerArg}</providerArg>
           <tsa>${hudson.sign.tsa}</tsa>
-          <!-- 
+          <!--
```
oleg-nenashev: Unrelated, I have not disabled the plugin in Atom on my Windows machine. But it should not hurt

jglick:
To confirm, try `deploy`ing a snapshot. Will succeed with or without IIRC, but you can check the logs for uploading messages.

oleg-nenashev:
Confirmed

PR #194, opened 2017-09-02
[JENKINS-46341] - Fix Unix PR Builder

oleg-nenashev:
- [x] - In WorkDir Tests verify that the R/W/X flags can be really set
- [x] - Skip tests of obsolete protocols in the PR builder (JNLP NIO tests are flaky)
- [ ] - The common PR build should run all tests
- [ ] - Provide reference Docker image for testing Remoting

PR #199, opened 2017-09-14
POM: Do not Downgrade plugins from parent pom 1.39

oleg-nenashev:
@reviewbybees @batmat

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,6 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.codehaus.mojo</groupId>
       <artifactId>animal-sniffer-annotations</artifactId>
-      <version>1.15</version>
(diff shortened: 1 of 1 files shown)


batmat:
Why not remove too:
* m-release-p :  https://github.com/oleg-nenashev/remoting/blob/12769dc5fd5c5dc4b7d4597a51ff99f31c89e9fd/pom.xml#L421
* m-surefire-p : https://github.com/oleg-nenashev/remoting/blob/12769dc5fd5c5dc4b7d4597a51ff99f31c89e9fd/pom.xml#L465

oleg-nenashev:
Release plugin can be removed. Regarding Surefire, I would keep it since there are pending pull requests which change its configuration

oleg-nenashev:
It won't 😿 
#194

PR #109, opened 2016-09-26
[JENKINS-37566] - FindBugs Cleanup. Part #1

oleg-nenashev:
Addresses 17 issues in the library, mostly related to File operations, concurrency and NITs. 34 issues are still need to be fixed.

@reviewbybees and @stephenc hence some issues come from his code

PR #206, opened 2017-10-28
[JENKINS-47714] - Introduce SerializableOnlyOverRemoting and cleanup FindBugs in Channel#current().

oleg-nenashev:
Discovered it during the experiments with #109

`Channel#current()` uses thread-local storage to determine the current channel. It returns null if the channel does not exist Some writeReplace/readObject/etc. serialization logic retrieves the channel in order to export the object via ExportTable. Obviously, such operations will fail if we try to serialize the object without Remoting context.

I propose to add a new interface to verify that serialization logic is being invoked for the remoting context and hence to avoid undesired NPEs.

https://issues.jenkins-ci.org/browse/JENKINS-47714

@reviewbybees @rysteboe @jglick @stephenc

### DIFF
--- a/src/main/java/hudson/remoting/Callable.java
+++ b/src/main/java/hudson/remoting/Callable.java
@@ -27,6 +27,7 @@ import org.jenkinsci.remoting.RoleSensitive;
 
 import java.io.Serializable;
 
+//TODO: Make it SerializableOnlyOverRemoting?
 /**
  * Represents computation to be done on a remote system.
  *
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1722,7 +1722,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      *
      * @return Current channel
      * @throws IllegalStateException the calling thread has no associated channel.
-     * @since TODO
+     * @since 3.14
      * @see org.jenkinsci.remoting.SerializableOnlyOverRemoting
      */
     @Nonnull
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -42,7 +42,6 @@ import java.io.InputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
-import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
@@ -1176,7 +1175,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             this.level = level;
         }
         public Void call() throws RuntimeException {
-            Channel.currentOrIllegalState().maximumBytecodeLevel = level;
+            Channel.currentOrFail().maximumBytecodeLevel = level;
             return null;
         }
 
@@ -1633,7 +1632,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     private static final class IOSyncer implements Callable<Object, InterruptedException> {
         @Override
         public Object call() throws InterruptedException {
-            Channel.currentOrIllegalState().syncLocalIO();
+            Channel.currentOrFail().syncLocalIO();
             return null;
         }
 
@@ -1727,7 +1726,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * @see org.jenkinsci.remoting.SerializableOnlyOverRemoting
      */
     @Nonnull
-    public static Channel currentOrIllegalState() throws IllegalStateException {
+    public static Channel currentOrFail() throws IllegalStateException {
         final Channel ch = CURRENT.get();
         if (ch == null) {
             final Thread t = Thread.currentThread();
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -42,6 +42,7 @@ import java.io.InputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.URL;
@@ -1175,7 +1176,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             this.level = level;
         }
         public Void call() throws RuntimeException {
-            Channel.current().maximumBytecodeLevel = level;
+            Channel.currentOrIllegalState().maximumBytecodeLevel = level;
             return null;
         }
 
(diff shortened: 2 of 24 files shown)


src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Gets current channel or fails with {@link NotSerializableException}.
+     *
+     * This method is designed for serialization/deserialization methods in the channel.
+     * @return Current channel
+     * @throws NotSerializableException the calling thread has no associated channel.
+     *      In such case the object cannot be serialized.
+     */
+    @Nonnull
+    default Channel getChannelForSerDes() throws NotSerializableException {
```
oleg-nenashev: no way to make it protected, unfortunately

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1716,6 +1718,24 @@ public static Channel current() {
         return CURRENT.get();
```
dwnusbaum: Should this method be deprecated, or maybe update the JavaDoc to indicate that `Channel#currentOrIllegalState` or `SerializableOnlyOverRemoting#getChannelForSerDes` should probably be used instead?

oleg-nenashev: The method itself is fine, API users just need to use it correctly. I would keep it

src/main/java/hudson/remoting/Channel.java
```diff
@@ -1716,6 +1718,24 @@ public static Channel current() {
...
+    /**
+     * Gets current channel or fails with {@link IllegalStateException}.
+     *
+     * @return Current channel
+     * @throws IllegalStateException the calling thread has no associated channel.
+     * @since TODO
+     * @see org.jenkinsci.remoting.SerializableOnlyOverRemoting
+     */
+    @Nonnull
+    public static Channel currentOrIllegalState() throws IllegalStateException {
```
jglick: Maybe `currentOrFail`?

src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
```diff
@@ -0,0 +1,72 @@
...
+package org.jenkinsci.remoting;
+
+import hudson.remoting.Channel;
+
+import javax.annotation.Nonnull;
+import java.io.NotSerializableException;
+import java.io.Serializable;
+
+/**
+ * This interface indicates objects, which are {@link Serializable} only for sending over the Remoting {@link Channel}.
```
jglick: delete `,` in this context (restrictive clause)

src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Usually it means that the object requires export of the class via {@link Channel}
+ * and {@code hudson.remoting.ExportTable}.
+ * Attempts to serialize the instance of this interface for different purposes lead to undefined behavior.
+ *
+ * @author Oleg Nenashev
+ * @since TODO
+ */
+public interface SerializableOnlyOverRemoting extends Serializable {
```
jglick: BTW `hudson.util.StreamTaskListener` (and perhaps `TaskListener` generally) ought to implement this interface.

src/main/java/hudson/remoting/Callable.java
```diff
@@ -27,6 +27,7 @@
 
 import java.io.Serializable;
 
+//TODO: Make it SerializableOnlyOverRemoting?
```
jglick: I was going to say that this marker should be applied only to concrete implementations which actually define a `writeReplace` etc. that assumes Remoting, since there could be a `NotReallyRoleSensitiveCallable` which is serialized for some other purpose. But then I saw various callable classes which need the current channel, for which having access to the `default` method would be useful. Anyway IIUC there is no actually harm done in implementing this interface but then serializing the object in some other way, provided you do not actually call the `default` method.

BTW if you mark it here, do so in core also on `FileCallable`.

src/main/java/hudson/remoting/Request.java
```diff
@@ -51,15 +52,16 @@
      *
      * @param channel
      *      The local channel. From the view point of the JVM that
-     *      {@link #call(Channel) made the call}, this channel is
+     *      {@link #call(Channel)} made the call, this channel is
```
jglick: `@linkplain` is what you were looking for

src/main/java/hudson/remoting/Request.java
```diff
@@ -51,15 +52,16 @@
...
      *      the remote channel.
      * @return
      *      the return value will be sent back to the calling process.
      * @throws EXC
      *      The exception will be forwarded to the calling process.
      *      If no checked exception is supposed to be thrown, use {@link RuntimeException}.
      */
-    protected abstract RSP perform(Channel channel) throws EXC;
+    @CheckForNull
```
jglick: `@Nullable` actually.

src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
jglick: `@Restricted(ProtectedExternally.class)` would work I think.

jglick: The name is pretty awkward. Looking at `serDes` I could not figure out for a while what that could mean. Why not simply `getChannel()` or `currentChannel()`?

dwnusbaum: FWIW `serDes` also confused me at first, I like Jesse's suggestions.

oleg-nenashev: Agreed, will patch it. For me SerDes is just a common term. E.g. https://en.wikipedia.org/wiki/SerDes

src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
oleg-nenashev: likely. Will add a follow-up. Also `RemoteProcImpl` & Co

src/main/java/org/jenkinsci/remoting/SerializableOnlyOverRemoting.java
oleg-nenashev: Added getChannelForSerialization. If I use simple method names, there may be collisions with interface implementations

oleg-nenashev:
It seems there is a glitch. I resolved the merge conflict, build passed. But the UI still reports it. Will merge manually

oleg-nenashev:
I created https://issues.jenkins-ci.org/browse/JENKINS-47896 for follow-ups

PR #205, opened 2017-10-17
Add insecure flag to allow bypassing SSL hostname verification and cert checks

MattLud:
Initial proposal - Let me know what sort of unit tests you'd like to see or if you want the whitespace changes reverted.

PR #210, opened 2017-11-03
Allow disabling HTTPs cert validation of JNLP endpoint when starting Remoting from Main

oleg-nenashev:
It is an attempt to improve https://github.com/jenkinsci/remoting/pull/205 and to reuse existing code from @stephenc . Originally the HTTPs certificate validation has been done in `hudson.remoting.Launcher`, but it was missing in `hudson.remoting.jnlp.Main`. So it was not possible to start a Java Web Start agent in the insecure mode. In https://github.com/jenkinsci/remoting/pull/205 @MattLud needs that.

How does it work now?
* When `-disableHttpsCertValidation` is passed from `Main`, the JNLP Endpoint HTTPs cert verification is disabled.
* JNLP agent ('Launcher.java') also sets the flag when `-noCertificateCheck` is passed

I also did some refactoring in order to prevent code duplication.

@reviewbybees @MattLud @Wadeck

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -352,7 +352,8 @@ public class Launcher {
                 }
             }
             if (noCertificateCheck) {
-                //TODO: Rename option?
+		// Generally it is not required since the default settings have been changed anyway.
+		// But we set it up just in case there are overrides somewhere in the logic
                 jnlpArgs.add("-disableHttpsCertValidation");
             }
             try {
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -41,6 +41,8 @@ import javax.net.ssl.TrustManagerFactory;
 
 import org.jenkinsci.remoting.engine.WorkDirManager;
 import org.jenkinsci.remoting.util.IOUtils;
+import org.jenkinsci.remoting.util.https.NoCheckHostnameVerifier;
+import org.jenkinsci.remoting.util.https.NoCheckTrustManager;
 import org.w3c.dom.Document;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
@@ -191,6 +193,12 @@ public class Launcher {
                     "certificate file to read.", forbids = "-noCertificateCheck")
     public List<String> candidateCertificates;
 
+    /**
+     * Disables HTTPs Certificate validation of the server when using {@link org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver}.
+     * This option is managed by the {@code -noCertificateCheck} option.
+     */
+    private boolean noCertificateCheck = false;
+
     public InetSocketAddress connectionTarget = null;
 
     @Option(name="-connectTo",usage="make a TCP connection to the given host and port, then start communication.",metaVar="HOST:PORT")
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -242,6 +251,12 @@ public Engine createEngine() {
             engine.setJarCache(new FileSystemJarCache(jarCache,true));
         engine.setNoReconnect(noReconnect);
         engine.setKeepAlive(!noKeepAlive);
+
+        if (disableHttpsCertValidation) {
+            LOGGER.log(WARNING, "Certificate validation for HTTPs endpoints is disabled");
```
Wadeck: Could be nice to make the rest of the calls also consistent. In that class there are currently 3 ways to create logs : LOGGER.fine(...), LOGGER.log(INFO,..., and LOGGER.log(Level.WARNING,...

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -407,8 +440,28 @@ static URLConnection openURLConnection(URL url, String credentials, String proxy
...
+
+                try {
+                    SSLContext ctx = SSLContext.getInstance("TLS");
+                    ctx.init(null, new TrustManager[]{new NoCheckTrustManager()}, new SecureRandom());
+                    sslSocketFactory = ctx.getSocketFactory();
+
+                    httpsConnection.setHostnameVerifier(new NoCheckHostnameVerifier());
+                    httpsConnection.setSSLSocketFactory(sslSocketFactory);
+                } catch (KeyManagementException | NoSuchAlgorithmException ex) {
+                    System.err.println(String.format("Error setting disableHttpsCertValidation; %s", ex.getMessage()));
```
Wadeck: Is there a reason to use System.err here instead of LOGGER ?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -407,8 +440,28 @@ static URLConnection openURLConnection(URL url, String credentials, String proxy
             String encoding = Base64.encode(proxyCredentials.getBytes("UTF-8"));
             con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
         }
-        if (con instanceof HttpsURLConnection && sslSocketFactory != null) {
-            ((HttpsURLConnection) con).setSSLSocketFactory(sslSocketFactory);
+
+        if (con instanceof HttpsURLConnection) {
+            final HttpsURLConnection httpsConnection = (HttpsURLConnection) con;
```
Wadeck: I think the `final` does not add value here

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -407,8 +440,28 @@ static URLConnection openURLConnection(URL url, String credentials, String proxy
...
+            if (insecure) {
+                System.out.println(String.format("Insecure Status: %s", insecure));
```
Wadeck: The use of LOGGER.fine / info should be preferred no ?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: It prevents somebody from modifying a field in the logic just in case

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: I think so, but the file is full of stdout/stderr. @MattLud has followed the current implementation. I think it makes sense to stick to it for a while, but I will create a follow-up ticket to rework it

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
oleg-nenashev: see above

src/main/java/hudson/remoting/jnlp/Main.java
oleg-nenashev: Let's do refactoring as a follow-up

oleg-nenashev:
@reviewbybees done, I'd guess.
@Wadeck let me know if you are not fine with my response.
@rysteboe maybe you want to take a look as well

oleg-nenashev:
@reviewbybees done. Will resolve the merge conflict and release

oleg-nenashev:
I decided to exclude it from the 3.14 release, because I want to do more evaluation + maybe add documentation for this flag.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -345,6 +351,10 @@ public void run() throws Exception {
                     jnlpArgs.add(c);
                 }
             }
+            if (noCertificateCheck) {
+                //TODO: Rename option?
```
oleg-nenashev: done

oleg-nenashev:
I tested it manually on the weekend. LGTM, will merge once the CI passes

oleg-nenashev:
Merging without squash, no plans to backport it

PR #211, opened 2017-11-08
[JENKINS-47901] - Stop using unchecked File#toPath()

oleg-nenashev:
It's a follow-up to #109 

https://issues.jenkins-ci.org/browse/JENKINS-47901

@reviewbybees @rysteboe . Also @stephenc since he was handling the same fun in the core


oleg-nenashev:
@reviewbybees done

PR #214, opened 2017-11-09
Performance: Reduce scope of Channel instance locks by property management

oleg-nenashev:
I was investigating the Channel synchronization logic and discovered that `Channel#setProperty()` and `Channel#getProperty` are inefficient in terms of the lock management. `Channel#properties` is a ConcurrentHashMap, and it does not need synchronization on its own.  The issue may be somehow contributing to the Channel instance deadlocks, but I see no strong evidence of that.

The only reason for synchronization is a wait()/nptifyAll() notification chain for `waitForProperty()`. I believe it needs to be reworked, but for now I just reduced the synchronization scope to these two calls.

It also cleans up the FindBugs issue.

@reviewbybees @rysteboe

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1429,9 +1429,6 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * @return The property or {@code null} if there is no property for the specified key
      */
     @Override
-    @SuppressFBWarnings(value = "UG_SYNC_SET_UNSYNC_GET", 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
PR builder is still unstable. After my investigation of Exceutor service implementations, I suspect that test hanging is caused by a race condition in request executors. To be investigated

oleg-nenashev:
@reviewbybees done. Merging towards the next release

oleg-nenashev:
Created placeholder for this issue: https://issues.jenkins-ci.org/browse/JENKINS-47942

PR #216, opened 2017-11-09
Cleanup usages of the custom Charsets class.

oleg-nenashev:
Remoting is on Java 8, so we can start using StandardCharsets from Java 7

@reviewbybees


src/main/java/org/jenkinsci/remoting/protocol/ProtocolLayer.java
```diff
@@ -45,8 +47,12 @@
     ByteBuffer EMPTY_BUFFER = ByteBufferUtils.EMPTY_BUFFER;
     /**
      * A handy constant until Java 7 compatibility can be assumed.
+     *
+     * @deprecated Usev {@link StandardCharsets}
```
rysteboe: Usev -> Use

PR #220, opened 2017-11-13
[JENKINS-38696] - huson.remoting.UtilTest#mkdirs() should not try creating global paths on Windows.

oleg-nenashev:
On Windows you CAN create such directories.
But it does not mean we should try it even in tests.

Ideally the test needs to be rewritten or removed at all.

@reviewbybees @jtnord


oleg-nenashev:
> Shouldn't this try to delete if it does succeed before failing?

It won't, but I can do it for sure

oleg-nenashev:
PR builder in CI is unstable again. It appears to be an INFRA issues, was unable to reproduce it even with custom Docker container

PR #221, opened 2017-11-13
[JENKINS-47965, JENKINS-38696] - IOHub should not wait infinitely for Selector in Windows.

oleg-nenashev:
Windows implementation of NIO Channel Selector does not request "Select now" in https://github.com/jenkinsci/remoting/blob/6165d6fb6a71a7f4fce52ce5fc4cac479052ce04/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L436 and hence calls selector#select()... and this method has no timeout.

When the code gets into this branch, Selector will be waiting infinitely without calling selector#isOpen(). Such implementation relies on the Selector implementation, which shout interrupt the select() wait if the selector is being closed. But apparently it does not on my machine (Win 10, amd64, Oracle JDK 8u131)

This change adds wait timeout and makes the IOHub implementation to loop in the logic.

- [x] PoC
- [x] Cleanup concurrency, threading, etc

https://issues.jenkins-ci.org/browse/JENKINS-47965

@reviewbybees @stephenc @jtnord

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -93,7 +93,7 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      */
     private final Selector selector;
     private volatile boolean ioHubRunning = false;
-    private Object winIoHubLockObject = new Object();
+    private final Object winIoHubLockObject = new Object();
 
     /**
      * Our executor.
@@ -516,7 +516,9 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
         } finally {
             selectorThread.setName(oldName);
             ioHubRunning = false;
-            winIoHubLockObject.notify();
+            synchronized (winIoHubLockObject) {
+                winIoHubLockObject.notifyAll();
+            }
         }
     }
 
@@ -540,9 +542,15 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
             LOGGER.log(Level.FINEST, "{0}: Started", watcherName);
             try {
                 watcherThread.setName(watcherName);
-                while (iohub.ioHubRunning) {
+                while (true) {
+                    synchronized (iohub.winIoHubLockObject) {
+                        if (iohub.ioHubRunning) {
+                            iohub.winIoHubLockObject.wait(1000);
+                        } else {
+                            break;
+                        }
+                    }
                     iohub.selector.wakeup();
-                    iohub.winIoHubLockObject.wait(1000);
                 }
             } catch (InterruptedException ex) {
                 // interrupted
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -433,7 +433,10 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
                         // in an immediately ready selection key, hence we use the non-blocking form
                         selected = selector.selectNow();
                     } else {
-                        selected = selector.select(); // WINDOWS!!! Waiting the whole timeout anyway if timeout specified?!?
+                        // We just wait for one second if the data is not available
+                        // Before Remoting 3.15 it was an infinite wait, which was causing hanging of the test logic
+                        // If the timeout happens, we will get 0 in selected, and all statistics calculation below will be skipped.
+                        selected = selector.select(1000);
                     }
                     if (selected == 0) {
                         // don't stress the GC by creating instantiating the selected keys
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -92,9 +92,7 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      * Our selector.
      */
     private final Selector selector;
-    private volatile boolean winWatcherSelectorIsWaiting = false;
     private volatile boolean ioHubRunning = false;
-    private long winWatcherSelectorPollTimestamp = 0;
 
     /**
      * Our executor.
@@ -456,10 +454,7 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
                         // On Windows the select(timeout) operation ALWAYS waits for the timeout,
                         // so we workaround it by WindowsIOHubSelectorWatcher
                         if (Launcher.isWindows()) {
-                            winWatcherSelectorIsWaiting = true;
-                            winWatcherSelectorPollTimestamp = System.currentTimeMillis();
                             selected = selector.select();
-                            winWatcherSelectorIsWaiting = false;
                         } else {
                             selected = selector.select(SELECTOR_WAKEUP_TIMEOUT_MS);
                         }
@@ -513,8 +508,6 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
                         cpuOverheatProtection = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(100);
                         Thread.yield();
                     }
-                } finally {
-                    winWatcherSelectorIsWaiting = false;
                 }
             }
         } catch (ClosedSelectorException e) {
@@ -542,23 +535,22 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
             final Thread watcherThread = Thread.currentThread();
             final String oldName = watcherThread.getName();
             final String watcherName = "Windows IOHub Watcher for " + iohub.getThreadNameBase(oldName);
+            LOGGER.log(Level.FINEST, "{0}: Started", watcherName);
             try {
                 watcherThread.setName(watcherName);
                 while (iohub.ioHubRunning) {
-                    if (iohub.winWatcherSelectorIsWaiting && (System.currentTimeMillis() - iohub.winWatcherSelectorPollTimestamp) > SELECTOR_WAKEUP_TIMEOUT_MS) {
-                        LOGGER.log(Level.FINE, "{0}: IOHub Selector is waiting for more that 1 second without events. Waking up the IOHub Selector", watcherName);
-                        iohub.selector.wakeup();
-                    }
                     Thread.sleep(1000);
+                    iohub.selector.wakeup();
                 }
             } catch (InterruptedException ex) {
                 // interrupted
-                if (LOGGER.isLoggable(Level.FINEST)) {
-                    LOGGER.log(Level.FINEST, "{0}: Interrupted", ex);
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    LOGGER.log(Level.FINE, "{0}: Interrupted", ex);
                 }
                 return;
             } finally {
                 watcherThread.setName(oldName);
+                LOGGER.log(Level.FINEST, "{0}: Finished", watcherName);
             }
         }
     }
--- a/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
@@ -92,9 +92,9 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
      * Our selector.
      */
     private final Selector selector;
-    private boolean selectorIsWaiting = false;
-    private boolean ioHubRunning = false;
-    private long selectorPollTimestamp = 0;
+    private volatile boolean winWatcherSelectorIsWaiting = false;
+    private volatile boolean ioHubRunning = false;
+    private long winWatcherSelectorPollTimestamp = 0;
 
     /**
      * Our executor.
@@ -155,6 +155,7 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
         IOHub result = new IOHub(executor);
         executor.execute(result);
         if (Launcher.isWindows()) {
+            LOGGER.log(Level.FINE, "IOHub runs on Windows. There are known issues with Selector#select() hanging, running an additional watcher thread. See JENKINS-47965 for more info");
             executor.execute(new WindowsIOHubSelectorWatcher(result));
         }
         return result;
@@ -425,6 +426,10 @@ public class IOHub implements Executor, Closeable, Runnable, ByteBufferPool {
         selectionKey.attach(null);
     }
 
+    private String getThreadNameBase(String executorThreadName) {
+        return "IOHub#" + _id + ": Selector[keys:" + selector.keys().size() + ", gen:" + gen + "] / " + executorThreadName;
+    }
+
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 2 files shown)


jtnord:
So if I recall when you specify a timeout in windows it waits the entire timeout even if there is something to do.

So basically this will kill performance on windows as all reads will stall for up to the wait after they arrive...

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -433,7 +433,10 @@ public final void run() {
                         // in an immediately ready selection key, hence we use the non-blocking form
                         selected = selector.selectNow();
                     } else {
-                        selected = selector.select(); // WINDOWS!!! Waiting the whole timeout anyway if timeout specified?!?
```
jtnord: this was by design.

if you wait for 1 second then on windows if something happens at t=0 the selector does not return (it should return immedatley something is available),  it will return all the keys that are selecteded at T=1s

oleg-nenashev:
@jtnord 

> So if I recall when you specify a timeout in windows it waits the entire timeout even if there is something to do.

It is, the defect is not fixed in 131 from what I see. I am working on an alternative implementation. Even if it is fixed in newer Java 8 versions, it would not help with remoting use-cases much

jtnord:
@oleg-nenashev 

> Even if it is fixed in newer Java 8 versions, it would not help with remoting use-cases much

Jenkins is JDK8 for a while now and jenkins is pretty much the only consumer of remoting AFAICT, so if we can say the bug can be fixed by using JDK 8 > update xyz then that should be ok with us.

oleg-nenashev:
@jtnord added the lame implementation of the watcher. Even the gir commit hash start from "bad...", so you know...

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -433,11 +450,18 @@ public final void run() {
                         // in an immediately ready selection key, hence we use the non-blocking form
                         selected = selector.selectNow();
                     } else {
-                        // We just wait for one second if the data is not available
-                        // Before Remoting 3.15 it was an infinite wait, which was causing hanging of the test logic
-                        // If the timeout happens, we will get 0 in selected, and all statistics calculation below will be skipped.
-                        selected = selector.select(1000);
```
jtnord: so another alternative which seems to me to be slightly better is to just use 
`selected = selector.select();`

and then create a thread that every one second does selector.wakeup() because we already have to handle spurious wakeups and then you avoid keeping lots of state with the timeout and `selectorPollTimestamp` not being thread safe because its not atomic being 64bit and not volatile...

oleg-nenashev: As I said, the impl is lame :) 
Just wanted to dry run the concept

jtnord:
just wondering but should we add a `selector.wakeup()` after [this line](https://github.com/oleg-nenashev/remoting/blob/bad5040625b22c360781d31f6a36df3f0dcff70b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L263)

oleg-nenashev:
> just wondering but should we add 

probably, didn't help me in tests

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -433,8 +450,18 @@ public final void run() {
                         // in an immediately ready selection key, hence we use the non-blocking form
                         selected = selector.selectNow();
                     } else {
-                        selected = selector.select(); // WINDOWS!!! Waiting the whole timeout anyway if timeout specified?!?
+                        // On Windows the select(timeout) operation ALWAYS waits for the timeout,
+                        // so we workaround it by WindowsIOHubSelectorWatcher
+                        if (Launcher.isWindows()) {
+                            selectorIsWaiting = true;
```
stephenc: none of these are volatile and there is no synchronization, the other thread will not be guaranteed to see them

stephenc:
I think @jtnord is on to a winner with https://github.com/oleg-nenashev/remoting/blob/bad5040625b22c360781d31f6a36df3f0dcff70b/src/main/java/org/jenkinsci/remoting/protocol/IOHub.java#L263

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
oleg-nenashev: Added the task to the PR to make it explicit

oleg-nenashev:
@stephenc @jtnord Before writing this PoC I have tried to add wakeup() to close(). It didn't work. Here is the default `close()` logic in SelectorImpl:

```
public void implCloseSelector() throws IOException {
    this.wakeup();
    synchronized(this) {
        Set var2 = this.publicKeys;
        synchronized(this.publicKeys) {
            Set var3 = this.publicSelectedKeys;
            synchronized(this.publicSelectedKeys) {
                this.implClose();
            }
        }
    }
}
```

It starts from the wakeup(), so no wonder that adding additional wakeup() after close does not help.

stephenc:
IIRC the windows selector is implemented differently

jtnord:
http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/d5a00b1e8f78/src/windows/classes/sun/nio/ch/WindowsSelectorImpl.java

but then the selector is doing a `wakeup` before the `close` not after - so this seems like an opportunity for a race condition... which is why @stephenc and I where seeing differences when run in a VM with one core...

oleg-nenashev:
@jtnord @stephenc  So adding `wakeup()` call after `close()` does not work for me.
Does it make sense to polish the current PoC in the current state?

oleg-nenashev:
@jtnord @stephenc ping

stephenc:
@oleg-nenashev entirely up to you. Cross check performance before and after on windows and *nix before merging... if the perf drops significantly for windows then that would be a "do not merge" argument... if you can find something that doesn't have the perf hit and works for windows consistently, then go for it

stephenc:
Keep in mind, we have https://github.com/jenkinsci/remoting/tree/master/src/assembly/agent-load-test to allow for these kind of measurements

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
oleg-nenashev: fixed

oleg-nenashev:
Tests on Windows take longer due to the hack, but the PR builder works well. I will submit load test results soon

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -83,6 +92,10 @@
      * Our selector.
      */
     private final Selector selector;
+    private volatile boolean winWatcherSelectorIsWaiting = false;
+    private volatile boolean ioHubRunning = false;
+    private long winWatcherSelectorPollTimestamp = 0;
```
jtnord: you can make this much simpler by removing this and just doing an unconditional wakeup() every second.
the worst that happens if you have 2 wakeups in an exceptionally short time (one for data then immediatly a spurious wakeup) - but that seems reasonable.

oleg-nenashev: OK, will do. It does not cause performance degradation, but I agree it is a bit overcomplicated

oleg-nenashev:
@stephenc I ran load tests on Windows/Mac. All performance stats are more or less same before and after changes, so I think it's ready for review. CC @jtnord

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -83,6 +92,10 @@
      * Our selector.
      */
     private final Selector selector;
+    private volatile boolean winWatcherSelectorIsWaiting = false;
```
stephenc: You can also make this even simpler still... don't even use this boolean, just all `iohub.selector.wakeup()` every `SELECTOR_WAKEUP_TIMEOUT_MS` as long as `ioHubRunning` that way the thread becomes ultra simple

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -83,6 +92,10 @@
...
+    private volatile boolean ioHubRunning = false;
```
stephenc: Use a lock object so that the wakeup thread can use `ioHubRunningLock.wait(SELECTOR_WAKEUP_TIMEOUT_MS)` rather than a volatile... otherwise you will have to wait for the remainder of the sleep to shut down the executor

oleg-nenashev:
@stephenc you were reviewing the old code snapshot, right?

oleg-nenashev:
> you were reviewing the old code snapshot, right?

nvm, I just forgot to cleanup some code

oleg-nenashev:
ah ok, just forgot to commit what I had on Windows machine

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -489,6 +514,44 @@ public final void run() {
...
+        @Override
+        public void run() {
+            final Thread watcherThread = Thread.currentThread();
+            final String oldName = watcherThread.getName();
+            final String watcherName = "Windows IOHub Watcher for " + iohub.getThreadNameBase(oldName);
+            LOGGER.log(Level.FINEST, "{0}: Started", watcherName);
+            try {
+                watcherThread.setName(watcherName);
+                while (iohub.ioHubRunning) {
+                    Thread.sleep(1000);
```
stephenc: :bug: causes the executor service to excessively wait on shutdown. Better to use a lock object for running and use `wait/notifyAll`

oleg-nenashev:
@stephenc Got rid of `Thread.sleep()`

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -489,6 +515,53 @@ public final void run() {
...
+            final Thread watcherThread = Thread.currentThread();
+            final String oldName = watcherThread.getName();
+            final String watcherName = "Windows IOHub Watcher for " + iohub.getThreadNameBase(oldName);
+            LOGGER.log(Level.FINEST, "{0}: Started", watcherName);
+            try {
+                watcherThread.setName(watcherName);
+                while (true) {
+                    synchronized (iohub.winIoHubLockObject) {
+                        if (iohub.ioHubRunning) {
+                            iohub.winIoHubLockObject.wait(1000);
```
jtnord: should this not use `SELECTOR_WAKEUP_TIMEOUT_MS` rather than be hard coded?

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -83,6 +92,9 @@
      * Our selector.
      */
     private final Selector selector;
+    private volatile boolean ioHubRunning = false;
```
jtnord: why introduce this new volatile?  Can't you just check `selector.isOpen()`?

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -489,6 +515,53 @@ public final void run() {
...
+        public void run() {
+            final Thread watcherThread = Thread.currentThread();
+            final String oldName = watcherThread.getName();
+            final String watcherName = "Windows IOHub Watcher for " + iohub.getThreadNameBase(oldName);
+            LOGGER.log(Level.FINEST, "{0}: Started", watcherName);
+            try {
+                watcherThread.setName(watcherName);
+                while (true) {
+                    synchronized (iohub.winIoHubLockObject) {
+                        if (iohub.ioHubRunning) {
```
jtnord: Why not use `iohub.selector.isOpen()` ?

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -433,8 +452,15 @@ public final void run() {
                         // in an immediately ready selection key, hence we use the non-blocking form
                         selected = selector.selectNow();
                     } else {
-                        selected = selector.select(); // WINDOWS!!! Waiting the whole timeout anyway if timeout specified?!?
+                        // On Windows the select(timeout) operation ALWAYS waits for the timeout,
+                        // so we workaround it by WindowsIOHubSelectorWatcher
```
jtnord: as part of this can you file a JDK bug with a reproducable scenario - just to make sure we are not doing something slightly wrong...

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -489,6 +515,53 @@ public final void run() {
             // ignore, happens routinely
         } finally {
             selectorThread.setName(oldName);
+            ioHubRunning = false;
```
jtnord: at this point `selector.isOpen()` should return `false` - so can you get rid of this?

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -489,6 +515,53 @@ public final void run() {
...
+                watcherThread.setName(watcherName);
+                while (true) {
+                    synchronized (iohub.winIoHubLockObject) {
+                        if (iohub.ioHubRunning) {
+                            iohub.winIoHubLockObject.wait(1000);
+                        } else {
+                            break;
+                        }
+                    }
+                    iohub.selector.wakeup();
```
jtnord: // wakeup is one of the "safe to call" methods after a selector has been closed

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
```diff
@@ -140,6 +153,10 @@ private IOHub(Executor executor) throws IOException {
     public static IOHub create(Executor executor) throws IOException {
         IOHub result = new IOHub(executor);
         executor.execute(result);
+        if (Launcher.isWindows()) {
```
jtnord: BTW - I think this is no longer correct.

The network stack in my experience behaves the same when run on the Linux inside windows (ubuntu on bash on windows, for one example) - so checking for the file separator character is no longer correct.

there are some things that you want to do because you are on windows like paths, using different programes,  yet there are some things that you want to do due to differences in a protocol stack (like here)

jtnord: SO having run the tests on "bash on Ubuntu on Windows" org.jenkinsci.remoting.protocol.impl.NetworkLayerTest is hanging.
alas I can not get a stack trace from surefire when using this environment :(

src/main/java/org/jenkinsci/remoting/protocol/IOHub.java
oleg-nenashev: Only if it is aborted properly. I see hanging of the selector close() operation in some unit tests

oleg-nenashev:
@jtnord @stephenc another review iteration would be appreciated

oleg-nenashev:
Works on Windows, works on Ubuntu, works on Mac.
I have some problems with local unit tests on Mac, but it does not come from this PR

PR #222, opened 2017-11-13
[JENKINS-38696] - Do not delete temporary files in Windows in PrefetchingTest.

oleg-nenashev:
It's a bit weird, but the test uses real classloader, which keeps the files open.
Ideally the tests needs to be fixed in order to stop keeping handlers in JVM when loading JARs (I started doing it for Remoting test client in RTC 1.1), but it's not the top priority right now.

@reviewbybees @jtnord

### DIFF
--- a/src/test/java/hudson/remoting/PrefetchingTest.java
+++ b/src/test/java/hudson/remoting/PrefetchingTest.java
@@ -30,7 +30,7 @@ public class PrefetchingTest extends RmiTestBase implements Serializable {
     private Checksum sum1,sum2;
 
     @Override
-    protected void setUp() throws Exception {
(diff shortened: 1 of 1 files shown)


src/test/java/hudson/remoting/PrefetchingTest.java
```diff
@@ -30,7 +30,7 @@
     private Checksum sum1,sum2;
 
     @Override
-    protected void setUp() throws Exception {
+    protected void setUp() throws Exception {
```
rysteboe: Nit:  extra whitespace

rysteboe:
:bee:

oleg-nenashev:
Merging, will rework the tests later

PR #226, opened 2017-11-15
Tests: Avoid using reflection in HandlerLoopbackLoadStress for CPU stats when possible.

oleg-nenashev:
This is not a part of the production code, but I use it sometimes. Reflection is not really required when you run with Oracle JDK / OpenJDK.

In longer term I am going to refactor the Remoting repository structure, so this code will be more explicit and documented in Developer guides.

@reviewbybees @stephenc @rysteboe


oleg-nenashev:
@reviewbybees done

PR #227, opened 2017-11-15
Explicitly deprecate and Restrict JNLP3 utility classes.

oleg-nenashev:
The protocol is deprecated, I would like to explicitly advice against using its classes internally and externally.

@reviewbybees @rysteboe @akshayabd


oleg-nenashev:
@reviewbybees done

PR #228, opened 2017-11-16
[JENKINS-48055] - Introduce convenient Channel retrieval methods in Callables.

oleg-nenashev:
This change introduces new API, which is expected to cleanup FindBugs reports in the core and to offer better diagnostics of edge cases.
The logic also allows failing fast when the channel is being terminated.

Downstream change: https://github.com/jenkinsci/jenkins/pull/3145

https://issues.jenkins-ci.org/browse/JENKINS-48055

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/Callable.java
+++ b/src/main/java/hudson/remoting/Callable.java
@@ -23,11 +23,18 @@
  */
 package hudson.remoting;
 
+import org.jenkinsci.remoting.ChannelStateException;
 import org.jenkinsci.remoting.RoleSensitive;
+import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
 
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.NotSerializableException;
 import java.io.Serializable;
 
 //TODO: Make it SerializableOnlyOverRemoting?
+// Oleg Nenashev: Formally there is no reason for that, you can serialize object over whatever binary stream and then
+// execute it on remote side if it has channel instance. Likely YAGNI, but I can imagine such Pipeline context class implementation.
 /**
  * Represents computation to be done on a remote system.
  *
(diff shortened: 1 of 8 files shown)


src/main/java/hudson/remoting/forward/PortForwarder.java
```diff
@@ -134,9 +134,10 @@ public static ListeningPort create(VirtualChannel ch, final int acceptingPort, F
 
         return ch.call(new Callable<ListeningPort,IOException>() {
             public ListeningPort call() throws IOException {
+                final Channel channel = getOpenChannelOrFail(); // We initialize it early, so the forwarder won's start its daemon if the channel is shutting down
```
oleg-nenashev: There is no sense in starting daemon when we cannot return the port back, right?

rysteboe: I agree.

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
@rysteboe could you please re-review after the change?

src/main/java/hudson/remoting/Callable.java
```diff
@@ -23,11 +23,18 @@
...
+import org.jenkinsci.remoting.SerializableOnlyOverRemoting;
 
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.NotSerializableException;
 import java.io.Serializable;
 
 //TODO: Make it SerializableOnlyOverRemoting?
+// Oleg Nenashev: Formally there is no reason for that, you can serialize object over whatever binary stream and then
+// execute it on remote side if it has channel instance. Likely YAGNI, but I can imagine such Pipeline context class implementation.
```
jglick: Pipeline would do no such thing.

src/main/java/hudson/remoting/Callable.java
```diff
@@ -40,4 +47,46 @@
...
+    }
+
+    /**
+     * Gets an open channel, which is ready to accept commands.
+     *
+     * It is a convenience method for cases, when a callable needs to invoke call backs on the master.
+     * In such case the requests will be likely failed by {@linkplain UserRequest} logic anyway, but it is better to fail fast.
+     * 
+     * @return Channel instance
+     * @throws ChannelStateException The channel is closing down or has been closed.
```
jglick: Could specify `ChannelClosedException` for this case.

oleg-nenashev: yup

PR #229, opened 2017-11-16
[JENKINS-37566] - FindBugs: Suppress serialization warning for UserRequest#classLoaderProxy.

oleg-nenashev:
The field is actually always serializable, but we cannot confirm it statically due to the reflection code.
I decided to just suppress it for now and to document the method.

@reviewbybees @rysteboe


oleg-nenashev:
Test failure is unrelated

PR #230, opened 2017-11-17
Javadoc: Update documentation for JNLP4 components and layers.

oleg-nenashev:
Just some minor polishing while I was around

@reviewbybees @rysteboe @stephenc


oleg-nenashev:
@stephenc thnks!

PR #231, opened 2017-11-20
[JENKINS-37566] - Suppress issue with non-serializable inner classes in NioChannelHub

oleg-nenashev:
The wrapper may decrease the performance a bit, but NioChannelHub is not being used in JNLP4 anyway. Ideally submissions should not use the selector tasks queue in such lame way, but it’s risky to change it to another executor service.

@reviewbybees @rysteboe

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
+++ b/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
@@ -535,6 +535,10 @@ public class NioChannelHub implements Runnable, Closeable {
         private Object readResolve() throws ObjectStreamException {
             throw new NotSerializableException("The class should not be serialized over Remoting");
         }
+        
+        private Object writeReplace() throws ObjectStreamException {
+            throw new NotSerializableException("The class should not be serialized over Remoting");
+        }
 
         @Override
         public void checkRoles(RoleChecker checker) throws SecurityException {
--- a/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
+++ b/src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
@@ -19,6 +19,8 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
+import java.io.NotSerializableException;
+import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.nio.channels.CancelledKeyException;
 import java.nio.channels.ClosedSelectorException;
@@ -39,6 +41,7 @@ import java.util.logging.Logger;
 import static java.nio.channels.SelectionKey.*;
 import static java.util.logging.Level.*;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 
 /**
  * Switch board of multiple {@link Channel}s through NIO select.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
```diff
@@ -508,11 +497,52 @@ protected CommandTransport makeTransport(InputStream is, OutputStream os, Mode m
...
+        }
+
+        @Override
+        public Void call() throws IOException {
+            if (task == null) {
+                throw new IOException("The callable " + this + " has been serialized somehow, but it is actually not serializable");
+            }
+            try {
+                return task.call();
+            } catch (Exception ex) {
```
stephenc: simpler:

```
} catch (IOException e) {
  throw e;
} catch (Exception e) {
  throw new IOException(e);
}
```

src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
```diff
@@ -508,11 +497,52 @@ protected CommandTransport makeTransport(InputStream is, OutputStream os, Mode m
...
+        private static final long serialVersionUID = -7331104479109353930L;
+        final transient java.util.concurrent.Callable<Void> task;
+
+        CallableRemotingWrapper(@Nonnull java.util.concurrent.Callable<Void> task) {
+            this.task = task;
+        }
+
+        @Override
+        public Void call() throws IOException {
+            if (task == null) {
```
stephenc: I personally thing you'd be much better with `assert task != null : "Serialization banned"`

src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
```diff
@@ -508,11 +497,52 @@ protected CommandTransport makeTransport(InputStream is, OutputStream os, Mode m
...
+                    throw (IOException)ex;
+                }
+                throw new IOException(ex);
+            }
+        }
+
+        private Object readResolve() throws ObjectStreamException {
+            throw new NotSerializableException("The class should not be serialized over Remoting");
+        }
+
```
stephenc: consider 

```
private Object writeReplace() throws ObjectStreamException {
  throw new NotSerializableException("The class should not be serialized over Remoting");
}
```

stephenc:
This crazy was part of the reason we had to move to IOHub in order to use SSLEngine

src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
oleg-nenashev: yep, makes sense

src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
rysteboe: +1

rysteboe: For github UI stuff

oleg-nenashev:
@rysteboe ack, will do

src/main/java/org/jenkinsci/remoting/nio/NioChannelHub.java
oleg-nenashev: I prefer to fail the production code as well. Just in case...

oleg-nenashev:
Addressed comments

PR #232, opened 2017-11-20
[JENKINS-37566] - Stop nullifying channel in ProxyOutputStream#finalize()

oleg-nenashev:
Just YAGNI, it does not help garbage collector much and actually makes it more complicated due to breaking the object chain.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/ProxyOutputStream.java
+++ b/src/main/java/hudson/remoting/ProxyOutputStream.java
@@ -181,7 +181,7 @@ final class ProxyOutputStream extends OutputStream implements ErrorPropagatingOu
         super.finalize();
         // if we haven't done so, release the exported object on the remote side.
         // if the object is auto-unexported, the export entry could have already been removed.
-        if(channel!=null) {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/ProxyOutputStream.java
```diff
@@ -183,7 +183,6 @@ protected void finalize() throws Throwable {
         // if the object is auto-unexported, the export entry could have already been removed.
         if(channel!=null) {
```
stephenc: `&& oid != -1`

oleg-nenashev: yep though there is another issue. Not sure I want to touch it

stephenc: Well the null assignment was preventing a double export...

oleg-nenashev: double unexport is actually fine (excepting logging). But OK, will add a check

PR #234, opened 2017-11-21
getBytes(Charset) don’t throw exception

ndeloof:


oleg-nenashev:
I would keep "UTF-8" in the PREAMBLE. It does not matter much in the case of this case, but I would prefer to keep UTF-8 strict so that it's aligned with encoding in the handshake logic: https://github.com/jenkinsci/remoting/blob/b7c0861b556363e933f7495a0d5730e871c05731/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol2Handler.java#L106

oleg-nenashev:
@ndeloof When one does force pushes, GitHub does not notify about new commits :(

PR #237, opened 2017-11-21
[JENKINS-37566] - Cleanup issues in Command#createdAt handling

oleg-nenashev:
I noticed that `Command#createdAt` is being handled incorrectly in some cases, so I annotated the related methods and fixed FindBugs issues.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/remoting/PipeWindow.java
+++ b/src/main/java/hudson/remoting/PipeWindow.java
@@ -24,6 +24,9 @@
 package hudson.remoting;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/remoting/PipeWindow.java
```diff
@@ -95,19 +104,23 @@
 
     /**
      * Indicates that the remote end has died and all the further send attempt should fail.
+     * @param cause Death cause. If {@code null}, the death will be still considered as dead, but there will be no cause recorded..
      */
-    void dead(Throwable cause) {
-        this.dead = cause;
+    void dead(@CheckForNull Throwable cause) {
+        // We need to record
+        this.dead = cause != null ? cause : new RemotingSystemException("Unknown cause", null) ;
```
oleg-nenashev: Without it the handling was just wrong in the case of the missing cause. It may be a root cause of some defects reported to remoting

oleg-nenashev:
Thanks @rysteboe!

PR #238, opened 2017-11-28
[JENKINS-37670] - RemoteClassLoader should throw UnsupportedClassVersionError when the bytecode is not supported

oleg-nenashev:
The case is well described by Thomas Maslen in https://issues.jenkins-ci.org/browse/JENKINS-37670. `UnsupportedClassVersionError` inherits `ClassFormatError`, so there should be no issues in any downstream code.

@reviewbybees @rysteboe


oleg-nenashev:
CC @MarkEWaite, who may have related reports in Git plugins due to the handling in JGit

PR #239, opened 2017-11-30
[JENKINS-37566] - FindBugs: Prevent the unconditional wait warning in Channel#waitForProperty()

oleg-nenashev:
Just rebalances the code a bit, which may helpful if waitForProperty() waits for the channel lock too long and it gets closed.

@reviewbybees @rysteboe


oleg-nenashev:
Yes, in this case it is just yak-shaving. It does not make the code worse, so why not?

oleg-nenashev:
@reviewbybees done

PR #240, opened 2017-11-30
[JENKINS-48309] - Prevent TimeoutException in AsyncFutureImpl in the case of spurious wakeups

oleg-nenashev:
**TL;DR**: A large part of Jenkins’ get-with-timeout logic is a subject for improper exit before the timeout.

It happens, because "wait(timeout)" is not in the loop. Object#wait(long) is explicit about that: "A thread can also wake up without being notified, interrupted, or timing out, a so-called spurious wakeup. While this will rarely occur in practice, applications must guard against it by testing for the condition that should have caused the thread to be awakened, and continuing to wait if the condition is not satisfied. In other words, waits should always occur in loops..."

https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait-long-

Impact:

* AsyncFutureImpl is being used explicitly in JarCacheSupport. Jar resolution with a timeout may fail due to the issue
* The class is overridden in Jenkins Core's public API, e.g. hudson.model.queue.FutureImpl. It's hard to estimate risks in the core, but it looks bad.

https://issues.jenkins-ci.org/browse/JENKINS-48309

@reviewbybees @rysteboe @jglick

### DIFF
--- a/src/main/java/hudson/remoting/AsyncFutureImpl.java
+++ b/src/main/java/hudson/remoting/AsyncFutureImpl.java
@@ -89,8 +89,7 @@ public class AsyncFutureImpl<V> implements Future<V> {
                 break;
             }
 
-            long timeToWaitMs = Math.max(1, timeToWait / 1000000);
-            wait(timeToWaitMs);
+            wait(timeToWait / 1000000, (int)(timeToWait % 1000000));
         }
 
         if(!completed)
--- a/src/main/java/hudson/remoting/AsyncFutureImpl.java
+++ b/src/main/java/hudson/remoting/AsyncFutureImpl.java
@@ -81,16 +81,16 @@ public class AsyncFutureImpl<V> implements Future<V> {
 
     @CheckForNull
     public synchronized V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        long timeoutMs = unit.toMillis(timeout);
-        // It is not very accurate, but all(proof?) production uses of this class pass seconds.
-        // Test cases pass 1 millisecond and other such fun, but wait(long) is not time-accurate anyway
-        long endWaitTime = System.currentTimeMillis() + timeoutMs;
+        // The accuracy of wait(long) operation is milliseconds anyway, but ok.
+        long endWaitTime = System.nanoTime() + unit.toNanos(timeout);
         while (!completed) {
-            long currentTime = System.currentTimeMillis();
-            if (currentTime > endWaitTime) {
+            long timeToWait = endWaitTime - System.nanoTime();
+            if (timeout < 0) {
                 break;
             }
-            wait(endWaitTime - currentTime);
+
+            long timeToWaitMs = Math.max(1, timeToWait / 1000000);
+            wait(timeToWaitMs);
         }
 
         if(!completed)
--- a/src/main/java/hudson/remoting/AsyncFutureImpl.java
+++ b/src/main/java/hudson/remoting/AsyncFutureImpl.java
@@ -23,6 +23,7 @@
  */
 package hudson.remoting;
 
+import javax.annotation.CheckForNull;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -78,13 +79,28 @@ public synchronized V get() throws InterruptedException, ExecutionException {
...
     }
 
+    @CheckForNull
     public synchronized V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        if(!completed)
-            wait(unit.toMillis(timeout));
+        long timeoutMs = unit.toMillis(timeout);
+        // It is not very accurate, but all(proof?) production uses of this class pass seconds.
+        // Test cases pass 1 millisecond and other such fun, but wait(long) is not time-accurate anyway
+        long endWaitTime = System.currentTimeMillis() + timeoutMs;
```
stephenc: :bug: will cause massive issues as the system time changes. You need to use nanotime instead... care for comparison logic though

src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -78,13 +79,28 @@ public synchronized V get() throws InterruptedException, ExecutionException {
...
+        }
+
         if(!completed)
             throw new TimeoutException();
         if(cancelled)
             throw new CancellationException();
+
+        //TODO: we may be calling a complex get() implementation in the override, which may hang the code
+        // The only missing behavior is "problem!=null" branch, but probably we cannot just replace the code without
+        // an override issue risk.
```
stephenc: Then perhaps move the get outside the synchronised

oleg-nenashev:
Probably it causes https://issues.jenkins-ci.org/browse/JENKINS-20947, not 100% sure.

src/main/java/hudson/remoting/AsyncFutureImpl.java
oleg-nenashev: I will need to investigate the implementations first. `get()` outside synchronized may help for sure, but I would like to find real examples before touching this code.

src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -78,13 +79,28 @@ public synchronized V get() throws InterruptedException, ExecutionException {
...
+        // The accuracy of wait(long) operation is milliseconds anyway, but ok.
+        long endWaitTime = System.nanoTime() + unit.toNanos(timeout);
+        while (!completed) {
+            long timeToWait = endWaitTime - System.nanoTime();
+            if (timeout < 0) {
+                break;
+            }
+
+            long timeToWaitMs = Math.max(1, timeToWait / 1000000);
+            wait(timeToWaitMs);
```
stephenc: There is `wait(long,int)` to wait with “nanosecond” precision

src/main/java/hudson/remoting/AsyncFutureImpl.java
stephenc: Ack

src/main/java/hudson/remoting/AsyncFutureImpl.java
oleg-nenashev: The implementation in my Java does the same, actually. Decompiled code:

```
public final void wait(long timeout, int nanos) throws InterruptedException {
        if (timeout < 0) {
            throw new IllegalArgumentException("timeout value is negative");
        }

        if (nanos < 0 || nanos > 999999) {
                                "nanosecond timeout value out of range");
        }

        if (nanos >= 500000 || (nanos != 0 && timeout == 0)) {
            timeout++;
        }

        wait(timeout);
    }
```

I will patch the impl though

oleg-nenashev:
ping @stephenc

src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -78,13 +79,27 @@ public synchronized V get() throws InterruptedException, ExecutionException {
...
-            wait(unit.toMillis(timeout));
+        // The accuracy of wait(long) operation is milliseconds anyway, but ok.
+        long endWaitTime = System.nanoTime() + unit.toNanos(timeout);
+        while (!completed) {
+            long timeToWait = endWaitTime - System.nanoTime();
+            if (timeout < 0) {
+                break;
+            }
+
+            wait(timeToWait / 1000000, (int)(timeToWait % 1000000));
```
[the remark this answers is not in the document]
jtnord: underscores in java 8  `1_000_000`

src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -78,13 +79,27 @@ public synchronized V get() throws InterruptedException, ExecutionException {
...
 
+    @CheckForNull
     public synchronized V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        if(!completed)
-            wait(unit.toMillis(timeout));
+        // The accuracy of wait(long) operation is milliseconds anyway, but ok.
+        long endWaitTime = System.nanoTime() + unit.toNanos(timeout);
+        while (!completed) {
+            long timeToWait = endWaitTime - System.nanoTime();
+            if (timeout < 0) {
```
stephenc: `s/timeout/timeToWait/`

oleg-nenashev: oh yes :( It makes me think that some test coverage could be actually useful

oleg-nenashev:
Test hanging is unrelated, it comes from the previous change. Git bisect points to `[7bd389e7991efedf0cd67afb7ce821abb50891bb] Javadoc: cleanup empty tag Javadoc warnings` after 2 attempts, looks promising 🤷‍♂️

PR #243, opened 2017-12-14
[FIXED JENKINS-48561] Give precedence to proxy exclusion list system property over env var

etiennebec:
https://issues.jenkins-ci.org/browse/JENKINS-48561

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -436,6 +436,7 @@ public class JnlpAgentEndpointResolver {
      * Gets URL connection.
      * If http_proxy environment variable exists,  the connection uses the proxy.
      * Credentials can be passed e.g. to support running Jenkins behind a (reverse) proxy requiring authorization
+     * FIXME: similar to hudson.remoting.Util.openURLConnection which is still used in hudson.remoting.Launcher
      */
     static URLConnection openURLConnection(URL url, String credentials, String proxyCredentials,
                                            SSLSocketFactory sslSocketFactory, boolean disableHttpsCertValidation) throws IOException {
@@ -505,6 +506,8 @@ public class JnlpAgentEndpointResolver {
      * - To match IPV4/IPV/FQDN: Regular Expressions Cookbook, 2nd Edition (ISBN: 9781449327453)
      *
      * Warning: this method won't match shortened representation of IPV6 address
+     *
+     * FIXME: duplicate of hudson.remoting.Util.inNoProxyEnvVar
      */
     static boolean inNoProxyEnvVar(String host) {
         String noProxy = System.getenv("no_proxy");
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -397,6 +397,9 @@ public class JnlpAgentEndpointResolver {
                             sj.add(Pattern.quote(entry.substring(0, entry.length() - 1)) + ".*");
                         else
                             sj.add(Pattern.quote(entry));
+                        // Detect when the pattern contains multiple wildcard, which used to work previous to Java 9 (e.g. 127.*.*.*)
+                        if(entry.split("\\*").length > 2)
+                            LOGGER.log(Level.WARNING, "Using more than one wildcard is not supported in nonProxyHosts entries: {0}", entry);
                     }
                     Pattern nonProxyRegexps = Pattern.compile(sj.toString());
                     if(nonProxyRegexps.matcher(host.toLowerCase(Locale.ENGLISH)).matches()) {
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -26,6 +26,8 @@ package org.jenkinsci.remoting.engine;
 import hudson.remoting.Base64;
 import org.jenkinsci.remoting.util.https.NoCheckHostnameVerifier;
 import org.jenkinsci.remoting.util.https.NoCheckTrustManager;
+import sun.misc.RegexpPool;
+import sun.net.NetProperties;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -377,7 +380,30 @@ static InetSocketAddress getResolvedHttpProxyAddress(@Nonnull String host, int p
...
+                // Unfortunately, the Proxy class does not provide a way to differentiate both cases to fallback to
+                // environment variables only when no proxy has been configured. Therefore, we have to recheck if the URI
+                // host is in the exclusion list.
+                String nonProxyHosts = NetProperties.get("http.nonProxyHosts");
+                if(nonProxyHosts != null && nonProxyHosts.length() != 0) {
+                    RegexpPool exclusionsPool = new RegexpPool();
+                    StringTokenizer stringTokenizer = new StringTokenizer(nonProxyHosts, "|", false);
+                    try {
+                        while(stringTokenizer.hasMoreTokens()) {
+                            exclusionsPool.add(stringTokenizer.nextToken().toLowerCase(), Boolean.TRUE);
```
oleg-nenashev: `toLowerCase()` is invoked without locale, it will cause FindBugs failures with #242 I'd guess

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -377,7 +380,30 @@ static InetSocketAddress getResolvedHttpProxyAddress(@Nonnull String host, int p
...
+                        }
+                    } catch(sun.misc.REException e) {
+                        System.err.println("Malformed exception list in http.nonProxyHosts system property: " + e.getMessage());
```
oleg-nenashev: I will need to rework everything to loggers

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -377,7 +380,30 @@ static InetSocketAddress getResolvedHttpProxyAddress(@Nonnull String host, int p
...
+                    }
+                    if(exclusionsPool.match(host.toLowerCase()) != null) {
```
oleg-nenashev: locale

etiennebec:
Thanks for reviewing it. I pushed another two commits to take in account your comments.

On a side note, I noticed that some portions of code have been partially copy/pasted from `hudson.remoting.Util` to `org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver`, mainly `inNoProxyEnvVar()` and `openURLConnection()`: https://github.com/jenkinsci/remoting/commit/2cea38c49820791c2cc1f974973391e50a443d71

`hudson.remoting.Launcher` is still using the `Util.openURLConnection`, but `JnlpAgentEndpointResolver.resolve` uses the other one.

Maybe it is a way to keep compatibility with older releases while transitioning from hudson namespace to org.jenkinsci?

etiennebec:
Noticed in mvn logs a warning regarding usage of Sun proprietary API (`sun.misc.RegexpPool` and `sun.net.NetProperties`). Should I try to replace them?

oleg-nenashev:
> On a side note, I noticed that some portions of code have been partially copy/pasted.... Maybe it is a way to keep compatibility with older releases while transitioning from hudson namespace to org.jenkinsci? IDK.

In this case it would be possible to have an utility class somewhere and to invoke its methods from other places. No need to invoke special magic, just keeping and probably deprecating the current methods, but moving the duplicated logic to another method

oleg-nenashev:
@etiennebec Are you still planning to change the classes?

etiennebec:
@oleg-nenashev yes, probably by the end of next week.

oleg-nenashev:
Jenkins project does not support Java 9 so far. Is it possible to have a Java8-compatible implementation without deprecated classes?

etiennebec:
OpenJDK9 removed references to old internal Sun lib, including the ones used in `DefaultProxySelector`:
- https://bugs.java.com/view_bug.do?bug_id=8035158
- http://hg.openjdk.java.net/jdk9/jdk9/jdk/rev/50a749f2cade

So I re-used the same logic which results in a highly similar code. I have two concerns:
- [diff-b72206907388927fa66809bdb849f86eR391](https://github.com/jenkinsci/remoting/pull/243/files#diff-b72206907388927fa66809bdb849f86eR391) is almost a copy paste of [50a749f2cade#l1.91](http://hg.openjdk.java.net/jdk9/jdk9/jdk/rev/50a749f2cade#l1.91) but I'm not able to find another, simpler, implementation. Is this OK?
- In the past, I've seen people using multiple wildcards in -DnonProxyHosts entries (e.g. `127.*.*.*`). It was working since the entries were interpreted as regex, even though it was not explicitly documented: https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html. This no longer work with the new implementation in Java 9 which only supports a wildcard at the beginning or the end of the hostname. I added a warning message which should be pretty obvious in case anyone get bite by this new behavior.

I will probably refactor and simplify a bit other proxy related methods, including the duplicate of `inNoProxyEnvVar`, but this will be the subject of another PR.

etiennebec:
Didn't expect you to review my new commits so quickly. :+1: 

> Jenkins project does not support Java 9 so far. Is it possible to have a Java8-compatible implementation without deprecated classes?

The code is actually Java8-compatible (I don't even have the JDK9 installed on my computer).

When I say in the comment that the implementation supports only Java9, I'm referring to the previous syntax of nonProxyHosts.

`DefaultProxySelector` implementation was more permissive than what was described in the official documentation (as stated in my previous message), but Java9 restored the expected and correct behavior.

oleg-nenashev:
Sorry, slipped down in my priority list

oleg-nenashev:
I will try to play with that in order to deliver it in Remoting 3.18 (which will be likely an LTS candidate). Sorry for the review delay, JEP-200 consumed much my capacity

etiennebec:
Thanks for all your effort on JEP-200. 👍 

Don't worry, we're compiling our own version of jenkins/remoting/plugins in my company, so it's not an issue if it takes some time for this PR to be mainlined. Just trying to give back to the community!

oleg-nenashev:
CC @rysteboe . I think we can land it

PR #244, opened 2017-12-19
[JENKINS-48686] Switch to calling the agent 'agent'

jsoref:
https://issues.jenkins-ci.org/browse/JENKINS-48686

### DIFF
--- a/src/main/java/hudson/remoting/InitializeJarCacheMain.java
+++ b/src/main/java/hudson/remoting/InitializeJarCacheMain.java
@@ -34,7 +34,7 @@ public class InitializeJarCacheMain {
     public static void main(String[] argv) throws Exception {
         if (argv.length != 2) {
             throw new IllegalArgumentException(
-                    "Usage: java -cp slave.jar hudson.remoting.InitializeJarCacheMain " +
+                    "Usage: java -cp agent.jar hudson.remoting.InitializeJarCacheMain " +
                     "<source jar dir> <jar cache dir>");
         }
 
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -128,7 +128,7 @@ public class Launcher {
     public File slaveLog = null;
 
     @Option(name="-text",usage="encode communication with the master with base64. " +
-            "Useful for running slave over 8-bit unsafe protocol like telnet")
+            "Useful for running agent over 8-bit unsafe protocol like telnet")
     public void setTextMode(boolean b) {
         mode = b?Mode.TEXT:Mode.BINARY;
         System.out.println("Running in "+mode.name().toLowerCase(Locale.ENGLISH)+" mode");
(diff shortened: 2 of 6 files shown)


src/main/java/hudson/remoting/InitializeJarCacheMain.java
```diff
@@ -34,7 +34,7 @@ public boolean accept(File dir, String name) {
     public static void main(String[] argv) throws Exception {
         if (argv.length != 2) {
             throw new IllegalArgumentException(
-                    "Usage: java -cp slave.jar hudson.remoting.InitializeJarCacheMain " +
+                    "Usage: java -cp agent.jar hudson.remoting.InitializeJarCacheMain " +
```
oleg-nenashev: It's compliant with https://github.com/jenkinsci/jenkins/pull/3004 from @ndeloof . LGTM though I would prefer to just use `remoting.jar` in the documentation

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -137,12 +137,12 @@ public void setTextMode(boolean b) {
     @Option(name="-jnlpUrl",usage="instead of talking to the master via stdin/stdout, " +
             "emulate a JNLP client by making a TCP connection to the master. " +
             "Connection parameters are obtained by parsing the JNLP file.")
-    public URL slaveJnlpURL = null;
+    public URL agentJnlpURL = null;
 
     @Option(name="-jnlpCredentials",metaVar="USER:PASSWORD",usage="HTTP BASIC AUTH header to pass in for making HTTP requests.")
-    public String slaveJnlpCredentials = null;
+    public String agentJnlpCredentials = null;
```
oleg-nenashev: 🐛 breaks the binary compatibility

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -137,12 +137,12 @@ public void setTextMode(boolean b) {
     @Option(name="-jnlpUrl",usage="instead of talking to the master via stdin/stdout, " +
             "emulate a JNLP client by making a TCP connection to the master. " +
             "Connection parameters are obtained by parsing the JNLP file.")
-    public URL slaveJnlpURL = null;
+    public URL agentJnlpURL = null;
```
oleg-nenashev: 🐛 breaks the binary compatibility

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -701,9 +701,9 @@ private static void ttyCheck() {
             // we seem to be running from interactive console. issue a warning.
             // but since this diagnosis could be wrong, go on and do what we normally do anyway. Don't exit.
             System.out.println(
-                    "WARNING: Are you running slave agent from an interactive console?\n" +
+                    "WARNING: Are you running agent.jar from an interactive console?\n" +
```
oleg-nenashev: Would be better to avoid ".jar" here, Remotig lib may have another name && may be packaged into WAR/*.jnlp/whatever

src/main/java/hudson/remoting/InitializeJarCacheMain.java
jsoref: Sorry, what do you mean by "documentation"? I can change this to remoting.jar if you prefer.

jsoref:
@oleg-nenashev : this is mostly updated (including the wiki link). I can make the other change if you like....

src/main/java/hudson/remoting/InitializeJarCacheMain.java
oleg-nenashev: Let's keep it as is for now. It's definitely a step forward, switching to `remoting.jar` is a separate story (which is probably not required at all)

jsoref:
@oleg-nenashev : fixed.

oleg-nenashev:
Test timeout is unrelated for sure.
@jsoref I will merge it now. Just for the future, it is not recommended to squash commits while the pull request is under review. They can be easily squashed during the merge.

PR #252, opened 2018-02-06
[JENKINS-49415] Add additional UncaughtExceptionHandler

rysteboe:
@reviewbybees

https://issues.jenkins-ci.org/browse/JENKINS-49415


oleg-nenashev:
Please create a JIRA issue if you want it to be backported

dwnusbaum:
Test failure looks like an infra problem. I did a quick search to see if there were any other places that didn't set an exception handler, but the only place I found was in [test code](https://github.com/jenkinsci/remoting/blob/94e90bbdf656d9b2695aa27282d5ad86deda342e/src/test/java/org/jenkinsci/remoting/protocol/IOHubRule.java#L93) so I don't think that should be an issue.

oleg-nenashev:
Would be nice to clarify the root cause of the issue in the ticket

PR #254, opened 2018-02-09
[JENKINS-49472] Log channel name if command not constructed

rysteboe:
@reviewbybees

https://issues.jenkins-ci.org/browse/JENKINS-49472


rysteboe:
@oleg-nenashev, I'm going to go ahead and merge so that we can get a snapshot with the relevant changes, but please let me know if there's something I overlooked that should be added in a later PR.

oleg-nenashev:
Works for me though I am not sure why would you need to merge it to spin snapshots. Especially before the PR builder passes/fails

PR #255, opened 2018-02-09
DO NOT MERGE - Temporary logging

rysteboe:
https://issues.jenkins-ci.org/browse/JENKINS-49480

I could not figure out what would cause the channel to not terminate, so I added an extreme amount of logging in an effort to determine that.  Unfortunately, I cannot reproduce the issue locally.

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -1007,6 +1007,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     @java.lang.SuppressWarnings("ToArrayCallWithZeroLengthArrayArgument")
     @SuppressWarnings("ITA_INEFFICIENT_TO_ARRAY") // intentionally; race condition on listeners otherwise
     public void terminate(@Nonnull IOException e) {
+        logger.log(Level.FINE, "Attempting to terminate channel");
         
         if (e == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -1030,21 +1031,29 @@ public void terminate(@Nonnull IOException e) {
...
                         executingCalls.clear();
                     }
+                    logger.log(Level.WARNING, "Aborting exported objects");
                     exportedObjects.abort(e);
                     // break any object cycles into simple chains to simplify work for the garbage collector
+                    logger.log(Level.WARNING, "Clearing reference");
                     reference.clear(e);
+                    logger.log(Level.WARNING, "Successfully terminated channel, still need to notify listeners");
+                } catch (Throwable t) {
+                    logger.log(Level.WARNING, "Caught error while terminating channel", t);
```
oleg-nenashev: Caught **unhandled** error. Maybe it should be severe since it may be fatal to the upstream logic

rysteboe: Ah whoops, forgot to commit the change where I rethrew it.  Thanks for the catch.

rysteboe:
If any of this logging proves useful, it will have its level dramatically reduced before being merged.  Also added in the channel names.

PR #257, opened 2018-02-23
Update Remoting to Parent POM 1.40 with integrated FindBugs check

oleg-nenashev:
Downstream of https://github.com/jenkinsci/pom/pull/18

@reviewbybees @dwnusbaum @rysteboe


oleg-nenashev:
Hmm, artifact has already been wiped somehow

PR #262, opened 2018-03-10
[JENKINS-49618] - Display Remoting version in the agent log when starting up the agent

pvtuan10:
This PR is to solve newbie-friendly issue in https://issues.jenkins-ci.org/browse/JENKINS-49618.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -232,6 +232,7 @@ public class Engine extends Thread {
      *               This method can be used for testing startup logic.
      */
     /*package*/ void startEngine(boolean dryRun) throws IOException {
+        LOGGER.log(Level.INFO, "Using Remoting version: {0}", Launcher.VERSION);
         @CheckForNull File jarCacheDirectory = null;
         
         // Prepare the working directory if required


src/main/java/hudson/remoting/Engine.java
```diff
@@ -268,7 +268,7 @@ public synchronized void startEngine() throws IOException {
         } else {
             LOGGER.log(Level.INFO, "Using custom JAR Cache: {0}", jarCache);
         }
-        
+        LOGGER.log(Level.INFO, "Using agent version: {0}", Launcher.VERSION);
```
oleg-nenashev: It would be better to do it earlier, preferably at the beginning of the method. Work Dir and Jar Cache initialization may cause fatal failures, and it would be nice to show version info before that.

`workDirManager.setupLogging()` may impact the behavior of `LOGGER`, but it should be fine to print the message before it. In such case the message will just go to initial `java.util.logging` appenders like CLI.

pvtuan10:
@oleg-nenashev I made the change as you proposed.

oleg-nenashev:
FYI @rysteboe

oleg-nenashev:
To be released tomorrow

PR #263, opened 2018-03-19
[JENKINS-50237] When returning an exception to a UserRequest, always include a ProxyException

jglick:
[JENKINS-50237](https://issues.jenkins-ci.org/browse/JENKINS-50237)

Inspired by https://github.com/jenkinsci/workflow-api-plugin/pull/64, but more complicated in this case since

* we prefer to retain the original exception type when possible
* we cannot be sure that the remote side is running a similarly updated Remoting

### DIFF
--- a/src/main/java/hudson/remoting/UserRequest.java
+++ b/src/main/java/hudson/remoting/UserRequest.java
@@ -354,6 +354,7 @@ final class UserRequest<RSP,EXC extends Throwable> extends Request<UserRequest.R
                 if (t == null) {
                     t = (Throwable) deserialize(channel, proxyResponse, cl);
                 }
+                channel.attachCallSiteStackTrace(t);
                 throw (EXC) t;
             } finally {
                 Channel.setCurrent(old);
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/UserRequest.java
```diff
@@ -51,7 +51,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
-final class UserRequest<RSP,EXC extends Throwable> extends Request<UserResponse<RSP,EXC>,EXC> {
+final class UserRequest<RSP,EXC extends Throwable> extends Request<UserRequest.ResponseToUserRequest<RSP,EXC>,EXC> {
```
oleg-nenashev: fine for package-internal class. I see no historical external usages.

jglick:
> It adds some overhead for the error propagation

Yes; I figured exception return values should be rare enough that it does not matter much.

> In order to fix JENKINS-50237 by this patch, a change needs to be applied on both master and agent side (Remoting update is required)

Right in general, though https://github.com/jenkinsci/jenkins/pull/3359 independently fixes the problem for the normal case of `FilePath.list`.

> in the case of JENKINS-50237 we lose the info about location

Actually we do not; discussed in the other PR.

jglick:
BTW if you are not going to fix your CI build to use the Azure mirror, you should just disable it.

PR #267, opened 2018-04-24
[JENKINS-50965] Fix log message

pjdarton:
Current code logs `INFO: %s class '%s' using classloader: %s` if system property `hudson.remoting.RemoteClassLoader.force` is set.  This is ugly and unhelpful.
The code (introduced in #82) should be using `java.text.MessageFormat.format` syntax but is (mistakenly) using `String.format` syntax instead.

I've checked the (new) log format locally (manually, as there's no unit-test for this code) and confirmed that the new format strong would log a message of the form `INFO: Loaded class 'com.foo.Bar' using classloader: com.foo.SomeClassLoader`.

### DIFF
--- a/src/main/java/hudson/remoting/UserRequest.java
+++ b/src/main/java/hudson/remoting/UserRequest.java
@@ -178,7 +178,7 @@ final class UserRequest<RSP,EXC extends Throwable> extends Request<UserRequest.R
                 final Logger logger = Logger.getLogger(RemoteClassLoader.class.getName());
                 if( logger.isLoggable(logLevel) )
                 {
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@pjdarton A JIRA ticket would be helpful if you want it to be backported to LTS. If you are fine with weekly, it can be merged as is

pjdarton:
I run LTS at work, so I'd much prefer it to be backported to LTS if that's viable.  Just as long as it's not the only thing being backported - it's not worth a new LTS revision on it's own ;-)

I've created [JENKINS-50965](https://issues.jenkins-ci.org/browse/JENKINS-50965) for this.

PR #269, opened 2018-05-09
[JENKINS-51223] catch more domain suffix based no_proxy matches (add …

gabemontero:
…curl-like algorithm to existing regexp)

proposed fix for https://issues.jenkins-ci.org/browse/JENKINS-51223

adds a simpler suffix based matching like we see in https://github.com/curl/curl to the existing regexp

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -30,7 +30,8 @@ import java.nio.file.Path;
  *
  * @author Kohsuke Kawaguchi
  */
-class Util {
+@Restricted(NoExternalUse.class)
+public class Util {
     /**
      * Gets the file name portion from a qualified '/'-separate resource path name.
      *
@@ -108,7 +109,7 @@ class Util {
      *
      * Warning: this method won't match shortened representation of IPV6 address
      */
-    static boolean inNoProxyEnvVar(@Nonnull  String host) {
+    public static boolean inNoProxyEnvVar(@Nonnull  String host) {
         String noProxy = System.getenv("no_proxy");
         if (noProxy != null) {
             noProxy = noProxy.trim()
@@ -124,6 +125,7 @@ class Util {
                 }
                 else {
                     int depth = 0;
+                    String originalHost = host;
                     // Loop while we have a valid domain name: acme.com
                     // We add a safeguard to avoid a case where the host would always be valid because the regex would
                     // for example fail to remove subdomains.
@@ -135,8 +137,13 @@ class Util {
                         if (noProxy.matches(".*(^|,)\\Q" + host + "\\E($|,).*"))
                             return true;
                         // Remove first subdomain: master.jenkins.acme.com -> jenkins.acme.com
-                        else
-                            host = host.replaceFirst("^[a-z0-9]+(-[a-z0-9]+)*\\.", "");
+                        host = host.replaceFirst("^[a-z0-9]+(-[a-z0-9]+)*\\.", "");
+                    }
+                    
+                    String[] noProxyArray = noProxy.split(",");
+                    // fix for https://issues.jenkins-ci.org/browse/JENKINS-51223, basic suffix match
+                    if (depth > 0 && suffixMatch(originalHost, noProxyArray)) {
+                    	return true;
                     }
                 }
             }
@@ -144,6 +151,20 @@ class Util {
 
         return false;
     }
+
+    // fix for https://issues.jenkins-ci.org/browse/JENKINS-51223
+    // adds curl-like algorithm for matching to existing regexp used in
+    // inNoProxyEnvVars
+    static boolean suffixMatch(String host, String[] noProxyArray) {
+		for (String proxy : noProxyArray) {
+			// still needs to capture some form of subdomain, like ".svc"
+			if (!proxy.contains("."))
+				continue;
+			if (host.endsWith(proxy))
+				return true;
+		}
+		return false;
+	}
 
     /**
      * Gets URL connection.
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -24,6 +24,8 @@
 package org.jenkinsci.remoting.engine;
 
 import hudson.remoting.Base64;
+import hudson.remoting.Util;
+
 import org.jenkinsci.remoting.util.https.NoCheckHostnameVerifier;
 import org.jenkinsci.remoting.util.https.NoCheckTrustManager;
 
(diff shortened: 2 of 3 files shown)


gabemontero:
@bparees FYI

gabemontero:
Hey @jeffret-b 

- I added a couple of entries to UtilTest.java to cover the basics of this change
- regarding code dupe ... yep, of course saw it, including the //FIXME comments around needing to address it;  at least for an initial drop, I wanted to keep the change to a minimum, but can certainly add some form of reduction to the mix, even if it is as simple as having the JnlpAgentEndpointResolver versions of the methods in question call the Util versions, and reconcile any impl differences, though if I recall correctly, they were identical; just let me know how you land on that 

thanks

jeffret-b:
The tests help a lot! They do a lot to demonstrate the exact intention, the problem we are trying to solve.

I think another test is needed. Something like UtilTest.testMixed(). Or maybe just some additions to that test to cover the comma-separated situation with the minimal suffix.

I think there's something a little bit wrong about the current proposal. The new part added into lines 138-141 of Util shouldn't be inside the while loop. I haven't been able to prove that it actually causes problems but it disrupts the logic flow and causes duplication of the new check or early loop termination. The new logic should come after loop completion, with a slight change needed because of how "host" is changed.

In my opinion, it would be fine to eliminate the duplication as part of this PR, especially as this PR would otherwise worsen it. It would be fantastic if you could do that. It might be safer to leave the method in JnlpAgentEndpointResolver and just delegate the implementation to Util. In case, for some odd reason someone else is using that method, though that seems very unlikely. It looks like Util would need to be upgraded to public access. Then it would be best to mark the class with @Restricted(NoExternalUse.class) as in WorkDirManager.

jeffret-b:
There is more info about NoExternalUse here: http://kohsuke.org/access-modifier/apidocs/org/kohsuke/accmod/restrictions/NoExternalUse.html . Basically it can be used to tag classes, fields, or methods. Java (at least before 9) lacks any mechanism for allowing something to be used outside of a package but within an assembly or module. This is a Jenkins specific mechanism for this capability. It helps limit the scope of the public API, enabling better opportunities for future change.

Thanks for looking into those suggestions!

gabemontero:
There is a slight difference in the `openURLConnection` implementations.  I am not going to take on reconciling that difference, determining which is correct, with this PR.

jeffret-b:
Fine. If you already know what it is, I'd appreciate if you could point it out to me.

gabemontero:
It is this snippet with the `FIXME`:

```
            } else if (sslSocketFactory != null) {
                httpsConnection.setSSLSocketFactory(sslSocketFactory);
                //FIXME: Is it really required in this path? Seems like a bug
                httpsConnection.setHostnameVerifier(new NoCheckHostnameVerifier());
            }
```

Now, if you are confident the FIXME line can be removed, I can move forward if you would like.

jeffret-b:
Thanks for pointing it out. Leave it for now.

gabemontero:
Meaning removing `httpsConnection.setHostnameVerifier(new NoCheckHostnameVerifier());`

gabemontero:
ok updates pushed @jeffret-b

gabemontero:
On the pr-merge failure, I see it marks it failed at the `[INFO] Running org.jenkinsci.remoting.engine.JnlpProtocolHandlerTest` test, where it looks like a timeout occurred.

When I run the tests locally, even before my change, I saw that test take a long time, but ultimately complete.

As an example, from my local run:

```
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 50.117 s - in org.jenkinsci.remoting.engine.JnlpProtocolHandlerTest
```

So
a) the duration is of that test seems independent of my change
b) either the pr-merge job needs to be adjusted to account for that, or the test needs to be pruned / adjusted somehow .... I of course don't have permissions to do the former, and I don't have the bandwidth the pursue the latter

jeffret-b:
@gabemontero 

1. Don't worry about that test failure in JnlpProtocolHandlerTest. It happens somewhat sporadically, more on some platforms than others. It doesn't really mean anything. Or also NetworkLayerTest if you see those ones. I'm working on PRs to rework those tests so they'll be reliable.

2. Thanks for the work on eliminating the duplication! That looks great.

3. There was something bugging me yesterday about the difference between your addition and the existing code. I couldn't quite figure it out yesterday. With your revised submission today to move your check first, before the other loop, I was able to quickly figure it out. The problem is that the existing code checks for the validity of domain names using the regex, while yours doesn't. With your latest version, line 143 in Util never gets called by the tests. That whole loop becomes a no-op.

Here are a few tests for UtilTest that demonstrate the problem. They pass before your change and fail afterwards. Unfortunately, these tests didn't exist before to make the existing behavior more clear to us.
    @Test
    public void testNoProxyWithInvalidChars() {
        PowerMockito.when(System.getenv("no_proxy")).thenReturn("foo+.co=m");

        assertEquals(false, Util.inNoProxyEnvVar("foo+.co=m"));
    }

    @Test
    public void testNoProxyWithInvalidChar() {
        PowerMockito.when(System.getenv("no_proxy")).thenReturn("foo.co=m");

        assertEquals(false, Util.inNoProxyEnvVar("foo.co=m"));
    }

    @Test
    public void testNoProxyWithInvalidCharInMinimalSuffix() {
        PowerMockito.when(System.getenv("no_proxy")).thenReturn(".sv=c");

        assertEquals(false, Util.inNoProxyEnvVar("foo.sv=c"));
    }
I'm not entirely sure all this domain name checking was necessary, but I'm hesitant to change the existing behavior.

gabemontero:
Interesting @jeffret-b  .....

In order to address your existing behavior concerns, could we opt into my change, vs. having it on by default, by specifying an env var on the Jenkins agent?

That would work easily with our openshift/kubernetes scenarios, as setting env vars are exposed in a first level fashion with the kubernetes plugin.

jeffret-b:
@gabemontero, in my opinion your proposal is a good one, as long as it is additive and doesn't change existing behavior. I think it's possible to still do that. I think something like these steps would do it.

1. Move your check from before the while loop to after. Your noProxy.split(",") needs to remain before the loop, I believe.

2. Use the value of host at the end of the loop, which I think just contains the final, right-most part of the domain. "svc" in your examples.

3. Continue to check for a leading dot in the no_proxy string.

4. Need to match up the fact that the no_proxy string has a leading dot and host has had it stripped by the end of the loop.

5. For consistency, we probably ought to have a validity check for that final, right-most portion. Something kind of like the regex in the line 
while (host.matches(...
but shortened to only test the one segment. I'm not great at doing regex without lots of checks / tests, but it might be something similar to "^\.[a-z0-9]+". That might also cover item #3.

jeffret-b:
There might be a way to incorporate that sort of behavior directly into the last iteration of the while loop, but I haven't been able to figure it out yet.

gabemontero:
Hey @jeffret-b - getting back to this after having to focus on other things Friday.  About to push a change for most of what you have articulated along these lines

1. add the 3 junits you specified for invalid chars
2. moved my new check to after the for loop, though I do not believe the noProxy.split is needed before the while ... noProxy is not changed in the while, and noProxyArray is not referenced in the while
3. Yep, your item 4) had to be accounted for 
4. regexp is not my strength either, but I think if we just check for depth > 0 along with suffixMatch, we should get the benefit of the regexp validate in the while above and not need to add an additional regexp against "^.[a-z0-9]+" ... that said, in my testing, that did accept valid suffixes (though it did NOT handle the leading do check ... it returned true for both `.svc` and `svc`

Will push the updates in a few moments (unless I catch responses from you beforehand which need attending).

gabemontero:
Ah ... actually, we can't just use the value of host after the while loop ... fails existing tests, where say the host has "foobar.com", but the no_proxy is "jenkins.com" .... will address, but will be a few more before I push.

gabemontero:
OK updates pushed.

jeffret-b:
Ah, yes, this is looking good. I hadn't thought about using the depth for that check, but that should cover it. My other numbered points in my previous comment were based upon how I thought it might go but I hadn't worked it through yet.  Looks like you were able to figure out how to fit it together.

I've got one remaining concern. In Util:165-166 it says:
`			if (!host.contains(".") && proxy.endsWith(host))
				return true;
`
I can't find that this condition is ever hit by the tests. I'm concerned that I haven't been able to figure out what case is covered by this and that there are no tests to demonstrate it.

gabemontero:
OK cool (first paragraph).

On your remaining concern,  that line was my attempt to address item 4) from https://github.com/jenkinsci/remoting/pull/269#issuecomment-390015511
- The first part of the conditional captures "host has had [the leading dot] stripped by the end of the loop"
- And then the second determines if they match

But I believe you are correct that the current tests to seem to stress that point.  I just commented out that line and the existing tests still passed.

I'll see about constructing a test that hits that, and if I can't, remove that line, with some explanation of how your item 4) is handled.

jeffret-b:
I'm satisfied that my #4 is covered by the way you used originalHost and the preceding lines to 165:166. My comment  was regarding something I played around with briefly about using the end result of host, but I think your implementation took care of that. I haven't been able to figure out another condition, so I think it's looking good as long we remove those two lines that are now extraneous.

gabemontero:
ah, you beat me to it (originalHost) :-) 

yep, removing those lines now ... will push momentarily

gabemontero:
update pushed

gabemontero:
my pleasure - likewise thanks

any sense on when a new version of the remoting jar will come out?  or when it will get picked up in jenkins itself?

thanks again

jeffret-b:
I've got to get a few more things in place before I can perform a release. I'm hoping that in a few weeks I can do that. I'm not sure yet what it will take after that to get it picked up in Jenkins itself. It could probably make it in a weekly release within a couple of weeks after I can get the release done.

gabemontero:
Thanks for the update @jeffret-b and yep seems reasonable 

Should we continue to periodically checkpoint either here or in the associated JIRA for updates on status, or is there a better place per y'all's process that I'm overlooking/missing to monitor things?

thanks again

jeffret-b:
@gabemontero This change went into the 3.21 release of Remoting today. That change will start to show up in upcoming core releases.

gabemontero:
Thanks for the update @jeffret-b

PR #270, opened 2018-05-17
[JENKINS-51378] Simplify NetworkLayerTest, removing duplication.

jeffret-b:
This test has been doing essentially the same thing over and over in a very repetitive way. Some slight, but pretty meaningless variations. And a whole bunch of absolute repetitions.

It kind of halfway used theories to reduce duplication, though it still used traditional duplication structures. And the theories mostly ended up just doing the same thing, as far as the Remoting specific code was concerned.

Ultimately, these repetitions were testing whether Java and the JVM works properly. Oddly and unfortunately, the answer is no, in this case. There are indications that issues exist, on certain platforms, in some extreme scenarios. Key among the scenarios is a very rapid creation and teardown of channels, file descriptors, or possibly most significantly pipes. This seems to particularly apply to the Mac OS. Linux can also show it, though in possibly a somewhat different way. This primarily involves nio. These situations can result in hangs in native calls from Java core APIs, post prominently preClose0(). Rather than testing variations in the Remoting library, all of these JVM issues are what these tests have been stressing, and hitting some of these key characteristics.

These hangs have existed since this test was first created. They're sporadic, though they're more likely to appear in some situations than others. As a result, it leads to the impression that these tests have worked correctly at some times, when it's just been a matter of not hitting the causes in exactly the right way.

There are not any indications that these highly repetitive tests hit conditions that are actually important to any user. This rapid-fire creation and teardown operations are not part of regular operation. These test issues have existed for some time, since the tests were first introduced, and are not known to be related to any customer-observed issued.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/remoting/protocol/impl/NetworkLayerTest.java
+++ b/src/test/java/org/jenkinsci/remoting/protocol/impl/NetworkLayerTest.java
@@ -87,6 +87,7 @@ public class NetworkLayerTest {
                         .on(clientFactory.create(hub, serverToClient.source(), clientToServer.sink()))
                         .build(new IOBufferMatcherLayer());
 
+
         ProtocolStack<IOBufferMatcher> server =
                 ProtocolStack
                         .on(serverFactory.create(hub, clientToServer.source(), serverToClient.sink()))
@@ -99,7 +100,8 @@ public class NetworkLayerTest {
         server.get().send(data);
         client.get().awaitByteContent(is(expected));
         assertThat(client.get().asByteArray(), is(expected));
-        server.get().close(null);
+        server.get().close();
+        client.get().awaitClose();
     }
 
 }
--- a/src/test/java/org/jenkinsci/remoting/protocol/impl/NetworkLayerTest.java
+++ b/src/test/java/org/jenkinsci/remoting/protocol/impl/NetworkLayerTest.java
@@ -25,30 +25,20 @@ package org.jenkinsci.remoting.protocol.impl;
 
 import java.nio.ByteBuffer;
 import java.nio.channels.Pipe;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
-import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


kuisathaverat:
It is known that the channel breaks under stress, maybe this is the goal of the test to check that, but your analysis said that these test work in a random mode or not work at all, Is that right?

jeffret-b:
@kuisathaverat, that's a concern that I struggled with for a long time.

First, a little clarification on what I've been able to observe on these failures. They manifest as hangs, which eventually time out, in all cases I've seen. I haven't observed the failures on Windows, by the CI environment or by any reports. On my Mac, I see the failures most of the time, though just infrequent enough to make it really frustrating and error prone. We see the failures sometimes on Linux, though less frequently. These match fragmentary discussions I've found about issues in low-level native JVM code on these platforms.

It's possible that these tests were intended to catch these very conditions under stress. I haven't been able to dig up enough information yet to determine if that is the case. If so, eliminating or simplifying these tests could be problematic. However, the tests don't really simulate a condition that exists in production. I'm unaware of any reports that these hangs are observed in production use. Even if these problems exist in production, these repeated tests are not doing anything to help us correct the problem. The hangs / failures get in the way of other changes. If these failures exist in production, we need to do something different to correct them.

src/test/java/org/jenkinsci/remoting/protocol/impl/NetworkLayerTest.java
```diff
@@ -159,193 +99,7 @@ public void doCloseRecv(NetworkLayerFactory serverFactory, NetworkLayerFactory c
         server.get().send(data);
         client.get().awaitByteContent(is(expected));
         assertThat(client.get().asByteArray(), is(expected));
-        server.get().closeRead();
-        client.get().awaitClose();
```
dwnusbaum: I think you should add `client.get().awaitClose()` back to make sure that the client is closed when the server is closed.

jeffret-b:
I was never able to locate an exact, filed JDK bug to reference for this. If I search hard enough I find various fragmentary discussions around this behavior, particularly preClose0(). Everybody that sees it seems to be in about the same position as ourselves -- unable to isolate or reproduce it sufficiently to submit it.

PR #271, opened 2018-05-17
[JENKINS-51379] Simplify JnlpProtocolHandlerTest.

jeffret-b:
Some of this is similar to the simplifications for NetworkLayerTest in e9c1af14 (See #270). This test has also been hitting sporadic hangs / failures, primarily on certain OSes, etc. These mostly occur in preClose0() also. Similar to that test, part of the solution here is to eliminate the repetition of the exact same test.

As part of understanding and cleaning up this test, I also made a major effort to remove other duplication. And to improve naming and clarity. The individual test methods were lengthy, with lots of duplicated code, making it hard to understand what they were doing. Indeed some of the tests didn't actually match their name / intention. I extracted out the commonality and tried to make it clearer the variation in each test.

This test had continued testing all old versions of the JNLP protocol. There had been TODOs marked to remove them from this test. They've been deprecated for a long time. Their implementations aren't quite behaving properly with respect to this test, particularly when it comes to closing them down, which causes contamination with testing the currently supported protocols. With Oracle deprecating JNLP entirely it's time to remove these old versions of the protocols from this test.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
+++ b/src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
@@ -3,6 +3,7 @@ package org.jenkinsci.remoting.engine;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.remoting.Channel;
 import hudson.remoting.TestCallable;
+
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
```diff
@@ -110,594 +113,197 @@
...
-        eastRemoting.callAsync(new TestCallable());
-        westRemoting = westChan.get(10, TimeUnit.SECONDS);
-        assertThat(westRemoting, notNullValue());
+        JnlpProtocolHandler<? extends JnlpConnectionState> serverProtocolHandler = createServerProtocolHandler(factory, useNioHubServer, SECRET_KEY, true);
+        JnlpProtocolHandler<? extends JnlpConnectionState> clientProtocolHandler = createClientProtocolHandler(factory, useNioHubClient);
+        HashMap<String, String> clientProps = createClientProperties(factory, SECRET_KEY);
+        Future<Channel> clientChannelFuture = createChannelConnector(clientSocketChannel, clientProtocolHandler, clientProps, APPROVING_STATE_CONSUMER);
+        readAndCheckProtocol(factory);
+        Future<Channel> serverChannelFuture = createChannelHandler(serverSocketChannel, serverProtocolHandler, new HashMap<>(), APPROVING_STATE_CONSUMER);
+        serverRemotingChannel = serverChannelFuture.get(10, TimeUnit.SECONDS);
```
jvz: It may be overkill in this scenario, but I like to use asynchronous-aware libraries when testing async code. For Java, I found [awaitility](https://github.com/awaitility/awaitility) which looks interesting. This type of pattern is a lot more common in Scala since they've had a `Future` class far longer than Java's actual equivalent.

jeffret-b: That looks like an interesting testing library. I think there's enough change already in this PR that I'd rather not introduce that one at this point.

PR #272, opened 2018-05-26
[JENKINS-51551] Allow CommandTransport and its sub-classes to be sub-typed from outside

pvtuan10:
JIRA task: https://issues.jenkins-ci.org/browse/JENKINS-51551

While working on the design for [JENKINS-51413](https://issues.jenkins-ci.org/browse/JENKINS-51413), I need to implement a new command transport to support Kafka communication. However, `CommandTransport` and `SynchronousCommandTransport` are not allowed to be sub-typed outside of the package because some of the fields, methods were made package-private. The proposed solution to this is to change them to protected access level.

@jenkinsci/code-reviewers @oleg-nenashev @jeffret-b @dwnusbaum @Supun94

### DIFF
--- a/src/main/java/hudson/remoting/CommandTransport.java
+++ b/src/main/java/hudson/remoting/CommandTransport.java
@@ -140,7 +140,7 @@ public abstract class CommandTransport {
      *      Informational flag that indicates that this is the last
      *      call of the {@link #write(Command, boolean)}.
      */
-    protected abstract void write(Command cmd, boolean last) throws IOException;
+    public abstract void write(Command cmd, boolean last) throws IOException;
 
     /**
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
Deployed the snapshot as 3.21-20180526.122857-1

pvtuan10:
thanks a lot @oleg-nenashev

oleg-nenashev:
@pvtuan10 I think this change is ready to go. Are you fine with that?
We have another API change, which may be required for https://issues.jenkins-ci.org/browse/JENKINS-51709. But it can be done in a separate PR IMHO

pvtuan10:
@oleg-nenashev This change is fine for me.

PR #265, opened 2018-04-23
[JENKINS-50933] - Integration tests with Jenkins Core

oleg-nenashev:
Currently Remoting has not Jenkins integration tests in the repository. It would be great to run a subset of tests in Remoting PR builder, including Jenkins Core JTH, Plugin PCT and ATH

- [x] Implement a sample flow with ATH and PCT for several plugins using Remoting: command-launcher, windows-slaves, ssh-slaves
- [x] Integrate the flow with existing tests

Downstream of (not blockers):

* https://github.com/jenkinsci/plugin-compat-tester/pull/74
* https://github.com/jenkinsci/custom-war-packager/pull/20

https://issues.jenkins-ci.org/browse/JENKINS-50933

@reviewbybees

PR #276, opened 2018-06-10
[JENKINS-51841] - Offer a method for standardized command deserialization from the channel

oleg-nenashev:
Remoting Kafka Plugin implements its own Command Transport, which creates Command Instances. Proper deserialization of Commands requires setting proper classloader and a local channel instances.

There is a Command#readFrom() method which does it, but it's package-private. Also, a construction of the ObjectInputStreamEx requires package-private Channel fields.

I propose to create a public Channel#readFrom(Channel, byte[] payload) method so that it can be used externally

- [x] - Offer a new method
- [x] - Refactor existing logic to use the new method

Reference implementation: https://github.com/pvtuan10/remoting-kafka-plugin/pull/1

@pvtuan10

### DIFF
--- a/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
+++ b/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
@@ -60,10 +60,7 @@ public abstract class AbstractByteArrayCommandTransport extends CommandTransport
         setup(new ByteArrayReceiver() {
             public void handle(byte[] payload) {
                 try {
-                    Command cmd = Command.readFrom(channel, new ObjectInputStreamEx(
-                            new ByteArrayInputStream(payload),channel.baseClassLoader,channel.classFilter));
-                    receiver.handle(cmd);
-                    channel.notifyRead(cmd, payload.length);
+                    Command cmd = Command.readFrom(channel, payload);
                 } catch (IOException | ClassNotFoundException e) {
                     LOGGER.log(Level.WARNING, "Failed to construct Command in channel " + channel.getName(), e);
                 }
--- a/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
+++ b/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
@@ -61,6 +61,7 @@ public abstract class AbstractByteArrayCommandTransport extends CommandTransport
             public void handle(byte[] payload) {
                 try {
                     Command cmd = Command.readFrom(channel, payload);
+                    receiver.handle(cmd);
                 } catch (IOException | ClassNotFoundException e) {
                     LOGGER.log(Level.WARNING, "Failed to construct Command in channel " + channel.getName(), e);
                 }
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
```diff
@@ -199,10 +199,8 @@ private void processCommand() throws IOException {
         try {
             FastByteBufferQueueInputStream is = new FastByteBufferQueueInputStream(receiveQueue, readCommandSizes[0]);
             try {
-                ObjectInputStreamEx ois = new ObjectInputStreamEx(is, channel.baseClassLoader, channel.classFilter);
-                Command cmd = Command.readFrom(channel, ois);
+                Command cmd = Command.readFrom(channel, is, readCommandSizes[0]);
                 receiver.handle(cmd);
-                channel.notifyRead(cmd, readCommandSizes[0]);
```
oleg-nenashev: @jglick I believe here was a glitch in the handling. Read notification should have happened before handling. Fixed by the patch. Please correct me if I miss something.

src/main/java/hudson/remoting/Command.java
```diff
@@ -108,9 +111,41 @@ void writeTo(Channel channel, ObjectOutputStream oos) throws IOException {
...
+    /**
+     * Reads command from the specified payload.
+     * @param channel Channel
+     * @param istream Input stream
+     * @param payloadSize Payload size to be read. Used only for logging
+     * @return Read command
+     * @throws IOException Read exception
+     * @throws ClassNotFoundException Deserialization error: class not found
+     */
+    /*package*/ static Command readFrom(@Nonnull Channel channel, @Nonnull InputStream istream, int payloadSize)
```
oleg-nenashev: I do not like payload size handling, so I decided to keep it private for now

src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
```diff
@@ -60,10 +60,7 @@ public final void setup(final Channel channel, final CommandReceiver receiver) {
         setup(new ByteArrayReceiver() {
             public void handle(byte[] payload) {
                 try {
-                    Command cmd = Command.readFrom(channel, new ObjectInputStreamEx(
-                            new ByteArrayInputStream(payload),channel.baseClassLoader,channel.classFilter));
-                    receiver.handle(cmd);
```
jglick: Uh, what happened to the `CommandReceiver.handle` call?

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
jglick: Not sure it really matters, but did you test it? (Integration test in `support-core`.) Anyway we only cared about `NioTransport`s.

oleg-nenashev:
@jglick fixed

src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java
oleg-nenashev: Testing

oleg-nenashev:
@jglick WRT Core test, stay tuned for the reincarnation of https://github.com/jenkinsci/remoting/pull/265

PR #282, opened 2018-08-20
[JENKINS-52945] AnonymousClassWarnings should not warn about enums

jglick:
Recently noticed this complaining about a class in `credentials` plugin, and found that the warning was spurious in that case.


jglick:
Found that this was filed as [JENKINS-52945](https://issues.jenkins-ci.org/browse/JENKINS-52945).

oleg-nenashev:
Yes, a release would be nice. Maybe even a release candidate

jeffret-b:
A release has been in my plans. I can bump its priority and get it done soon.

Would you like a release candidate instead?

batmat:
@jeffret-b whatever makes it easier for you. Any release would be great.

PR #283, opened 2018-08-20
Fix [JENKINS-42533] Trying to unexport already unexported object

jeffret-b:
See [JENKINS-42533](https://issues.jenkins-ci.org/browse/JENKINS-42533)

In rare cases, the Unexport command gets a little out of sequence and attempts to unexport an object that has already been unexported. It logs this at SEVERE level, when there is no indication of any harm. The issue is marked as minor.

Implement the fix in a similar pattern to [JENKINS-22853](https://issues.jenkins-ci.org/browse/JENKINS-22853), as noted in the PR. That fix reduced the log level for a similar message coming from the EOF command. As with that fix, it is unclear how to create a meaningful test, particularly as it is only a log level change..

### DIFF
--- a/src/main/java/hudson/remoting/ProxyOutputStream.java
+++ b/src/main/java/hudson/remoting/ProxyOutputStream.java
@@ -341,7 +341,7 @@ final class ProxyOutputStream extends OutputStream implements ErrorPropagatingOu
         protected void execute(final Channel channel) {
             channel.pipeWriter.submit(ioId,new Runnable() {
                 public void run() {
-                    channel.unexport(oid,createdAt);
+                    channel.unexport(oid,createdAt, false);
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/ProxyOutputStream.java
```diff
@@ -341,7 +341,7 @@ public Unexport(int ioId, int oid) {
         protected void execute(final Channel channel) {
             channel.pipeWriter.submit(ioId,new Runnable() {
                 public void run() {
-                    channel.unexport(oid,createdAt);
+                    channel.unexport(oid,createdAt, false);
```
MRamonLeon: minor comment: it seems like the standard is not to leave a blank after the comma. As I see in the arguments of the _submit_ method and in the old arguments of the unexport method. Do you consider removing the blank before the _false_?

src/main/java/hudson/remoting/ProxyWriter.java
```diff
@@ -377,7 +377,7 @@ public Unexport(int ioId, int oid) {
         protected void execute(final Channel channel) {
             channel.pipeWriter.submit(ioId,new Runnable() {
                 public void run() {
-                    channel.unexport(oid,createdAt);
+                    channel.unexport(oid,createdAt, false);
```
MRamonLeon: Ditto.

jeffret-b:
Yes, it is just a change to the log level. That is primarily what Oleg changed in the referenced fix. I think that change is also appropriate in this similar situation.

PR #286, opened 2018-09-07
[JENKINS-53461] - Add support of connecting to masters without Web UI

oleg-nenashev:
Currently Remoting agents always require Jenkins master to have a Tcp Agent Listener endpoint. It complicates usage of Remoting with CloudAPI-provided agents together with Jenkinsfile Runner which does not have a fixed Web UI port.

I propose to add new options to Remoting so that it can connect without polling metadata from the master

- [x] - Introduce new CLI options and adjust JNLP Endpoint Resolver
- [x] - Add Dockerfile for https://hub.docker.com/r/jenkins4eval/remoting/
- [x] - Add support of Instance Identity key so that JNLP4 protocol can be used
- [ ] - Add support of passing the list of the protocols to be used 

Sample command:

```
java -cp remoting-3.27-SNAPSHOT.jar hudson.remoting.jnlp.Main -noHttpEndpoint -url http://127.0.0.1:50000 -headless e2a312c17a3a2191786bf54825649ab7e2e33ebaae5b3e913bf90c9fdf9f510e test
```

https://issues.jenkins-ci.org/browse/JENKINS-53461

@carlossg @jglick @jeffret-b @dwnusbaum


Dockerfile
```diff
@@ -0,0 +1,40 @@
+#
+# This Dockerfile builds Experimental Remoting image for jenkins4eval/remoting
+# This image is similar to jenkins/slave, so extra tweaks need to be applied for better UX
+#
```
jeffret-b: I'm not clear on exactly what this Dockerfile is used for or how it relates to the jenkins/slave one. This looks like a generic Dockerfile, which will build an image from the current local source. I'm confused about the comment regarding how this is for "jenkins4eval/remoting". Maybe it's just the comment that is confusing me and this can be used to build an image for any current version. Basically the same as jenkinsci/docker-slave, except it builds from local source.

Should these two Dockerfiles be in the same space or otherwise better connected to each other?

oleg-nenashev:
@jeffret-b will try to address the comments ASAP

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -168,6 +189,34 @@ public void setDisableHttpsCertValidation(boolean disableHttpsCertValidation) {
...
+            final int port;
+            if (tunnel != null) {
+                String[] tokens = tunnel.split(":", 3);
+                if (tokens.length != 2)
+                    throw new IOException("Illegal tunneling parameter: " + tunnel);
+                host = tokens[0];
+                port = Integer.parseInt(tokens[1]);
+            } else if (!jenkinsUrls.isEmpty()) {
+               // URL like tcp://myjenkins:50000
+               URL url = new URL(jenkinsUrls.get(0));
```
jglick: Use `URI` to parse, since actually using `tcp` will throw a `MalformedURLException` since that is not a real protocol. And recommend using `tcp` in documentation; it is very confusing to pass an `http` URL here, when there is no HTTP service there! For that matter, do we even need to support `-url tcp://myjenkins:50000` when `-tunnel myjenkins:50000` would do just as well? And do we even need the `-noHttpEndpoint` argument if `-tunnel` is being used?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -306,6 +311,23 @@ public URL getHudsonUrl() {
         return hudsonUrl;
     }
 
+    /**
+     * Disable Http Enfpoint check if needed.
```
alxsap: Typo
```suggestion
     * Disable Http Endpoint check if needed.
```

oleg-nenashev:
I have changed the teams, and I have been unable to get this work prioritized since that. No ETA so far, but I hope I will find some time to get it over the line

jglick:
Nothing against this enhancement per se, but it seems to me that the more fundamental problem is cloud providers used with JFR which only work with “JNLP”-type agents, such as the `kubernetes` plugin currently. We should rather spend a bit of time improving them to make outbound connections (for example, using what I understand to be a WebSocket API underlying `kubectl exec`) so that the JFR container need expose no ports whatsoever.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -324,6 +333,9 @@ public void run() throws Exception {
               jnlpArgs.add("-jar-cache");
               jnlpArgs.add(jarCache.getPath());
             }
+            if (this.disableHttpEndpointCheck) {
```
alxsap: Isn't the `slaveJnlpUrl` in the surrounding `if` an http URL? Does the `disableHttpEndpointCheck` parameter make sense here then?

alxsap:
We implemented the third missing option:
>  Add support of Instance Identity key so that JNLP4 protocol can be used

You can see the changes in this [dummy pull request](https://github.com/sap-production/jenkinsci-remoting/pull/1).

Connecting to a Jenkins without http port now already works. You can test it with the command below. 

```sh
java -cp slave.jar hudson.remoting.jnlp.Main \
  -headless -noHttpEndpoint \
  -tunnel <jenkinsHost>:<jnlpPort> \
  -instanceIdentity <instanceIdentityString> \
  <secretString> <agentName>
```

- The agent needs to be prepared on the master  - copy the name and secret.  
- The instanceIdentityString is the base64 encoded byte array of the masters Instance Identity. You can get it via `hudson.remoting.Base64.encode(org.jenkinsci.main.modules.instance_identity.InstanceIdentity.get().getPublic().getEncoded())`.

### Background
Let me quickly describe what we are going to do and why we need this change.   
We are using Jenkinsfile runner (in a Docker image built by Custom War Packager) to execute single Pipelines on a Kubernetes cluster. Those pipelines use the Kubernetes plugin to execute single pipeline steps in specific other Docker images. And this is the point where it currently breaks since the Jenkinsfile runner does (rightly) not provide an http port and since the Kubernetes plugin relies on the current JNLP mechanism and requires an http port.

The changes in this pull request are the easiest way to get our scenario working.   
@jglick I agree there are 'more fundamental problems', but those changes here would provide a quick (still not dirty) solution to speed up the Jenkins Cloud Native efforts. Once there is more time and resources the more fundamental problems can be tackled with the experience we gained. What do you think?

holgpar:
@jglick 

> We should rather spend a bit of time improving them to make outbound connections (for example, using what I understand to be a WebSocket API underlying kubectl exec) so that the JFR container need expose no ports whatsoever.

I am not sure what you have in mind but extending the use of connections like those used for `kubectl exec` is most likely not a good idea.
These connections are always relayed over the K8S API Server and technically they go into the "wrong" direction (from the K8S API Sever to the K8S node). In K8S, the business as usual communication is always from one component (the kubelet for example) to the API server. Only for debugging functionalities like exec, logs viewing, and port forwarding the order of direction is reversed.

That the Jenkins K8S plugin relies so strongly on `kubectl exec` is one of its weaknesses. To alleviate this the durable task plugin is being used to avoid having long standing connections.

If pods have to communicate with each other they should do so via the pod network and not via the API server.

For these reasons I think that in the K8S context the approach of  @alxsap is actually the more sustainable one.

jglick:
> These connections are always relayed over the K8S API Server and technically they go into the "wrong" direction

Yes, I have brought up these concerns with @carlossg (`kubernetes` plugin maintainer). He seemed to think that the WebSocket connection used for the like of `kubectl exec` (from what I gather) would be able to handle modest Remoting traffic (especially if we had a [JEP-210](https://jenkins.io/jep/210) implementation for something like in-cluster Fluentd), but certainly for heavier loads it would be preferable to have a direct TCP socket between the master pod and the agent pod.

That does not, however, mean that our only option is to use inbound connections via the (misleadingly named) `JNLPLauncher` + `JnlpSlaveAgentProtocol4`. There are other prebuilt outbound connectors; for example you could throw together a solution using the `ssh-slaves` plugin if you generated a new keypair for each launch, set the public key as an agent pod variable, and had the agent expose port 22. Nor is it difficult to write a new launcher with another transport, as all you need is something that can send and receive opaque `byte[]` frames somehow—for example, a socket with blocking I/O, or something using gRPC—especially if TLS can be injected by something else like Istio.

jglick:
To be clear, I am not objecting to this PR (neither am I reviewing it carefully)—just commenting that it should be considered a workaround until we can improve the `kubernetes` plugin to support connecting to pods without exposing any port on the master.

oleg-nenashev:
Personally I think that we rather want to use other agent types in Cloud, like Remoting-over-Kafka agent or similar ones. But I still want to get this feature over the line in the next few weeks. I will coordinate with reviewers about addressing comments and integrating extra changes from @alxsap

alxsap:
Hello @oleg-nenashev, what is the status of this pull request? Is there anything I can do?
We currently use a fork based on this PR to make Jenkinsfile Runner work on Kubernetes together with Kubernetes plugin.

oleg-nenashev:
Hi Alexander, It would be great if you submit a pull request against my
fork so that I could integrate your changes into this pull request. I still
keep this PR in my task queue, and I would love to land it eventually

alxsap:
Hi Oleg, I created PR https://github.com/jenkinsci/remoting/pull/323

alxsap:
Hello @oleg-nenashev, any progress?
Thanks a lot, Alex

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -362,6 +411,11 @@ private URL toAgentListenerURL(@Nonnull String jenkinsUrl) throws MalformedURLEx
     }
 
     public void waitForReady() throws InterruptedException {
+        if (disableHttpEndpointCheck) {
+            LOGGER.log(Level.INFO, "HTTP Endpoint Check is disabled, assuming that the target master is reasy");
```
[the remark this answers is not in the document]
oleg-nenashev:
```suggestion
            LOGGER.log(Level.INFO, "HTTP Endpoint Check is disabled, assuming that the target master is ready");
```

oleg-nenashev:
i was waiting for @jeffret-b to provide feedback in #323 . Went ahead and integrated all the chages into a single branch. Looks like it is code-complete, but we may need to add some documentation for the new feature

jeffret-b:
> was waiting for @jeffret-b to provide feedback in #323 . Went ahead and integrated all the chages into a single branch. Looks like it is code-complete, but we may need to add some documentation for the new feature

I was expecting you to merge that into here to have a comprehensive proposal. :) Guess we were waiting for each other. I'll review it in the next few days as I can get a chance.

> [x] - Add Dockerfile for https://hub.docker.com/r/jenkins4eval/remoting/

I see this still in the original description, but it looks like you removed that part. Am I reading correctly?

alxsap:
> we may need to add some documentation for the new feature

Hello @oleg-nenashev, here a proposal for documentation: https://github.com/sap-production/jenkinsci-remoting/commit/c1a23b0a461c16a9e2acdb525a52f771ed83e049
It looks [like this](https://github.com/sap-production/jenkinsci-remoting/blob/PR_286_docu/docs/jenkins-specifics.md#jenkinsfile-runner-and-kubernetes-plugin).

Let me know if you prefer a Pull Request to this PR again.

oleg-nenashev:
Yes, it is what I meant. Basically we can move the entire piece of
documentation to  https://github.com/jenkinsci/jenkinsfile-runner
And we can have a reduced version in the remoting repo which does not focus
K8s specifically

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -168,6 +206,42 @@ public void setDisableHttpsCertValidation(boolean disableHttpsCertValidation) {
     @CheckForNull
     public JnlpAgentEndpoint resolve() throws IOException {
         IOException firstError = null;
+
+        if (disableHttpEndpointCheck) {
```
jeffret-b: I find it confusing the way this new logic is inserted into this existing method, but is completely different. Especially because this is already an overly long method. At a minimum it would be nice to separate the new logic out into a separate method.

jglick:
FYI @Vlatombe, with whom I have discussed enhancements in this area.

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -217,7 +242,7 @@ public static void _main(String[] args) throws IOException, InterruptedException
         p.parseArgument(args);
         if(m.args.size()!=2)
             throw new CmdLineException("two arguments required, but got "+m.args);
-        if(m.urls.isEmpty())
+        if(m.urls.isEmpty() && !m.disableHttpEndpointCheck)
             throw new CmdLineException("At least one -url option is required.");
```
Vlatombe: if `disableHttpEndpointCheck` is true, it should complain if `instanceIdentity` is not provided. Otherwise it issues the following message (when running the sample command given in the PR description)

```
GRAVE: X-Instance-Identity parameter seems to be invalid
java.io.IOException: X-Instance-Identity parameter seems to be invalid
	at org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver.resolve(JnlpAgentEndpointResolver.java:234)
	at hudson.remoting.Engine.innerRun(Engine.java:574)
	at hudson.remoting.Engine.run(Engine.java:523)
```

which is cryptic if you don't know it has to be provided through command line.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -362,6 +443,11 @@ private URL toAgentListenerURL(@Nonnull String jenkinsUrl) throws MalformedURLEx
     }
 
     public void waitForReady() throws InterruptedException {
+        if (disableHttpEndpointCheck) {
+            LOGGER.log(Level.INFO, "HTTP Endpoint Check is disabled, assuming that the target master is ready");
```
Vlatombe: Maybe add a sleep here to be consistent with the logic below. If the master is restarting, it's not going to be available immediately and we'd better avoid all agents from hammering the TCP port.

alxsap:
In #337 you find proposed changes for some of the review comments. Again, a [pull request](https://github.com/jenkinsci/remoting/pull/337) on top of this pull requests due to missing other options.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -139,11 +139,28 @@ public Thread newThread(final Runnable r) {
...
+    private String instanceIdentity;
 
     private boolean disableHttpsCertValidation;
 
     private boolean noReconnect;
 
+    /**
+     * {@code true} if the target master has no Tcp Agent Listener endpoint
+     */
+    private boolean disableHttpEndpointCheck;
```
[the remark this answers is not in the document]
alxsap: see #337

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
alxsap: see #337 (minimum option)

src/main/java/hudson/remoting/jnlp/Main.java
alxsap: see #337

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
alxsap: How long should we sleep to be sure the master is up and running - without losing too much time unnecessarily?

alxsap: What about checking the JNLP port instead?

Vlatombe: I'd wait 10 seconds, as the loop below. I think in any case if the connection fails it goes back here, right?

alxsap: You're right as far as I can see.
see [#337](https://github.com/jenkinsci/remoting/pull/337/files#diff-b72206907388927fa66809bdb849f86eR452)

alxsap:
> Can we figure out some better argument names / usage? We've got two new arguments, instanceIdentity and noHttpEndpoint, but I don't see that they're really separate. If noHttpEndpoint is provided, then instanceIdentity is required. If instanceIdentity is provided, then it is ignored unless noHttpEndpoint is provided. What we really have is one piece of information we need, the instanceIdentity -- if that is provided then we need to also have any other information required to complete the connection. I'm not sure what to call it, though. Maybe say that if instanceIdentity is provided that means it won't try to connect to the httpEndpoint.
> https://github.com/jenkinsci/remoting/pull/286#pullrequestreview-268070276

@jeffret-b I agree (from my perspective) it only works if both parameters are set and it is somehow redundant. But would it be more intuitive to activate this "direct" mode by only providing the `instanceIdentity` parameter with this long, cryptic value? 🤔

> One of the checks is failing. Is that because of the dockerfile that was once there but has been removed?

That was my thought as well, what should be built without Dockerfile... guess this voter is obsolete.

alxsap:
@oleg-nenashev @Vlatombe @jeffret-b see my [comments](https://github.com/jenkinsci/remoting/pull/286#issuecomment-517243389) above and [proposed changes](https://github.com/jenkinsci/remoting/pull/337).

Any chance to continue reviewing/working on this pull request here and finally  get the changes released? We would like to get rid of our forks.

Thanks a lot, Alex

jeffret-b:
I'm taking a look to see if I can pull the different ideas around the simple CLI approach together.

jeffret-b:
Superseded by #338.

PR #288, opened 2018-09-13
Downgrade error messages from SynchronousCommandTransport

jglick:
Observing a lot of these in Evergreen coming from the `docker` plugin. Not actionable by developers (probably).


daniel-beck:
> Not actionable by developers

This I do not understand. Warnings are shown to admins, and not all warnings are actionable to plugin developers. Could just be shitty infrastructure. But hiding them by lowering the severity seems unhelpful. What am I missing?

jglick:
For Evergreen’s Sentry we would like to suppress `WARNING`s unless there is a demonstrable problem with Jenkins software that we could fix. Here we were seeing things like socket closures, which are noteworthy if you are checking whether there are some issues in your environment, but not something we currently expect a Jenkins update to address.

jglick:
Discussed in person. I am still +0.5 on this: exceptions of the type relevant in this PR may or may not even be noteworthy to the instance admin, so `WARNING` seems excessive.

jeffret-b:
Agreed with Jesse: I'm about +0.5 on this. Unless there are other concerns I'd like to merge this in and get a release out soon.

PR #289, opened 2018-09-19
[JENKINS-53569] Remove unnecessary locking.

jeffret-b:
See [JENKINS-53569](https://issues.jenkins-ci.org/browse/JENKINS-53569) for a description of the deadlock, including stack traces of the deadlocked threads.

There is no indication that this section of code constitutes a critical section. The only meaningful thing it does is atomic and is handled elsewhere in the code by good design. The extra checks serve no significant purpose.

The code in Filter that calls this grabs a lock on Filter before it calls this meaning that it already holds one lock. If another thread grabs the lock on ProtocolStack before the Filter than a thread is always possible. Locking occurs too frequently without a strategy for preventing deadlock.

Locking is complex enough without abusing locks in weird ways.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
@@ -597,7 +597,6 @@ public class ProtocolStack<T> implements Closeable, ByteBufferPool {
         /**
          * Flag to track this {@link ProtocolLayer} as removed from the stack.
          */
-        @GuardedBy("ProtocolStack.stackLock")
         private boolean removed;
 
         /**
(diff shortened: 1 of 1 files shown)


jeffret-b:
@stephenc Are there some specific load tests? I'm not familiar with any yet.

stephenc:
https://github.com/jenkinsci/remoting/blob/master/src/assembly/agent-load-test.xml Creates the load test client... there are docs *somewhere* in the repo... or look at the PR where I merged JNLPv4 as I reported the commands I used there (some years ago... I forgot details)

Try both single thread and many thread for different packet sizes

jeffret-b:
I don't know what to look for, but here's the output of the suggested different runs for that agent-load-test. Anything meaningful in there we should be looking at?
[remoting-stress.txt](https://github.com/jenkinsci/remoting/files/2398607/remoting-stress.txt)

jeffret-b:
@stephenc, do you have any further comments or concerns after I attached the testing results?

jeffret-b:
I'll plan on merging this on Friday unless there are additional concerns.

PR #293, opened 2018-10-24
[JENKINS-48778] Re-work the implementation of the no_proxy capability.

jeffret-b:
See [JENKINS-48778](https://issues.jenkins-ci.org/browse/JENKINS-48778)

We've had several issues submitted against the no_proxy implementation in Remoting. We resolved one a couple of months ago. Implementing that one turned out to be harder and less clean than it needed to be. As I started working on this one, I realized the area could use some restructuring, improved capabilities, and better documentation.

Unfortunately, there isn't a single standard or reference on how no_proxy should behave. Implementations and rules vary by application or system. I modeled this proposal after [this](https://github.com/golang/go/issues/16704#issuecomment-312502347) well-researched proposal for Go. There are a few differences.

This new version should retain all existing capabilities. If something worked before to specify a no_proxy system, that should continue working. This introduces a few new formats and rules, such as subnets and compressed IPv6. It also relaxes a correctness check that doesn't need to be performed at this level.

@reviewbybees

### DIFF
--- a/docs/no_proxy.md
+++ b/docs/no_proxy.md
@@ -1,0 +1,28 @@
+# NO_PROXY Environment Variable
+
+This optional setting provides a way to specify one or more hosts where the connection should not be proxied.
+Similar settings are available in many applications. Unfortunately, there is no standard specification across the different implementations.
+Each one defines its own syntax and supported capabilities.
+
+Remoting provides an implementation that is similar to a number of other applications and systems. Its simple rules provide
+sufficient power and flexibility for many needs.
+
+Rules:
+1. Environment variable may be named `NO_PROXY` or `no_proxy`.
+1. Different specification pieces may be separated by a comma (`,`) or a pipe (`|`).
+1. Each piece may specify an IP address, a network, or a host / domain.
+1. An IP address may be IPv4 or IPv6.
+1. An IPv6 address may be expressed in full or compressed form.
+1. A network is expressed in CIDR notation, for example, `192.168.17.0/24`.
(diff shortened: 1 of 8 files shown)


docs/no_proxy.md
```diff
@@ -0,0 +1,28 @@
...
+1. Environment variable may be named `NO_PROXY` or `no_proxy`.
+1. Different specification pieces may be separated by a comma (`,`) or a pipe (`|`).
+1. Each piece may specify an IP address, a network, or a host / domain.
+1. An IP address may be IPv4 or IPv6.
+1. An IPv6 address may be expressed in full or compressed form.
+1. A network is expressed in CIDR notation, for example, `192.168.17.0/24`.
+1. Localhost and loopback addresses are not proxied by default.
+1. All subdomains matching a domain or host are not proxied.
+1.1. A `NO_PROXY` setting of `jenkins.io` matches `repo.jenkins.io`, `sub.sub.jenkins.io`, and `jenkins.io`. It also matches `myjenkins.io`.
+1.1. The following forms are identical: `jenkins.io`, `.jenkins.io`, and `*.jenkins.io`.
```
jvz: Wait, does that mean `.jenkins.io` matches `myjenkins.io`?

jeffret-b: Yes, that's how this version works. As indicated in lines 19 and 20. And as you noticed. That wasn't entirely intentional but I thought it is probably good enough. There aren't good standard references as to how this should behave and I'm not sure how other systems handle this situation.

PR #310, opened 2018-12-06
[DO NOT MERGE] test updated parent

batmat:
Testing https://github.com/jenkinsci/pom/pull/37


batmat:
Seems like the failure is real, but no idea if this could be related to this update. I would think not, but we'll need to be sure obviously.

batmat:
So, now green. Was probably an infra issue. There has been quite a lot of them these days apparently.

PR #312, opened 2018-12-07
[JENKINS-48778] Refactoring earlier unreleased implementation.

jeffret-b:
Tweak the implementation to not draw in new dependencies.

For the changes to the no_proxy capability, I added a couple of dependencies for checking and handling ip addresses and subnets.
One of them was clearly the wrong one to use when I started integrating.
Rather, than add dependencies, we decided to copy in utilities. This same pattern was followed recently in Jenkins core to avoid introducing new dependencies into Jenkins core that plugins may become dependent on.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,16 +191,6 @@ THE SOFTWARE.
       <artifactId>test-annotations</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
-      <version>4.5.6</version>
-    </dependency>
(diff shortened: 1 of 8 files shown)


jeffret-b:
> Not sure about including the test from the library.

I think that having a good set of tests covering any code is a good thing. Sure, we've got tests that cover the specific pieces that we needed, but now that we've included these pieces, they could be used for other purposes or other parts of these classes could be used. If we're going to use the code, it's good to also make use of the available tests.

The tests have now passed. There appears to be some issues in the CI environment causing the earlier failures. We're not even seeing any consistently flaky tests.

@Wadeck, do you have any other concerns that need fixing?

PR #318, opened 2019-03-18
[JENKINS-50095] Check for Remoting minimum version when connecting to master

longngn:
See [JENKINS-50095](https://issues.jenkins-ci.org/browse/JENKINS-50095).

Remoting agent will also check for minimum supported version when resolving to `jenkinsHost/tcpSlaveAgentListener`

This PR is related to another PR on Core's side: https://github.com/jenkinsci/jenkins/pull/3938

### Proposed changelog entries

* Entry 1: Agent will refuse connection if agent's version is older than minimum supported version ([issue 50095](https://issues.jenkins-ci.org/browse/JENKINS-50095))

### Desired reviewers

@oleg-nenashev @jeffret-b


src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -202,6 +204,18 @@ public JnlpAgentEndpoint resolve() throws IOException {
                             salURL + " is invalid: " + con.getResponseCode() + " " + con.getResponseMessage()));
                     continue;
                 }
+
+                // Check if current version of agent is supported
+                String minimumSupportedVersionHeader = first(header(con, "X-Remoting-Minimum-Version"));
```
oleg-nenashev: Header might be missing while connecting to an older Jenkins version. It needs to be handled here

oleg-nenashev:
CC @batmat who is a maintainer of the Remoting Version Column Plugin

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
longngn: >Header might be missing while connecting to an older Jenkins version. It needs to be handled here

Oops I did forget that bit, thanks for telling me!
> Here I would recommend to consider just copying the relevant code into a restricted utility class

I see, then I'll copy the relevant code as well as testcases to remoting's source
>In some cases Jenkins admins may want to ignore the old version warning, e.g. if they use packaged client versions they cannot easily update. Maybe it makes sense to introduce a system property or a CLI argument which allows to opt-out from the check (so that there is a warning instead of an error)

One interesting thing I found out is that Jenkins master only displays a warning for deprecated agent version ([code](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/SlaveComputer.java#L630)). Maybe we can do the same on agent side, i.e. warning instead of error. I think that introducing a new property/argument for backward compatibility check sounds a tad cumbersome though. And if we use warning we can have the same standard as master's side.

Please share with me what you think, thank you for your time.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+ * </pre>
+ *
+ * This class is re-implemented in 1.415. The class was originally introduced in 1.139
+ *
+ * @since 1.139
+ * @author Stephen Connolly (stephenc@apache.org)
+ * @author Kenney Westerhof (kenney@apache.org)
+ * @author Hervé Boutemy (hboutemy@apache.org)
+ */
+public class VersionNumber implements Comparable<VersionNumber> {
```
jeffret-b: This class should be restricted from external use, since we really don't want to provide it as part of the API for Remoting. In other words, add the annotation, `@Restricted(NoExternalUse.class)` to the class.

Also, could you leave a comment about where this was copied from?

src/test/java/org/jenkinsci/remoting/util/VersionNumberTest.java
```diff
@@ -0,0 +1,100 @@
...
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.remoting.util;
+
+import junit.framework.TestCase;
+
+/**
+ * @author Xavier Le Vourch
+ */
+public class VersionNumberTest extends TestCase {
```
jeffret-b: And a comment here about where this class was copied from, also.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        int INTEGER_ITEM = 0;
+
+        int STRING_ITEM = 1;
+
+        int LIST_ITEM = 2;
+
+        int WILDCARD_ITEM = 3;
+
+        // Cannot set name to compareTo because FindBugs will think this interface extend Comparable and fire CO_ABSTRACT_SELF bug
+        int compare(Item item);
```
jvz: Why not make `Item` extend `Comparable<Item>` and implement as usual?

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+            return "*";
+        }
+    }
+
+    /**
+     * Represents a numeric item in the version item list.
+     */
+    private static class IntegerItem
+            implements Item {
+        private static final BigInteger BigInteger_ZERO = new BigInteger("0");
```
jvz: No need for this constant. You can refer to [`BigInteger.ZERO`](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#ZERO).

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        public String toString() {
+            return value.toString();
+        }
+    }
+
+    /**
+     * Represents a string in the version item list, usually a qualifier.
+     */
+    private static class StringItem implements Item {
+        private final static String[] QUALIFIERS = {"snapshot", "alpha", "beta", "milestone", "rc", "", "sp"};
```
jvz: Typical sort order in Java is `private static final`

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+    private String value;
+
+    private String snapshot;
+
+    private String canonical;
+
+    private ListItem items;
+
+    private interface Item {
+        int INTEGER_ITEM = 0;
```
jvz: These integer constants should probably be an enum instead.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+
+    /**
+     * Represents a string in the version item list, usually a qualifier.
+     */
+    private static class StringItem implements Item {
+        private final static String[] QUALIFIERS = {"snapshot", "alpha", "beta", "milestone", "rc", "", "sp"};
+
+        private final static List<String> _QUALIFIERS = Arrays.asList(QUALIFIERS);
+
+        private final static Properties ALIASES = new Properties();
```
jvz:
```suggestion
        private final static Map<String, String> ALIASES = new HashMap<>();
```

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        public String toString() {
+            return value;
+        }
+    }
+
+    /**
+     * Represents a version list item. This class is used both for the global item list and for sub-lists (which start
+     * with '-(number)' in the version specification).
+     */
+    private static class ListItem extends ArrayList<Item> implements Item {
```
jvz: I would not suggest extending collection classes. You can just delegate to an instance of one.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        Matcher matcher = SNAPSHOT.matcher(version);
+        if (matcher.matches()) {
+            snapshot = matcher.group(1);
+            version = version.substring(0, matcher.start(1)) + "-SNAPSHOT";
+        }
+        version = version.toLowerCase(Locale.ENGLISH);
+
+        ListItem list = items;
+
+        Stack<Item> stack = new Stack<Item>();
```
jvz: Better to use

```suggestion
        Deque<Item> stack = new ArrayDeque<>();
```

And then you can still push() and pop() as a stack. Stack is deprecated in Java.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+                    if ((i + 1 < version.length()) && Character.isDigit(version.charAt(i + 1))) {
+                        // new ListItem only if previous were digits and new char is a digit,
+                        // ie need to differentiate only 1.1 from 1-1
+                        list.add(list = new ListItem());
+
+                        stack.push(list);
+                    }
+                }
+            } else if (c == '*') {
+                list.add(new WildCardItem());
```
jvz: Can this `WildCartItem` instance be a constant value?

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+        if (snapshot == null) {
+            return o.snapshot == null ? 0 : -1;
+        }
+        if (o.snapshot == null) {
+            return 1;
+        }
+        if ("-SNAPSHOT".equals(snapshot) || "-SNAPSHOT".equals(o.snapshot)) {
+            // cannot compare literal with timestamped.
+            return 0;
```
jvz: So they're considered equal then? Maybe this should throw an `IllegalArgumentException` instead.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+        result = snapshot.substring(1, 16).compareTo(o.snapshot.substring(1, 16));
```
jvz: Why is this limited to only 16?

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+     */
+    public int getDigitAt(int idx) {
+        if (idx < 0) {
+            return -1;
+        }
+
+        Iterator it = items.iterator();
+        int i = 0;
+        Item item = null;
+        while (i <= idx && it.hasNext()) {
```
jvz: Should use a for each loop instead of an iterator here.

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+            switch (item.getType()) {
+                case INTEGER_ITEM:
+                    return -1; // 1-1 < 1.0.x
+
+                case STRING_ITEM:
+                    return 1; // 1-1 > 1-sp
+
+                case LIST_ITEM:
+                    if (item instanceof ListItem) {
+                        Iterator left = iterator();
```
jvz:
```suggestion
                        Iterator<Item> left = iterator();
```

src/main/java/org/jenkinsci/remoting/util/VersionNumber.java
```diff
@@ -0,0 +1,579 @@
...
+                        Iterator right = ((ListItem) item).iterator();
```
jvz:
```suggestion
                        Iterator<Item> right = ((ListItem) item).iterator();
```

jeffret-b:
@jvz, I think the VersionNumber and VersionNumberTest classes are copied from somewhere else. Ideally we would want to reference them and include their dependencies, but we purposefully try to keep the Remoting library small and not introduce new dependencies.

If that's the case, maybe we should just accept the copies as they are, without making those changes. That way they would be more consistent with their source in case the source updates.

I'm not sure, though. Maybe it would be better to make the corrections you suggest to update it.

Thoughts anyone?

jvz:
I'd rather see dependency classes automatically copied and renamed via a maven plugin rather than manually committed to git.

jeffret-b:
That sounds like a good idea. A downside in that case is that then I don't know that we can readily mark it as `@Restricted(NoExternalUse.class)`.

jvz:
Maybe apply the annotation to a `package-info.java` where the destination package for rewriting the dependency classes goes?

jeffret-b:
Do you think the access-modifier-annotation checker would process that correctly?

jvz:
I don't know offhand, but from what I've looked at the library, it shouldn't be too hard to support if it doesn't already.

oleg-nenashev:
I will try to re-review it this week

oleg-nenashev:
> One interesting thing I found out is that Jenkins master only displays a warning for deprecated agent version (code). Maybe we can do the same on agent side, i.e. warning instead of error. I think that introducing a new property/argument for backward compatibility check sounds a tad cumbersome though. And if we use warning we can have the same standard as master's side.

I would suggest keeping the fail-fast logic, same as @jeffret-b

oleg-nenashev:
We discussed this PR at the GSoC Status call, and we agreed to try using the Maven Shade Plugin to minimize the patch

longngn:
I used Maven Shade to rename VersionNumber package name but FindBugs just fired off the same old errors again. Seems like this is problems from inside VersionNumber's implementation itself.

pom.xml
```diff
@@ -161,6 +161,11 @@ THE SOFTWARE.
       <artifactId>constant-pool-scanner</artifactId>
       <version>1.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.6</version>
```
oleg-nenashev:
```suggestion
      <version>1.6</version>
      <scope>compile</scope>
```

jeffret-b:
I like where this is going. It looks like you're probably going to add a couple more exclusion definitions to the findbugs filter file at src/findbugs/excludeFilter.xml. They should match the specific file and issue.

longngn:
It has passed FindBugs now but failed at some error I don't understand. I would really appreciate if anyone can help me decipher the logs.

afalko:
@longngn I think it might be an infrastructure error; I think we need to restart the build somehow. 
```
[2019-05-27T15:37:14.619Z] Also:   hudson.model.Computer$TerminationRequest: Termination requested at Mon May 27 15:35:43 GMT 2019 by Thread[Ping thread for channel hudson.remoting.Channel@772e3139:JNLP4-connect connection from 20.186.111.120/20.186.111.120:49188,5,main] [id=325187]
```

longngn:
@jeffret-b So Oleg restarted the build and it has passed. Is this PR ready to be merged now, or is something need to be changed?

jeffret-b:
After merging this into master I discovered it causes a compilation failure when incorporated into Jenkins core. It must be something about the version-number dependency and how it's being pulled in. I'll probably revert it tomorrow and continue with some of the other changes.

PR #320, opened 2019-04-22
Incorrect Javadoc for Main.args

jglick:
See size check below, and the indexing inside `createEngine`. Not sure how long this Javadoc has been wrong. Perhaps it would be better to use

```java
@Argument(index = 0) public String secretKey;
@Argument(index = 1) public String agentName;
```

but I will leave that for another day.

### DIFF
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -178,8 +178,7 @@ public class Main {
 
 
     /**
-     * 4 mandatory parameters.
-     * Host name (deprecated), Jenkins URL, secret key, and agent name.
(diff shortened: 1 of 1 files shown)


jglick:
> This is for the launch-from-browser capability and not the launch from command-line.

This code path is used in all modes of inbound agents. Most—not all—users start with `Launcher`, but that anyway delegates to `Main` which then does its own argument parsing.

jeffret-b:
> This code path is used in all modes of inbound agents. Most—not all¹—users start with `Launcher`, but that anyway delegates to `Main` which then does its own argument parsing.

Yes, that's correct. I was misreading what exactly the comment refers to.

jeffret-b:
Trying to kick the build on this again.

PR #321, opened 2019-04-22
[JENKINS-57146] Add support for loading -secret value from file

jeffret-b:
Add another mechanism for obtaining the connection secret for agent-initiated, inbound connection mechanisms.

This augments the "-secret" option. If the value starts with a '@', then it provides a file path to read. With the "@", read the secret from the file given by the argument value. Without it, the secret is the argument.

Writing tests for this type of capability is difficult, as indicated by the lack of other, existing tests in this area. A meaningful test would involve significant setup, would need to be implemented in a different test harness, and seems like more effort than the gains justify. Instead, I have performed various manual testing steps on multiple platforms.

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -473,17 +473,12 @@ public class Launcher {
      * Parses the connection arguments from JNLP file given in the URL.
      */
     public List<String> parseJnlpArguments() throws ParserConfigurationException, SAXException, IOException, InterruptedException {
+        checkTooManySecrets();
         if (secretFile != null) {
-            if (slaveJnlpCredentials != null || secret != null) {
-                throw new IOException("-jnlpCredentials, -secret, and -secretFile are mutually exclusive");
-            }
             secret = readSecretFromFile(secretFile);
         }
         if (secret != null) {
             slaveJnlpURL = new URL(slaveJnlpURL + "?encrypt=true");
-            if (slaveJnlpCredentials != null) {
-                throw new IOException("-jnlpCredentials and -secret are mutually exclusive");
-            }
         }
         while (true) {
             URLConnection con = null;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/Launcher.java
```diff
@@ -465,6 +473,12 @@ private SSLSocketFactory getSSLSocketFactory()
      * Parses the connection arguments from JNLP file given in the URL.
      */
     public List<String> parseJnlpArguments() throws ParserConfigurationException, SAXException, IOException, InterruptedException {
+        if (secretFile != null) {
+            if (slaveJnlpCredentials != null || secret != null) {
+                throw new IOException("-jnlpCredentials, -secret, and -secretFile are mutually exclusive");
```
daniel-beck: also needs to update line 485:

    throw new IOException("-jnlpCredentials and -secret are mutually exclusive");

Perhaps determine separately how many of these are set and fail there rather that repeat checks?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -568,6 +582,23 @@ private SSLSocketFactory getSSLSocketFactory()
         }
     }
 
+    private String readSecretFromFile(String secretFile) throws IOException {
+        BufferedReader reader = null;
+        String line = null;
+        try {
```
fcojfernandez: What about using a try-with-resources? Is it not possible for any reason?

jeffret-b: Yes, that would be better.

src/main/java/hudson/remoting/Launcher.java
jeffret-b: Yeah, I thought about that, but I figured it was easier to add and about as easy to read by just adding the one check. If secretFile is null then the message in 485 is still valid.

jeffret-b:
> It is unclear to me what makes `-secret` special compared to `-jnlpCredentials` or `-proxyCredentials`. Wouldn't they have the same problem? And if so, wouldn't a common solution be preferable over adding `-jnlpCredentialsFile` and `-proxyCredentialsFile` once someone asks for that?

It's a valid question. I'm really not sure what makes `-secret` special, but we have a specific request for that. I question how much it really accomplishes, but it is important to some users. I'm not sure it's worth expanding the scope of this change to provide a more thorough solution.

src/main/java/hudson/remoting/Launcher.java
daniel-beck: Well, it should mention `secretFile` though.

jeffret-b:
Implemented it a little differently, at Daniel's suggestion.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -568,6 +569,20 @@ private SSLSocketFactory getSSLSocketFactory()
         }
     }
 
+    private boolean parameterReferencesFile(String parameter) {
+        return parameter.charAt(0) == '@';
```
fcojfernandez: is there any chance of getting a false positive here?

jvz: No, I don't believe so. Secrets are encoded in hexadecimal (AES 128b secret key).

jeffret-b:
Doesn't look like this needed after all as there are already sufficient capabilities built into args4j to mee the use case. I'll close this unless for some reason we need something different than what exists.

PR #323, opened 2019-06-03
Addition to PR 286: Add Instance Identity Key to Support JNLP4

alxsap:
This pull request adds changes on top of https://github.com/jenkinsci/remoting/pull/286. The changes here implement the missing
> support of Instance Identity key so that JNLP4 protocol can be used

Connecting to a Jenkins without http port now already works. You can test it with the command below. 

```sh
java -cp slave.jar hudson.remoting.jnlp.Main \
  -headless -noHttpEndpoint \
  -tunnel <jenkinsHost>:<jnlpPort> \
  -instanceIdentity <instanceIdentityString> \
  <secretString> <agentName>
```

- The agent needs to be prepared on the master  - copy the name and secret.  
- The instanceIdentityString is the base64 encoded byte array of the masters Instance Identity. You can get it via `hudson.remoting.Base64.encode(org.jenkinsci.main.modules.instance_identity.InstanceIdentity.get().getPublic().getEncoded())`.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -141,8 +141,13 @@ public class Engine extends Thread {
     private String tunnel;
     
     /**
-     * The base64 encoded byte array of the Instance Identities public key of the Jenkins master. 
-     * See https://wiki.jenkins.io/display/JENKINS/Instance+Identity
+     * The base64 encoded byte array of the <a href="https://wiki.jenkins.io/display/JENKINS/Instance+Identity">Instance Identities</a> public key of the Jenkins master.<br>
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
It would have been better if the changes were integrated in a single PR, but it would do the job as well

alxsap:
@oleg-nenashev What do you mean by `if the changes were integrated in a single PR`? I thought such a PR with target your PR branch is what you proposed? If I shall change something let me know...

oleg-nenashev:
🤦‍♂ , my fault

oleg-nenashev:
Merging towards the target branch so that we do not longer have splitbrain

PR #327, opened 2019-06-11
Switch to SpotBugs

basil:
Updated the parent POM to [the latest version](https://github.com/jenkinsci/pom/releases), which has SpotBugs support, and migrated the annotations dependency from FindBugs to SpotBugs. This exposed the following new SpotBugs error:

```
[ERROR] Dead store to $L4 in hudson.remoting.UserRequest.perform(Channel) [hudson.remoting.UserRequest] At UserRequest.java:[line 171] DLS_DEAD_LOCAL_STORE
```

I fixed this by removing the unnecessary variable declaration and assignment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.50</version>
+    <version>1.52</version>
     <relativePath />
   </parent>
 
@@ -162,9 +162,9 @@ THE SOFTWARE.
       <version>1.2</version>
     </dependency>
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>annotations</artifactId>
-      <version>3.0.1</version>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <version>3.1.12</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -164,8 +164,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>com.github.spotbugs</groupId>
       <artifactId>spotbugs-annotations</artifactId>
-      <version>3.1.12</version>
-      <scope>provided</scope>
+      <optional>true</optional>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -162,9 +162,9 @@ THE SOFTWARE.
       <version>1.2</version>
     </dependency>
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>annotations</artifactId>
-      <version>3.0.1</version>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <version>3.1.12</version>
```
oleg-nenashev: Version could be added to `dependencyManagement` in Parent POM

basil: > Version could be added to `dependencyManagement` in Parent POM

Agreed. That seems like a pre-existing issue, though, and not something related to the switch to SpotBugs.

basil: > Version could be added to `dependencyManagement` in Parent POM

The version _already is_ present in a `dependencyManagement` section in the parent POM; see [here](https://github.com/jenkinsci/pom/commit/f15580982fe7cc7f486483bee01108d909940292#diff-600376dffeb79835ede4a0b285078036R110). So I removed the version from this POM, and I also changed `<scope>provided</scope>` to `<optional>true</optional>` as described in [the SpotBugs documentation](https://spotbugs.readthedocs.io/en/stable/migration.html).

basil:
@oleg-nenashev Any thoughts on the use of `<scope>provided</scope>` vs `<optional>true</optional>` for declaring the dependency on the SpotBugs annotations? I agree that we shouldn't have a version number in downstream POMs, since the version is specified in a `dependencyManagement` section upstream. But I'm not sure whether to use `<scope>provided</scope>` in downstream POMs (as is currently done in Remoting and Jenkins) versus `<optional>true</optional>` (as described in [the SpotBugs documentation](https://spotbugs.readthedocs.io/en/stable/migration.html)). I'm inclined to go with the latter, since that seems to be the recommendation in the SpotBugs documentation. But I'm not sure if there's some specific reason the Jenkins project has been using the former.

jeffret-b:
Tried kicking off a build. Hopefully that will pass.

jeffret-b:
Passed this time, so it should be ready to go.

PR #331, opened 2019-06-13
[JENKINS-50095] Fix version number issues in Jenkins core

jeffret-b:
See #318. After I merged it into master, I discovered that Jenkins core won't build with this updated Remoting. The failure gives no information. I have tried various changes to get it to build but the only thing I've gotten to work is to remove the shading.

I should have squash-merged #318 and then I could just revert it, but since I didn't reverting is more complicated. I could still continue down that route, or use this proposal, which builds fine. I'm inclined to go ahead and move forward by merging this change in.

I haven't changed or updated `VersionNumber` and `VersionNumberTest` except to add a comment indicating where we copied them from and why.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,11 +162,6 @@ THE SOFTWARE.
       <version>1.2</version>
     </dependency>
     <dependency>
-      <groupId>org.jenkins-ci</groupId>
-      <artifactId>version-number</artifactId>
-      <version>1.6</version>
-    </dependency>
-    <dependency>
       <groupId>com.github.spotbugs</groupId>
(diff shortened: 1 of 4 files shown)


oleg-nenashev:
@jeffret-b Hi. You may want to recover https://github.com/jenkinsci/remoting/pull/265 which got prematurely merged and then reverted. It could have prevented such issues

jeffret-b:
> You may want to recover #265 which got prematurely merged and then reverted. It could have prevented such issues

If someone could ever get enough time to complete it, it would be great to have those additional checks. I've thought about it a few times but have trouble eking out enough time for it.

PR #338, opened 2019-08-13
[JENKINS-53461]  Direct inbound TCP agent connection

jeffret-b:
See [JENKINS-53461](https://issues.jenkins-ci.org/browse/JENKINS-53461).

An alternative to #286. I've pulled together many of the suggestions and ideas from the earlier proposal and put together something complete with documentation. I want to isolate the new pieces as much as can be easily done for clarity and proper separation.

I added documentation about some inbound TCP agent launch mechanisms. I needed to be able to document the new stuff, but there isn't a good location documenting the old stuff. I put together a few simple explanations of existing mechanisms and then added discussion of this new mechanism.

To understand what this PR does and how to use it, read https://github.com/jeffret-b/remoting/blob/866ac08ad98ddbe00f5aa179a13bce3ab6657ea3/docs/tcpAgent.md#connect-directly-to-tcp-port

This mechanism is intended to facilitate a single, fairly narrow scenario.  There are other, more significant approaches that could also be taken that might work better in other scenarios, particularly kubernetes or other cloud-based configurations. More work would need to be done to clearly specify those needs and provide the most productive solution. This PR makes small changes to enable a simple scenario.


docs/tcpAgent.md
```diff
@@ -0,0 +1,97 @@
+# Launching inbound TCP agents
```
jeffret-b: Document some existing launch mechanisms. And then add the new one.

docs/tcpAgent.md
```diff
@@ -0,0 +1,97 @@
...
+
+A typical example of this command, as shown in the above image, is
+``` 
+java -jar agent.jar \
+  -jnlpUrl <jnlp url> \
+  -secret <secret key> \
+  -workDir <work directory> 
+```
+
+### Connect directly to TCP port
```
jeffret-b: Here's the documentation for the new mechanism.

docs/tcpAgent.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Parameters
+
+There are a number of different launch parameters that control how the agent connects and behaves. The parameters available and the default behavior may vary depending upon the entry point. You can obtain usage information by executing `java -cp agent.jar hudson.remoting.jnlp.Main` or `java -jar agent.jar --help`. Not all parameters work together and some parameters require the use of others.
+
+There are also system or environment variables that control some advanded behaviors documented at [Remoting Configuration](https://github.com/jenkinsci/remoting/blob/master/docs/configuration.md). Many of these need to be set at both master and agent. Changing some of these may result in unreliable behavior. One of these parameters that is supported with defined behavior is [`NO_PROXY`](https://github.com/jenkinsci/remoting/blob/master/docs/no_proxy.md). Similar to the usage in a number of other applications, this controls which hosts should be excluded from proxy configurations.
+
+### The '@' argument annotation
```
jeffret-b: And also add some documentation on the '@' argument annotation that has been missing.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -140,9 +143,9 @@ public Thread newThread(final Runnable r) {
     @CheckForNull
     private String tunnel;
 
-    private boolean disableHttpsCertValidation;
+    private boolean disableHttpsCertValidation = false;
 
-    private boolean noReconnect;
+    private boolean noReconnect = false;
```
jeffret-b: Just being a little more clear with a reminder about the default value.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -678,6 +643,23 @@ public void afterChannel(@Nonnull JnlpConnectionState event) {
         }
     }
 
+    private JnlpEndpointResolver createEndpointResolver(List<String> jenkinsUrls) {
```
jeffret-b: Create either a traditional resolver that downloads the JNLP file and extracts the information from it or a new one that just builds the necessary endpoint information from parameters.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -688,7 +670,7 @@ private void onConnectionRejected(String greeting) throws InterruptedException {
      * @param endpoint Connection endpoint
      * @throws IOException Connection failure or invalid parameter specification
      */
-    private Socket connect(@Nonnull JnlpAgentEndpoint endpoint) throws IOException, InterruptedException {
+    private Socket connectTcp(@Nonnull JnlpAgentEndpoint endpoint) throws IOException, InterruptedException {
```
jeffret-b: Rename to be a little clearer on what it's doing as the name was kind of vague.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -849,4 +831,55 @@ private SSLSocketFactory getSSLSocketFactory()
      * @since 2.4
      */
     static final int SOCKET_TIMEOUT = Integer.getInteger(Engine.class.getName()+".socketTimeout",30*60*1000);
+
+    private class EngineJnlpConnectionStateListener extends JnlpConnectionStateListener {
```
jeffret-b: Extract this out to a named class to make it easier to read.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointConfigurator.java
```diff
@@ -0,0 +1,68 @@
...
+package org.jenkinsci.remoting.engine;
+
+import java.io.IOException;
+import java.security.interfaces.RSAPublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class JnlpAgentEndpointConfigurator extends JnlpEndpointResolver {
```
jeffret-b: Creates an endpoint from arguments.

jeffret-b:
@alxsap does this satisfy your needs, as described on #286?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -678,6 +643,23 @@ public void afterChannel(@Nonnull JnlpConnectionState event) {
...
     }
 
+    private JnlpEndpointResolver createEndpointResolver(List<String> jenkinsUrls) {
+        JnlpEndpointResolver resolver;
+        if (instanceIdentity == null) {
+            SSLSocketFactory sslSocketFactory = null;
+            try {
+                sslSocketFactory = getSSLSocketFactory();
+            } catch (Exception e) {
+                events.error(e);
```
jvz: Should this perhaps return `null` on error or something? Otherwise, the following `resolver = ...` line could be partially initialized which doesn't sound right to me. Or should it use the default configurator like in line 658?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -849,4 +831,55 @@ private SSLSocketFactory getSSLSocketFactory()
      * @since 2.4
      */
     static final int SOCKET_TIMEOUT = Integer.getInteger(Engine.class.getName()+".socketTimeout",30*60*1000);
+
+    private class EngineJnlpConnectionStateListener extends JnlpConnectionStateListener {
+
+        private final RSAPublicKey publicKey;
```
jvz: Leave it to me to worry about cryptography choices years before they're a problem. Does this actually need to use an RSA key, or can it be any `PublicKey` type technically? Even if that's usually RSA.

I'm not actually sure about alternative algorithms available through X509 certificates, though later versions of Java have added more support there (if BouncyCastle doesn't cover us already).

Looking further down, I see you had to cast the `PublicKey` to the appropriate subtype anyways. I don't see much in here that's specific to using RSA (besides the fact that that's the algorithm in play here).

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointConfigurator.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(JnlpAgentEndpointConfigurator.class.getName());
+
+    private String instanceIdentity;
```
jvz: Seem like these can all be final?

src/main/java/org/jenkinsci/remoting/engine/JnlpEndpointResolver.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    protected static class HostPort {
+        String host;
+        int port;
+    }
+
+    protected HostPort splitHostPort(String value) throws IOException {
+        String[] tokens = value.split(":", 3);
```
jvz: This will only work for IPv4 addresses, right? Though I see that this isn't new.

src/main/java/hudson/remoting/Engine.java
jeffret-b: It's an interesting question, but not terribly relevant for the current PR. We may want to look into something like that as an enhancement in some later PR.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointConfigurator.java
jeffret-b: Could be, but the fields in the other, existing class weren't so I didn't bother putting it here.

src/main/java/org/jenkinsci/remoting/engine/JnlpEndpointResolver.java
jeffret-b: It's a good point. Let me see if I can put together something that will work.

It's not new, but maybe not a bad time to correct a little issue like this.

src/main/java/hudson/remoting/Engine.java
jeffret-b: This particular logic hasn't changed in this PR. It's just been moved around a little. It could probably be changed to be more robust but there aren't any complaints against it so it's probably not worth the bother.

Basically, if it fails to create the custom SSL socket factory it passes null and the following steps use the default one. Not necessarily a bad choice. The flow is a little odd, though, because most likely it has already done this same operation earlier. And there isn't much of a real reason why it shouldn't be able to create the socket factory.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+Additional descriptions of configuring this mechanism are located at [Installing Jenkins as a Windows service](https://wiki.jenkins.io/display/JENKINS/Installing+Jenkins+as+a+Windows+service) or [How to Install Several Windows Slaves as a Service?](https://support.cloudbees.com/hc/en-us/articles/217423827-How-to-Install-Several-Windows-Slaves-as-a-Service-)
+
+## Parameters
+
+There are a number of different launch parameters that control how the agent connects and behaves. 
+The parameters available and the default behavior may vary depending upon the entry point. 
+You can obtain usage information by executing `java -cp agent.jar hudson.remoting.jnlp.Main` or `java -jar agent.jar --help`. 
+Not all parameters work together and some parameters require the use of others.
+
+There are also system or environment variables that control some advanded behaviors documented at [Remoting Configuration](https://github.com/jenkinsci/remoting/blob/master/docs/configuration.md).
```
alxsap:
```suggestion
There are also system or environment variables that control some advanced behaviors documented at [Remoting Configuration](https://github.com/jenkinsci/remoting/blob/master/docs/configuration.md). 
```

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+For example, more recent versions of Jenkins show in the UI that one form of launching inbound TCP agents is
+```
+echo <secret key> > secret-file
+java -jar agent.jar -jnlpUrl <jnlp url> -secret @secret-file -workDir <work directory> 
+```
+This varies from the form shown above in "Download JNLP file" in that the secret key is stored in a file rather than passed directly in the command-line. 
+
+The behavior is more general than shown on the agent status page. 
+Any argument passed to the program (not interpreted by Java) may start with an '@'. 
+The agent will intpret the rest of that argument as the name of a file, "secret-file" in this example.
```
alxsap:
```suggestion
The agent will interpret the rest of that argument as the name of a file, "secret-file" in this example. 
```

src/main/java/hudson/remoting/Engine.java
```diff
@@ -203,14 +206,27 @@ public Thread newThread(final Runnable r) {
 
     private DelegatingX509ExtendedTrustManager agentTrustManager = new DelegatingX509ExtendedTrustManager(new BlindTrustX509ExtendedTrustManager());
 
+    private String directUrl;
```
alxsap: Name it differently, since it is not a "URL" (only host:port)?
The CLI parameter is called `-direct`.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -203,14 +206,27 @@ public Thread newThread(final Runnable r) {
...
         this.listener = listener;
+        this.directUrl = directUrl;
         this.events.add(listener);
         this.candidateUrls = hudsonUrls;
         this.secretKey = secretKey;
         this.slaveName = slaveName;
-        if(candidateUrls.isEmpty())
+        this.instanceIdentity = instanceIdentity;
+        this.protocols = protocols;
+        if(candidateUrls.isEmpty() && instanceIdentity == null) {
```
alxsap: Shouldn't we also check `directUrl` and `instanceIdentity` to fail early if not set properly? Something like:
```
if(isEmpty(directUrl) ^ isEmpty(instanceIdentity)) {
        "Both 'direct' and 'instanceIdentity' need to be specified for direct connections");
} else if(candidateUrls.isEmpty() && isEmpty(directUrl)) {
...
```

src/main/java/hudson/remoting/Engine.java
```diff
@@ -678,6 +643,23 @@ public void afterChannel(@Nonnull JnlpConnectionState event) {
         }
     }
 
+    private JnlpEndpointResolver createEndpointResolver(List<String> jenkinsUrls) {
+        JnlpEndpointResolver resolver;
+        if (instanceIdentity == null) {
```
alxsap: Wouldn't it be better readable and more intuitive if `directUrl` would be used to determine the behaviour?
```suggestion
        if (isEmpty(directUrl)) {
```

src/main/java/hudson/remoting/Engine.java
jeffret-b: Take a look at how I defined the "@Options", here and in nearby lines: https://github.com/jenkinsci/remoting/pull/338/files#diff-60b1d0697cd30e065bf0996f3b1ff438R183

The "depends" and "forbids" values should handle that in a clean way. We might could tighten that up a, especially with existing arguments but I think I got it pretty good for the new ones.

I tested the behavior at one point and it did pretty much what I expected. Probably worth testing again that I got it lined up correctly.

jeffret-b:
Thanks a lot @alxsap for the review! Great catch on those typos and misnaming from an earlier iteration I tried. I look forward to hearing how this works in your tests.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
+# Launching inbound TCP agents
+Jenkins provides a number of ways of connecting remote agents.
+Two of the most popular are SSH agents and inbound TCP agents.
```
jvz:
```suggestion
Two of the most popular are outbound SSH agents and inbound TCP agents.
```

Or do we also have inbound SSH agents? Since Jenkins has some sort of ssh-server plugin.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+SSH agents, most commonly used on Unix platforms, are master initiated.
+The master creates the connection when it needs.
+Inbound TCP agents, most commonly used on Windows platforms, are client initiated.
```
jvz:
```suggestion
Inbound TCP agents, most commonly used on Windows platforms, are agent initiated.
```

Or am I misunderstanding how that works? Same for the next line.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+This document describes some of the primary mechanisms for launching inbound TCP agents.
+
+For additional information about Jenkins agents see [Distributed builds](https://wiki.jenkins.io/display/JENKINS/Distributed+builds#Distributedbuilds-HavemasterlaunchagentonWindows).
+
+## Launch mechanisms
+Part of the agent status page for TCP agents looks something like this:
+![Tcp agent status UI](tcpAgentStatus.jpg)
+This shows a couple of launch methods, but there are other mechanisms.
+If you use a different mechanism than clicking the "Launch" button, this page contains parameters that you will need. 
+The long string of hex digits is a secret key that the client needs to establish the connection.
```
jvz: Reading "secret key" here made me do some digging. While there is indeed a secret key (one generated from HmacSHA256 currently), what is actually returned and used here is a MAC of the name of the agent computer. As a budding crypto nerd, I'd be happy if you could make this sentence less ambiguous somehow.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
+# Launching inbound TCP agents
+Jenkins provides a number of ways of connecting remote agents.
+Two of the most popular are SSH agents and inbound TCP agents.
+SSH agents, most commonly used on Unix platforms, are master initiated.
+The master creates the connection when it needs.
+Inbound TCP agents, most commonly used on Windows platforms, are client initiated.
+The client must first connect to the master and then the master sends commands as needed.
```
jeffret-b:
```suggestion
The agent must first connect to the master and then the master sends commands as needed.
```

docs/tcpAgent.md
jeffret-b: I'm open to suggestions on how to re-word it. I don't think this is the place to go into details of the construction of the key, though. That could be useful in a more detailed, technical implementation document. This document is about giving users information on how to launch the agent.

jvz: It almost makes sense to call it a "secret token" rather than a "secret key", but then it seems as though the common use of the word "token" is for something revocable (like the crumbs and API keys sorta). Maybe you can simplify it to just "secret" rather than "secret key"?

jvz: Come to think of it, each secret is agent-specific, so maybe you can call it an "agent secret"? I'm almost tempted to say "agent identity", but it works completely different from the instance identity thing.

jeffret-b: I'm not as fond of using "token". It seems to commonly have a different meaning, as you describe.

"secret" might work, but though technically correct I feel it might be confusing to a number of users. It's called "secret key" a couple of places and I think that makes more sense to general users. I don't think adding the modifier "agent" onto the beginning helps the user much.

I'm not committed to the current term, but I'm not convinced the alternatives help much.

jvz: Well, as long as we aren't actually using any AES or other symmetric keys in an ambiguous context, then keeping "secret key" would be fine with me.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+If you use a different mechanism than clicking the "Launch" button, this page contains parameters that you will need. 
+The long string of hex digits is a secret key that the client needs to establish the connection. 
+Depending on how you connect you may also need the agent name, "AgentName" in this example, or the (misnamed) jnlpUrl, "https://myjenkins.example.com/computer/AgentName/slave-agent.jnlp".
+
+### "Launch" button
+This is the only launch mechanism that actually uses JNLP.
+When properly configured and supported, clicking this button causes a JNLP file to be downloaded via the web browser and launched with Java WebStart. 
+
+While this is the simplest mechanism it has many drawbacks.
+This is an outdated technology that is being phased out.
```
jvz: These three lines read like bullet points. Do you want to make them bullets?

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Connect directly to TCP port
+With the latest versions of the Remoting library (since TODO), it is possible to connect directly to the TCP port. 
+This skips the download of the connection information file, described in step 1) above. 
+To use this you must provide some of the information that the downloaded file would contain.
+
+You must provide the agent name. 
+This is easily obtained from the agent status page or other places.
+
+You need to specify the host and port, in a typical `HOST:PORT` fashion.
```
jvz: If you do manage to support IPv6 here, it might be good to note that the IPv6 address should be surrounded in square brackets just like it would in a URL.

docs/tcpAgent.md
```diff
@@ -0,0 +1,140 @@
...
+  -direct <HOST:PORT> \
+  -protocols JNLP4-connect \
+  -instanceIdentity <instance identity> \
+  <secretString> <agentName>
+```
+The "-protocols" parameter is optional, but is useful to limit the agent to protocols the server supports. 
+The only currently supported and recommended protocol is "JNLP4-connect".
+
+Note that this mechanism uses a different JAR entry point than used for "Download JNLP file". 
+The parameters available and the default behavior may be different between the entry points.
```
jvz: Do you think it would be worthwhile to note here to see the "@" parameter section later for a more secure way of launching the agent without having the secrets in the process list? Though it's not exactly far away from here, it might be useful?

src/main/java/org/jenkinsci/remoting/engine/JnlpEndpointResolver.java
```diff
@@ -0,0 +1,54 @@
...
+import hudson.remoting.Base64;
+
+import java.io.IOException;
+import java.security.KeyFactory;
+import java.security.NoSuchAlgorithmException;
+import java.security.interfaces.RSAPublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
+
+public abstract class JnlpEndpointResolver {
```
jvz: Any preference in making this an abstract class versus an interface with default methods?

src/test/java/org/jenkinsci/remoting/engine/HostPortTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Test
+    public void testIPv4() {
+        HostPort hostPort = new HostPort("1.2.3.4:5555");
+        assertThat(hostPort.getHost(), is("1.2.3.4"));
+        assertThat(hostPort.getPort(), is(5555));
+    }
+
+    @Test
+    public void testIPv6() {
+        HostPort hostPort = new HostPort("1:2::3:4:5555");
```
jvz: Might want to also support the square bracket syntax?

docs/tcpAgent.md
jeffret-b: That's a possibility, but I don't think it lays out / reads as well. It might work better in a system with richer formatting. Especially since the third point spans multiple paragraphs.

docs/tcpAgent.md
jeffret-b: Good point. I'll adjust the implementation.

docs/tcpAgent.md
jeffret-b: I don't think so. The documentation for the "@" parameters is close enough that referencing it here and in the prior subsection would just add clutter. And it's quite debatable that format is more secure.

src/main/java/org/jenkinsci/remoting/engine/JnlpEndpointResolver.java
jeffret-b: No real preference. I just thought it felt better as an abstract class.

src/main/java/hudson/remoting/Engine.java
alxsap: I saw the "depends" and "forbids" in `Main` &ndash; nice.
But if remoting is used as library `Engine` might be used directly, bypassing `Main`.

jeffret-b: True, but I don't know if it's worth duplicating / complicating the code for what is probably a rare situation. This code is already messy enough. It really needs to be reworked with better structured class responsibility and isolation, but that introduces API changes. If people think it's important to duplicate the check I can add it.

alxsap:
> Thanks a lot @alxsap for the review! Great catch on those typos and misnaming from an earlier iteration I tried. I look forward to hearing how this works in your tests.

@jeffret-b I tested your version of remoting.jar inside a modified docker-jnlp-slave image, a pipeline referencing this image in the Kubernetes plugin config, a patched Kubernetes plugin which provides the new required parameters, and all running in a Jenkinsfile Runner (on the same K8s cluster) without http port.
**The direct connect worked**, the Kubernetes plugin spawned a new pod and the plugin executed code there.

Thanks a lot

jeffret-b:
Thanks for testing that @alxsap! It's much better to proceed with an enhancement like this when we know it works correctly in at least one real configuration.

I'll leave this for a few days to see if there are any additional comments or reviews and then see about getting a release out.

jeffret-b:
Close to trigger a rebuild and see if it can get past CI now.

alxsap:
@jeffret-b thanks a lot. Do you know when the next release is planned?

oleg-nenashev:
@alxsap @jeffret-b thanks a lot for your work on this story! Sorry that I dropped the ball on it (was not even able to review PRs), my work assignments were completely different, and all spare time got consumed on GSoC

jeffret-b:
I was hoping to get the release out yesterday but ran out of time due to a few little complications. I should manage it today. Then I'll put up the PR to include the new Remoting in Jenkins. That may not get merged in until next week's release.

jeffret-b:
Released in Remoting 3.34 today. PR for inclusion into Jenkins at jenkinsci/jenkins#4165 .

PR #341, opened 2019-08-22
[JENKINS-59047] - Use base64 from java

res0nance:
Replaced existing base64 implementation with the one provided by java and deprecated the class.
Using the RFC 4648 encoder instead of the RFC 2045 encoder. The RFC 2045 encoder contains linefeeds unlike RFC 4648 which is what the current implementation does.

### DIFF
--- a/src/main/java/hudson/remoting/Base64.java
+++ b/src/main/java/hudson/remoting/Base64.java
@@ -15,6 +15,8 @@
  */
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
@res0nance Would be great to have a JIRE ticket for it. It is a pretty important change IMHO

res0nance:
@oleg-nenashev Filed one

res0nance:
Yeah the class being final allowed me to remove the remaining code and just stub the functions to the new call. I was originally paranoid that I would break backwards compatibility if I removed the protected bits but just noticed it was final.

PR #345, opened 2019-08-28
[JENKINS-59094] Jenkins connection failure with tunneling

jeffret-b:
See [JENKINS-59094](https://issues.jenkins-ci.org/browse/JENKINS-59094)

Under some conditions, agents using tunneling have failed to connect after changes in the 3.34 Remoting release. I haven't been able to reproduce it or isolate it yet, but this change restores the host:port parsing logic to previous behavior for handling an empty port. At a minimum, this should at least get us past the exception. It may well solve the problem.

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
@@ -15,13 +15,10 @@ class HostPort {
             return;
         }
         int portSeparator = hostPortValue.lastIndexOf(':');
-        if (portSeparator <= 0) {
+        if (portSeparator < 0) {
             throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
         }
         host = hostPortValue.substring(0, portSeparator).trim();
-        if (host.isEmpty()) {
-            throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
-        }
         String portString = hostPortValue.substring(portSeparator + 1).trim();
         if (portString.length() > 0) {
             port = Integer.parseInt(portString);
--- a/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
@@ -3,7 +3,7 @@ package org.jenkinsci.remoting.engine;
 class HostPort {
 
     private static final int PORT_MIN = 0;
-    private static final int PORT_MAX = (1 << 16) -1;
+    private static final int PORT_MAX = (1 << 16) - 1;
 
     private String host;
     private int port;
--- a/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
@@ -1,6 +1,10 @@
 package org.jenkinsci.remoting.engine;
 
 class HostPort {
+
+    private static final int PORT_MIN = 0;
+    private static final int PORT_MAX = (1 << 16) -1;
+
     private String host;
     private int port;
 
@@ -22,6 +26,9 @@ class HostPort {
         String portString = hostPortValue.substring(portSeparator + 1).trim();
         if (portString.length() > 0) {
             port = Integer.parseInt(portString);
+            if (port < PORT_MIN || port > PORT_MAX) {
+                throw new IllegalArgumentException("Invalid port value: " + value);
+            }
         }
     }
 
@@ -45,4 +52,5 @@ class HostPort {
     public int getPort() {
         return port;
     }
+
 }
--- a/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
@@ -22,7 +22,10 @@ class HostPort {
         if (host.isEmpty()) {
             throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
         }
-        port = Integer.parseInt(hostPortValue.substring(portSeparator + 1).trim());
+        String portString = hostPortValue.substring(portSeparator + 1).trim();
+        if (portString.length() > 0) {
+            port = Integer.parseInt(portString);
+        }
     }
 
     private void extractIPv6(String hostPortValue) {
--- a/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/HostPort.java
@@ -8,11 +8,11 @@ class HostPort {
     private String host;
     private int port;
 
-    public HostPort(String value) {
-        splitHostPort(value);
+    public HostPort(String value, String defaultHost, int defaultPort) {
+        splitHostPort(value, defaultHost, defaultPort);
     }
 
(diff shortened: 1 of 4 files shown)


bbara:
Hi!

I have a pretty similar problem but in my case the HOST is empty.

This line suggests that the tunnel string, on some point in code, is auto-completed:
https://github.com/jenkinsci/remoting/blob/58588ead47ed053ac2cd8e77d4052bba6795a5c2/src/main/java/hudson/remoting/jnlp/Main.java#L70-L74

However, this is not done until this point:
https://github.com/jenkinsci/remoting/blob/bdfcda0772a883ada2355194aa1c547906dae078/src/main/java/org/jenkinsci/remoting/engine/HostPort.java#L17-L24

I currently don't have time to investigate into it further, but at some point later in code, there must have happened some kind of auto-completion.

Unfortunately, I also haven't figured out where the tunnel is set at all, because I'm not sure where [this jelly](https://github.com/jenkinsci/jenkins/blob/jenkins-2.190/core/src/main/resources/hudson/slaves/SlaveComputer/slave-agent.jnlp.jelly) is located in the UI.

BR,
BB

bbara:
I currently have limited access to my Jenkins instance but I will provide feedback asap.

Some more findings:
[Help for the JNLP Tunnel](https://github.com/jenkinsci/jenkins/blob/jenkins-2.191/war/src/main/webapp/help/system-config/master-slave/jnlp-tunnel.html)
[Setting up the tunnel is an advanced configuration](https://github.com/jenkinsci/jenkins/blob/b342fb752d9f9cd4146e8617c75737ca0fd7dd14/core/src/main/resources/hudson/slaves/JNLPLauncher/config.jelly#L31-L38)
[Function to get the ServerName that is actually never called](https://github.com/jenkinsci/jenkins/blob/b342fb752d9f9cd4146e8617c75737ca0fd7dd14/core/src/main/java/hudson/Functions.java#L1677-L1683)

bbara:
Does not work for me:
```
Aug 29, 2019 9:12:16 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Handshaking
Aug 29, 2019 9:12:16 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connecting to :8000
Aug 29, 2019 9:12:27 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connecting to :8000 (retrying:2)
java.io.IOException: Failed to connect to :8000
        at org.jenkinsci.remoting.engine.JnlpAgentEndpoint.open(JnlpAgentEndpoint.java:243)
        at hudson.remoting.Engine.connectTcp(Engine.java:680)
        at hudson.remoting.Engine.innerRun(Engine.java:558)
        at hudson.remoting.Engine.run(Engine.java:490)
```
So for my case, it seems like the server name/HOST got lost somewhere.

However, I found the advanced setting to configure the tunnel of a node which is, in my setup, not needed anymore anyways.

For Reproduction:
As this setup worked in the last version, simply setting the tunnel of a node to the TCP agent port, which is set in the security settings of Jenkins, should be sufficient. The same should be valid for the HOST part, so simply setting the HOST to the actual server name and omit the actual port number should auto complete it to the one mentioned above.

jeffret-b:
Thanks to @bbara for the suggestion on how to configure this for testing I was able to make some progress and figure out at least some of the Jenkins magic going on here. I tweaked the PR to handle those cases. I've tested it out with no tunnel set and with it set to values like "hostname:port", "hostname:", ":port", and even ":". (The latter doesn't do anything useful that I know of, but kind of exists to support the others.)

I feel better after being able to reproduce at least some of the problems, at least in a simplified case.

Any other reviews and testing would be welcome. I'm hopeful this is ready to go.

src/main/java/org/jenkinsci/remoting/engine/HostPort.java
```diff
@@ -1,28 +1,38 @@
 package org.jenkinsci.remoting.engine;
 
 class HostPort {
+
+    private static final int PORT_MIN = 0;
+    private static final int PORT_MAX = (1 << 16) -1;
```
jvz:
```suggestion
    private static final int PORT_MAX = (1 << 16) - 1;
```

Otherwise looks like multiplication.

jeffret-b: That's a good point.

src/main/java/org/jenkinsci/remoting/engine/HostPort.java
```diff
@@ -1,28 +1,38 @@
...
-        host = hostPortValue.substring(0, portSeparator).trim();
-        if (host.isEmpty()) {
-            throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
+        String hostValue = hostPortValue.substring(0, portSeparator).trim();
+        host = hostValue.length() > 0 ? hostValue : defaultHost;
+        String portString = hostPortValue.substring(portSeparator + 1).trim();
+        if (portString.length() > 0) {
+            port = Integer.parseInt(portString);
+            if (port <= PORT_MIN || port > PORT_MAX) {
+                throw new IllegalArgumentException("Invalid port value: " + value);
```
jvz:
```suggestion
                throw new IllegalArgumentException("Port " + value + " out of valid range [" + PORT_MIN + ", " + PORT_MAX + ")");
```

Or something like that perhaps?

src/main/java/org/jenkinsci/remoting/engine/HostPort.java
```diff
@@ -1,28 +1,38 @@
...
+
+    private static final int PORT_MIN = 0;
+    private static final int PORT_MAX = (1 << 16) - 1;
+
     private String host;
     private int port;
 
-    public HostPort(String value) {
-        splitHostPort(value);
+    public HostPort(String value, String defaultHost, int defaultPort) {
```
oleg-nenashev: Breaking change

src/main/java/org/jenkinsci/remoting/engine/HostPort.java
```diff
@@ -1,28 +1,38 @@
...
             throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
         }
-        host = hostPortValue.substring(0, portSeparator).trim();
-        if (host.isEmpty()) {
-            throw new IllegalArgumentException("Invalid HOST:PORT value: " + value);
+        String hostValue = hostPortValue.substring(0, portSeparator).trim();
+        host = hostValue.length() > 0 ? hostValue : defaultHost;
+        String portString = hostPortValue.substring(portSeparator + 1).trim();
+        if (portString.length() > 0) {
+            port = Integer.parseInt(portString);
```
oleg-nenashev: Unhandled `NumberFormatException`

jeffret-b: `NumberFormatException` was unhandled in prior versions as well. I checked it by running 3.33 (before any of these changes) with a port value that is not a number. The agent fails on startup with the `NumberFormatException`.

Do you think we should change that behavior here now? Should it throw a different exception? Or log and ignore the error?

jeffret-b:
@oleg-nenashev, could you please take a look again?

PR #346, opened 2019-09-09
Fix missing `@since` annotations in Javadoc

res0nance:
Cleanup the remaining since documentation


jeffret-b:
Kick the build again. It looks like something was wrong with the environment the previous time. Once the build passes, I'll mark this ready-for-release and include it in the next release.

PR #347, opened 2019-09-10
Remove engineutil usage

res0nance:
- Use input stream reader instead. The one situation I think this is different is that `readline()` from stream reader returns null on EOF but `EngineUtil.readline()` returns a blank string. I'm not sure if that matters but it seems like that behaviour masks issues.

EDIT: Turns out spotbugs complains about it so I throw an exception in the case the socket stream EOFs


jeffret-b:
No clue what's up with the build. Let's try it again.

PR #351, opened 2019-10-28
Clean up "slave" term from documentation.

jeffret-b:
Where possible remove the term "slave" from documentation, replacing with "agent" as needed. In some instances, both were used but we're past the time where we need to continue using the old name.
Removes from AsciiDoc pages, JavaDoc, and code comments.
The remaining documentation references are for external or remaining API references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Jenkins Remoting layer
 
 Jenkins remoting is an executable JAR, 
 which implements communication layer in [Jenkins](https://jenkins.io) automation server. 
-It's being used for master <=> agent(fka "slave") and master <=> CLI communications.
+It's being used for master <=> agent and master <=> CLI communications.
 
 In general, this library contains the bootstrap code to bridge separate JVMs into a single semi-shared space.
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -6,7 +6,7 @@ Jenkins Remoting layer
 
 Jenkins remoting is an executable JAR, 
 which implements communication layer in [Jenkins](https://jenkins.io) automation server. 
-It's being used for master <=> agent(fka "slave") and master <=> CLI communications.
```
oleg-nenashev: I think it still makes sense to mention it here. We can amend it somehow to be explicit that the term is discouraged

jeffret-b: I thought about it, but I don't think we need to maintain this mention here. We could add it as a footnote somewhere. I've thought about trying to clean up the documentation some. I could add a note in there somewhere.

It's been quite a while since the terminology change and there are no shortage of places that mention that agents used to be known as slaves. Probably don't need to keep this one.

jeffret-b:
Let's make sure this builds before proceeding.

PR #352, opened 2019-10-28
Remove internal usages of "slave" term.

jeffret-b:
A couple of internal usages that really didn't need to old name.


jeffret-b:
And make sure this one builds.

PR #353, opened 2019-10-28
Remove some usages of slave from public fields.

jeffret-b:
I'm not sure if we want to remove the term in these instances. It depends on what we consider the API and how much we prioritize removing the term "slave" vs. maintaining any possible library use.

Essentially the API at this point are the command-line parameters (the `@Option` tags) and the object fields are internal (though they have the access modifier `public`). Any integration that steps beyond the CLI parameters into the fields is deep and rare.

I'm inclined to think it's better to make this change and continue the removal of the disfavored term. How concerned are others about maintaining the existing API at this level?

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -133,7 +133,7 @@ public class Engine extends Thread {
     @CheckForNull
     private URL hudsonUrl;
     private final String secretKey;
-    public final String slaveName;
+    public final String agentName;
     private String credentials;
     private String proxyCredentials = System.getProperty("proxyCredentials");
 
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -133,7 +133,7 @@ public Thread newThread(final Runnable r) {
     @CheckForNull
     private URL hudsonUrl;
     private final String secretKey;
-    public final String slaveName;
+    public final String agentName;
```
jvz: If you're renaming it, you could make it private.

oleg-nenashev:
FTR it could be tracked as a part of https://issues.jenkins-ci.org/browse/JENKINS-42816

jeffret-b:
Thanks for reviewing, Oleg! I'd like to get things like this cleaned up where we can. Like you, I don't find any usages. It really doesn't make sense to support arbitrary integrations at this level.

I haven't yet decided whether I want to proceed with this one or continue to ignore the word usage.

oleg-nenashev:
So let's also make the field private

jeffret-b:
> So let's also make the field private

Agreed on that point.

src/main/java/hudson/remoting/Engine.java
jeffret-b:
```suggestion
    private final String agentName;
```

PR #357, opened 2019-11-22
[JEP-222] WebSocket-based agents

jglick:
See [JEP-222](https://jenkins.io/jep/222).


pom.xml
```diff
@@ -342,6 +348,18 @@ THE SOFTWARE.
               <includeGroupIds>org.jenkins-ci</includeGroupIds>
             </configuration>
           </execution>
+          <execution>
+            <id>bundle-tyrus</id>
```
jglick: Perhaps this should be using the Shade plugin, to avoid leaking JSR classes into the classpath of everything using `remoting`.

jeffret-b:
A few comments as I've started looking at this. 

1. It works!
2. So far, the code is looking good.
3. We're going to need some documentation. (Unless I'm missing it somewhere.)
We could add it on to here: https://github.com/jenkinsci/remoting/blob/master/docs/tcpAgent.md . I've wanted a better place that gathers together more information about setting up agents but haven't been able to get anything assembled.
We could also document it in-app, on the Jenkins node page. It's already getting kind of cluttered, though.
4. Do we want to have a way of disabling WebSocket connectivity? I haven't seen that anywhere yet. My first thought is that we would want a UI control for that.
5. Final log message when launching is a little confusing. 
In the JNLP4 protocol once the agent is up and ready, the agent console log reads
"INFO: Connected".
For the WebSocket, it reads
"INFO: Waiting for channel"
This sounds to me like it is still waiting to establish the channel. (Yes, it's waiting for messages over the channel, but it's not clear that it's completed the connection sequence and it's different than what people might be used to.)
6. Do we have any reasons to recommend WebSockets over JNLP4 or vice versa?

jglick:
> We're going to need some documentation.

There is a bit in inline help downstream, but yes there should be more, including some here.

> Do we want to have a way of disabling WebSocket connectivity?

You mean on the master side? I see no particular reason to offer such an option. What would be the purpose?

> Final log message when launching is a little confusing.

Agreed.

> Do we have any reasons to recommend WebSockets over JNLP4

Mainly for nontrivial network topologies, as discussed in the JEP.

> or vice versa?

Maturity; perhaps scalability (unknown); built-in encryption (discussed in JEP).

jeffret-b:
Let's try a build again. (This test never fails on my machine.)

jeffret-b:
> > Do we want to have a way of disabling WebSocket connectivity?
> 
> You mean on the master side? I see no particular reason to offer such an option. What would be the purpose?

People seem to really like being able to control connectivity, particularly for complex things like agents, from a single, central location. I'm not sure that's a real requirement, though. Will there be some use case or scenario in which that capability is required? I'm not coming up with one -- it's probably best to not bother until we identify the need.

jglick:
> being able to control connectivity

Well we offer controls about listening on _ports_, for good reason. But `-webSocket` agents are not using a different port, just the usual Jenkins HTTP(S) URL, and if no connection is made to this URI then there is no difference in behavior. So it seems harmless to enable it unconditionally (rather, whenever using the built-in Jetty).

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -264,6 +268,38 @@ public static void _main(String[] args) throws IOException, InterruptedException
...
+                throw new CmdLineException(p, "-url is required in -webSocket mode", null);
+            }
+            if (m.urls.size() > 1) {
+                throw new CmdLineException(p, "multiple -url is not currently supported in -webSocket mode", null);
+            }
+            if (m.directConnection != null) {
+                throw new CmdLineException(p, "-webSocket and -direct are mutually exclusive", null);
+            }
+            if (m.tunnel != null) {
+                throw new CmdLineException(p, "-tunnel is not currently supported in -webSocket mode", null);
```
jeffret-b: I suggest dropping the word "currently" from this and the following messages. It's not clear why we would ever add support for some of these. If it comes up then we can change things anyway.

You could use the `forbids` attribute from args4j for each of these conflicting options, but this way also works.

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java
```diff
@@ -23,6 +23,7 @@
  */
 package org.jenkinsci.remoting.engine;
```
jeffret-b: Is something outdated or incomplete in this file? I'm not seeing that any of the changes here are actually needed.

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
```diff
@@ -301,7 +301,7 @@ public void onReceiveHeaders(Map<String, String> headers) throws ConnectionRefus
                     case INVALID:
                         LOGGER.log(Level.WARNING,
                                 "An attempt was made to connect as {0} from {1} with an invalid client certificate",
-                                new Object[]{clientName, event.getSocket().getRemoteSocketAddress()});
+                                new Object[]{clientName, event.getRemoteEndpointDescription()});
```
jeffret-b: The changes here are needed to adapt to the changes in `JnlpConnectionState` but those don't seem to be needed. These particular uses do make more sense, though.

docs/protocols.md
```diff
@@ -27,10 +26,15 @@ If stronger algorithms are needed (for example, AES with 256-bit keys), the [JCE
 
 Protocol uses non-blocking I/O wherever possible which removes the performance bottleneck of the <code>JNLP3-connect</code> protocol.
 
+### WebSocket
+
```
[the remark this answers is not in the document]
jglick: Agreed.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -496,6 +517,116 @@ public void run() {
...
+                    headers.put(JnlpConnectionState.SECRET_KEY, Collections.singletonList(secretKey));
+                    headers.put(Capability.KEY, Collections.singletonList(localCap));
+                    // TODO use JnlpConnectionState.COOKIE_KEY somehow (see EngineJnlpConnectionStateListener.afterChannel)
+                    LOGGER.fine(() -> "Sending: " + headers);
+                }
+                @Override
+                public void afterResponse(HandshakeResponse hr) {
+                    LOGGER.fine(() -> "Receiving: " + hr.getHeaders());
+                    List<String> remotingMinimumVersion = hr.getHeaders().get("X-Remoting-Minimum-Version");
+                    if (remotingMinimumVersion != null && !remotingMinimumVersion.isEmpty()) {
```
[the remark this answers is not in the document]
jglick: Well, the header names could become constants. That would also allow `jenkins-core` to share.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -496,6 +517,116 @@ public void run() {
...
+                        VersionNumber minimumSupportedVersion = new VersionNumber(remotingMinimumVersion.get(0));
+                        VersionNumber currentVersion = new VersionNumber(Launcher.VERSION);
+                        if (currentVersion.isOlderThan(minimumSupportedVersion)) {
+                            // TODO these errors should trigger a connection close
```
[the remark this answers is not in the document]
jglick: This one is tricky since `afterResponse` is not allowed to throw an exception or otherwise cancel the connection. Maybe `onOpen` could track the condition and immediately close the new connection, or maybe a `RuntimeException` works here. Can check into it.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -496,6 +517,116 @@ public void run() {
...
+                        remoteCapability = Capability.fromASCII(hr.getHeaders().get(Capability.KEY).get(0));
+                        LOGGER.fine(() -> "received " + remoteCapability);
+                    } catch (IOException x) {
+                        events.error(x);
+                    }
+                }
+            }
+            HeaderHandler headerHandler = new HeaderHandler();
+            class AgentEndpoint extends Endpoint {
+                @SuppressFBWarnings(value = "UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR", justification = "just trust me here")
```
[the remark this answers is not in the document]
jglick: Heh. The tangled code structure used by event-driven APIs makes it really hard to reason about variable initialization. Maybe one day Project Loom will solve some of these cases.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -496,6 +517,116 @@ public void run() {
...
+            }
+            ContainerProvider.getWebSocketContainer().connectToServer(new AgentEndpoint(),
+                ClientEndpointConfig.Builder.create().configurator(headerHandler).build(), URI.create(candidateUrls.get(0).toString().replaceFirst("^http", "ws") + "wsagents/"));
+            while (ch.get() == null) {
+                Thread.sleep(100);
+            }
+            LOGGER.info("Connected");
+            ch.get().join();
+            // TODO handle multiple candidate URLs
+            // TODO handle reconnection
```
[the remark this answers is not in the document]
jglick: > Do we need to resolve these TODOs before release?

TBD. It is not like we would be introducing a regression; WebSocket mode is a new feature you need to opt in to, and for some users it is a choice between that and not connecting a certain kind of agent at all. Of course it would be good to implement these things.

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -264,6 +268,38 @@ public static void _main(String[] args) throws IOException, InterruptedException
         if(m.urls.isEmpty() && m.directConnection == null) {
             throw new CmdLineException(p, "At least one -url option is required.", null);
         }
+        if (m.webSocket) {
+            if (m.urls.isEmpty()) {
+                throw new CmdLineException(p, "-url is required in -webSocket mode", null);
+            }
+            if (m.urls.size() > 1) {
+                throw new CmdLineException(p, "multiple -url is not currently supported in -webSocket mode", null);
```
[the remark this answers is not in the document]
jglick: A fine point of English grammar. Not really sure what is right. I was using the singular to refer to the whole argument line.

src/main/java/hudson/remoting/jnlp/Main.java
jglick: > It's not clear why we would ever add support for some of these.

I used _currently_ when there was some plausible reason why we might want such a combination. In the case of `-tunnel`, I had thought this would apply even to the HTTP metadata connection, but a closer reading of the TCP-mode code suggests that this is not so, so I will fix that.

> You could use the `forbids` attribute from args4j

Neat, I did not know about that.

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java
jglick: You need to look at the upstream PR.

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java
jglick: See above.

src/main/java/hudson/remoting/jnlp/Main.java
jeffret-b: The tunnel one is particularly one we don't want to propagate. It's problematic as it is.

src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java
jeffret-b: I'm only part way through that one. Sounds good.

src/main/java/hudson/remoting/Engine.java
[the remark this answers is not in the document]
jeffret-b: Yeah, I can make an argument either way. I'm concerned about introducing a new mode that is likely to cause people problems and report bugs. But, until people try it it's hard to know what the problems might be.

jeffret-b:
We should probably wait a day or two to see if anyone else wants to review but after that I'm satisfied that we can merge this and release it whenever we want.

docs/inbound-agent.md
```diff
@@ -33,7 +33,12 @@ This mechanism is not recommended, is deprecated, and may not be supported or av
 ### Download JNLP file
 Another mechanism, shown in the above status page fragment, runs the agent from a script or command-line to retrieve the JNLP file. 
 This mechanism does not use JNLP or WebStart but uses the downloaded file to obtain connection information. 
-Internally, this is a two-step process: 1) connect to an HTTP(S) port to retrieve the connection information, and 2) extract that information and connect to the TCP port.
+
+For an inbound agent using the TCP port, this is a two-step process:
+* Connect to an HTTP(S) port to retrieve the connection information.
+* Extract that information and connect to the TCP port.
+
+(When using the `-webSocket` option, only a single connection needs to be made.)
```
oleg-nenashev: Also for Direct inbound TCP agent connection in #338. Would be great to reference it as well

docs/protocols.md
```diff
@@ -27,10 +26,16 @@ If stronger algorithms are needed (for example, AES with 256-bit keys), the [JCE
...
+* Introduced in: Remoting version TODO, [JEP-222](https://jenkins.io/jep/222)
```
oleg-nenashev: Should it firstly be released with "Beta" disclaimers until there is a community feedback and JEP being accepted?

pom.xml
```diff
@@ -219,25 +212,57 @@ THE SOFTWARE.
           <targetJdk>1.${java.level}</targetJdk>
         </configuration>
       </plugin>
+      <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
```
oleg-nenashev: Might make sense to submit as a separate PR or to add to the changelog. Dependency shading is a major enhancement and, probably, just another justification for 4.0

docs/protocols.md
jglick: Makes sense. I think that can be inserted by the merger when inserting the actual version number here?

pom.xml
jglick: This was required to make sure that the newly added dependencies did not get dumped in the core classpath. (Same for the CLI JAR in the core PR.)

jglick:
> the WebSocket mode works well with thread UnhandledException failover for NIo threads in JNLP4

Not sure what you are saying.

docs/inbound-agent.md
jglick: Maybe…seems redundant. We might consider rewriting the whole page to discuss the various modes more clearly. For example, the Docker image bypasses the JNLP entry point and uses a different entry point in the JAR; this is alluded to in the direct connection section but there is no mention that this can be done even when _not_ using direct connection mode.

jglick:
> 4.0

This was suggested for the removal of the deprecated protocols, but too late. I have no strong opinion, and anyway consider the version numbers in this repository to be more or less meaningless—it is just a bundled component of Jenkins. Will leave it to @jeffret-b to decide this sort of thing.

docs/protocols.md
jeffret-b: I need to remember to do that.

jeffret-b: What sort of beta disclaimers do we want? Just a brief comment that this capability is considered beta until further notice?

jeffret-b:
The version numbers here really are mostly meaningless. It's a bundled Jenkins component. It's recommended to keep agent versions current with the server. There have been bigger user-facing changes in the past that haven't rolled the major version, such as upgrading Java versions or the deprecation of the older protocols. Changing the minimum version for the server is a bigger breaking change. If this were a generic library incorporated into other tools, it would make sense to follow semver. I'll go ahead and update it to 4.x, though. (It kind of looks like we update for adding major new functionality but not for breaking dependencies or changes.)

docs/protocols.md
jglick: Sounds right.

docs/inbound-agent.md
jeffret-b: I tossed this page together to get a minimal amount of information out there, though describing the minimum still took quite a bit. For a long time I've been wanting to get better user-facing documentation on how to use agents and distribute Jenkins. It's still on a list of things to do and we talk about it from time to time.

For now, let's leave this page as currently proposed in this PR.

pom.xml
jeffret-b: I'll plan on releasing as 4.0 and we'll leave this as proposed.

jeffret-b:
I have been planning to release this later today if I don't get distracted with something else. @oleg-nenashev, if you want to clarify your comment about "UnhandledException failover for NIo threads" I can delay. Especially since you've approved, it seems like that could be handled in a follow-on if we discover something needed there.

docs/protocols.md
jglick: Going to do that as a follow-up commit or what?

jeffret-b: Apparently. I missed my note to merge it in with the other. I'll just do it now.

Asgoret:
@jglick @oleg-nenashev @jeffret-b  Hi! 
I catch some error vie WebSocket connection, that slave tried to send a binary message more than 64Kib (actual about 98Kib in my case) and I didn't find any exposed configures for a max binary message size for WebSocket connection. 64Kib is a [RFC 6455](https://tools.ietf.org/html/rfc6455#section-5.7) standard. Can you help with it plz?

jglick:
@asgoret file an issue in JIRA with steps to reproduce.

Asgoret:
@jglick and what's the link to the jira?

jglick:
https://issues.jenkins-ci.org/

PR #361, opened 2019-12-05
Add findsecbugs plugin to spotbugs.

jeffret-b:
Add findsecbugs plugin to spotbugs. And suppress existing warnings. We should clean some of them up, but that's for different PRs at a later time.

Some of the issues findsecbugs just misidentifies. With some it doesn't understand enough context.

As always with these sort of of analysis tools, adding them to existing code correctly can be a bit of a challenge. It's easier to suppress things that aren't really problems than to clean all of them up. Most of the benefit from applying these tools comes from checking newly added code and preventing new issues from creeping in.

### DIFF
--- a/src/main/java/hudson/remoting/Capability.java
+++ b/src/main/java/hudson/remoting/Capability.java
@@ -1,5 +1,6 @@
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.remoting.Channel.Mode;
 import java.io.IOException;
 import java.io.InputStream;
@@ -145,6 +146,7 @@ public final class Capability implements Serializable {
     /**
      * The opposite operation of {@link #writePreamble(OutputStream)}.
      */
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Capability is used for negotiating channel between authorized agent and server.")
     public static Capability read(InputStream is) throws IOException {
         try (ObjectInputStream ois = new ObjectInputStream(Mode.TEXT.wrap(is)) {
                 // during deserialization, only accept Capability to protect ourselves
--- a/src/main/java/hudson/remoting/Capability.java
+++ b/src/main/java/hudson/remoting/Capability.java
@@ -146,7 +146,7 @@ public final class Capability implements Serializable {
     /**
      * The opposite operation of {@link #writePreamble(OutputStream)}.
      */
-    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Capability is used for negotiating channel between authorized agent and server.")
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Capability is used for negotiating channel between authorized agent and server. Whitelisting and proper deserialization hygiene are used.")
     public static Capability read(InputStream is) throws IOException {
         try (ObjectInputStream ois = new ObjectInputStream(Mode.TEXT.wrap(is)) {
                 // during deserialization, only accept Capability to protect ourselves
--- a/src/main/java/hudson/remoting/Checksum.java
+++ b/src/main/java/hudson/remoting/Checksum.java
@@ -1,5 +1,7 @@
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import java.io.ByteArrayInputStream;
 import java.io.DataInputStream;
 import java.io.File;
@@ -72,6 +74,7 @@ final class Checksum {
     /**
      * Returns the checksum for the given URL.
      */
+    @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "This is only used for managing the jar cache as files, not URLs.")
     static Checksum forURL(URL url) throws IOException {
         try {
             MessageDigest md = MessageDigest.getInstance(JarLoaderImpl.DIGEST_ALGORITHM);
--- a/src/main/java/hudson/remoting/Command.java
+++ b/src/main/java/hudson/remoting/Command.java
@@ -23,6 +23,8 @@
  */
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -146,6 +148,7 @@ public abstract class Command implements Serializable {
 
 
     /** Consider calling {@link Channel#notifyRead} afterwards. */
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Used for sending commands between authorized agent and server.)")
     static Command readFromObjectStream(Channel channel, ObjectInputStream ois) throws IOException, ClassNotFoundException {
         Channel old = Channel.setCurrent(channel);
         try {
--- a/src/main/java/hudson/remoting/Command.java
+++ b/src/main/java/hudson/remoting/Command.java
@@ -148,7 +148,7 @@ public abstract class Command implements Serializable {
 
 
     /** Consider calling {@link Channel#notifyRead} afterwards. */
-    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Used for sending commands between authorized agent and server.)")
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Used for sending commands between authorized agent and server. Class filtering is done through JEP-200.")
     static Command readFromObjectStream(Channel channel, ObjectInputStream ois) throws IOException, ClassNotFoundException {
         Channel old = Channel.setCurrent(channel);
         try {
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -23,6 +23,7 @@
  */
 package hudson.remoting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.remoting.Channel.Mode;
 import java.io.File;
 import java.io.FileInputStream;
@@ -438,6 +439,7 @@ public class Engine extends Thread {
     }
 
     @Override
+    @SuppressFBWarnings(value = "HARD_CODE_PASSWORD", justification = "Password doesn't need to be protected.")
     public void run() {
         // Create the engine
         try {
@@ -702,6 +704,7 @@ public class Engine extends Thread {
 
     private static final Logger LOGGER = Logger.getLogger(Engine.class.getName());
 
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "File path is loaded from system properties.")
     static KeyStore getCacertsKeyStore()
             throws PrivilegedActionException, KeyStoreException, NoSuchProviderException, CertificateException,
             NoSuchAlgorithmException, IOException {
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -623,6 +623,7 @@ public class Launcher {
      * Listens on an ephemeral port, record that port number in a port file,
      * then accepts one TCP connection.
      */
+    @Deprecated
     @SuppressFBWarnings(value = {"UNENCRYPTED_SERVER_SOCKET", "DM_DEFAULT_ENCODING"}, justification = "This is an old, insecure mechanism that should be removed. port number file should be in platform default encoding.")
     private void runAsTcpServer() throws IOException, InterruptedException {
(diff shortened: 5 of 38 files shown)


src/main/java/hudson/remoting/Capability.java
```diff
@@ -145,6 +146,7 @@ protected void annotateClass(Class<?> c) throws IOException {
     /**
      * The opposite operation of {@link #writePreamble(OutputStream)}.
      */
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Capability is used for negotiating channel between authorized agent and server.")
```
jvz: Perhaps you can use a justification that the ObjectInputStream used in this method clearly uses whitelisting and proper deserialization hygiene.

src/main/java/hudson/remoting/Checksum.java
```diff
@@ -72,6 +74,7 @@ static Checksum forFile(File file) throws IOException {
     /**
      * Returns the checksum for the given URL.
      */
+    @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "This is only used for managing the jar cache as files, not URLs.")
```
jvz: Could you add an assertion that the URL's scheme is file or at least not http/s?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -438,6 +439,7 @@ public void removeListener(EngineListener el) {
     }
 
     @Override
+    @SuppressFBWarnings(value = "HARD_CODE_PASSWORD", justification = "Password doesn't need to be protected.")
```
jvz: Maybe if you put the default `password` value as a constant?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -619,7 +623,7 @@ private static Document loadDom(URL agentJnlpURL, InputStream is) throws ParserC
      * Listens on an ephemeral port, record that port number in a port file,
      * then accepts one TCP connection.
      */
-    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings("DM_DEFAULT_ENCODING")    // port number file should be in platform default encoding
+    @SuppressFBWarnings(value = {"UNENCRYPTED_SERVER_SOCKET", "DM_DEFAULT_ENCODING"}, justification = "This is an old, insecure mechanism that should be removed. port number file should be in platform default encoding.")
```
jvz: Perhaps deprecate this?

src/main/java/hudson/remoting/Command.java
```diff
@@ -146,6 +148,7 @@ public static Command readFrom(@Nonnull Channel channel, @Nonnull byte[] payload
 
 
     /** Consider calling {@link Channel#notifyRead} afterwards. */
+    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Used for sending commands between authorized agent and server.)")
```
jvz:
```suggestion
    @SuppressFBWarnings(value = "OBJECT_DESERIALIZATION", justification = "Used for sending commands between authorized agent and server. Class filtering is done through JEP-200.")
```

src/main/java/org/jenkinsci/remoting/util/KeyUtils.java
```diff
@@ -96,6 +98,7 @@ private static boolean equals(String str1, String str2) {
      * @return the MD5 fingerprint of the key.
      */
     @Nonnull
+    @SuppressFBWarnings(value = "WEAK_MESSAGE_DIGEST_MD5", justification = "Used for fingerprinting, not security.")
```
oleg-nenashev: You never know, because it is a utility class

jeffret-b: Yes, we really ought to clean this up, replace some of these old algorithms. I'd like to see if we can get some of that going sometime.

src/main/java/hudson/remoting/Checksum.java
jeffret-b: You mean something like this
```
assert url.getProtocol().equalsIgnoreCase("file") : "Non-file URL protocol generating checksum";
```

src/main/java/hudson/remoting/Engine.java
jeffret-b: It's smart enough to detect that attempt to deceive it.

src/main/java/hudson/remoting/Checksum.java
jvz: Yeah basically. Might want to support vfs type URLs for some servlet runtimes (JBoss/Wildfly used to do this for files in the war rather than unzipping it like Tomcat/Jetty do).

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -610,6 +613,7 @@ private SSLSocketFactory getSSLSocketFactory()
         return r;
     }
 
+    @SuppressFBWarnings(value = "XXE_DOCUMENT", justification = "This is the (so-called) JNLP file from the server, which has to be trusted from the perspective of this agent.")
```
jvz: You can still disable external entities in the DBF/DB instances here.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -665,6 +670,7 @@ private void runOnSocket(Socket s) throws IOException, InterruptedException {
     /**
      * Connects to the given TCP port and then start running
      */
+    @SuppressFBWarnings(value = "UNENCRYPTED_SOCKET", justification = "This implements an old, insecure connection mechanism.")
```
jvz: Deprecate

src/main/java/hudson/remoting/Pipe.java
```diff
@@ -97,6 +98,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
+@SuppressFBWarnings(value = "DESERIALIZATION_GADGET", justification = "Serializable only over remoting.")
```
jvz: Good to mention the serialization hygiene again (JEP-200) as it's not immediately obvious what is being suppressed here (various `ObjectInputStream` usage below). Can this be narrowed to just the `readObject` method?

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -794,6 +794,7 @@ public ClassLoaderProxy(@Nonnull ClassLoader cl, Channel channel) {
             this.channel = channel;
         }
 
+        @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "This is only used for managing the jar cache as files.")
```
jvz: Just because that's how it's supposed to be used doesn't mean that's what this method does. It's public, and it's exploitable by using non-file URLs.

src/main/java/hudson/remoting/RemoteInputStream.java
```diff
@@ -48,6 +49,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
+@SuppressFBWarnings(value = "DESERIALIZATION_GADGET", justification = "Serializable only over remoting.")
```
jvz: This seems overly broad as well. Should include the JEP-200 note.

src/main/java/hudson/remoting/URLDeserializationHelper.java
```diff
@@ -69,12 +69,14 @@
     
     private static class SafeURLStreamHandler extends URLStreamHandler {
         @Override
+        @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "Used for safely handling URLs, not for opening a connection.")
```
jvz: > not for opening a connection

> `return new URL(u.toString()).openConnection(p);`

😕 ?

src/main/java/hudson/remoting/Util.java
```diff
@@ -60,6 +61,7 @@ public static void copy(InputStream in, OutputStream out) throws IOException {
     }
 
     @Nonnull
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "This path exists within a temp directory so the potential traversal is limited.")
```
jvz: Unless it's in some sort of jail, I don't see how being in a temporary directory makes it any less exploitable.

src/main/java/hudson/remoting/forward/ForwarderFactory.java
```diff
@@ -68,6 +69,7 @@ private ForwarderImpl(String remoteHost, int remotePort) {
             this.remotePort = remotePort;
         }
 
+        @SuppressFBWarnings(value = "UNENCRYPTED_SOCKET", justification = "Unused mechanism.")
```
jvz: Unused? Deprecate!

src/main/java/hudson/remoting/forward/PortForwarder.java
```diff
@@ -52,6 +53,7 @@
     private final Forwarder forwarder;
     private final ServerSocket socket;
 
+    @SuppressFBWarnings(value = "UNENCRYPTED_SERVER_SOCKET", justification = "Unused")
```
jvz: Deprecate!

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -500,6 +502,7 @@ else if(entry.endsWith("*"))
      * Credentials can be passed e.g. to support running Jenkins behind a (reverse) proxy requiring authorization
      * FIXME: similar to hudson.remoting.Util.openURLConnection which is still used in hudson.remoting.Launcher
      */
+    @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "Used by the agent for retrieving connection info from the server.")
```
jvz: So what's the secbug it's complaining about here? Can it be guarded just in case?

src/main/java/org/jenkinsci/remoting/util/KeyUtils.java
jvz: This should be marked as deprecated to further dissuade use then.

src/main/java/hudson/remoting/Checksum.java
jeffret-b: I'm concerned here by your comment that there might be other types we might want to also include in the assert. That was why I was reluctant to add the assertion initially. I suppose it won't hurt, but since we're not certain about the total set maybe it would be better to leave it off for now.

src/main/java/hudson/remoting/Launcher.java
jeffret-b: Good point. It doesn't hurt to add that, even though the server has to be trusted.

src/main/java/hudson/remoting/Pipe.java
jeffret-b: It looks like the annotation can be narrowed to just the `readObject` method. It reports the finding on the entire class so that's where I put the suppression, but putting it on the method also quiets it.

src/main/java/hudson/remoting/RemoteClassLoader.java
jeffret-b: That's true. This really should have been implemented differently, but it's complicated to change that implementation at this point. Might be worth creating another PR at some point to clean some of this up. I don't know if there's a better way to phrase the suppression or to fix it at this point.

src/main/java/hudson/remoting/URLDeserializationHelper.java
jeffret-b: Yeah, this one is kind of iffy. The history of this bit of code is weird. Part of it comes from a specific, odd concern. Looking at this further, it might allow opening URL's on the agent side, if the job requires it. That's part of trusting the server, because the agent jobs can do about anything anyway. I could dig into this one more maybe.

src/main/java/hudson/remoting/Util.java
jeffret-b: It is in some weird kind of jail. I can't remember the details but I had trouble finding a way for this to escape. The things I tried didn't work.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
jeffret-b: "Server-Side Request Forgery occur when a web server executes a request to a user supplied destination parameter that is not validated. Such vulnerabilities could allow an attacker to access internal services or to launch attacks from your web server. "

But, it's the agent requesting the JNLP file from the server from user / administrator supplied values. The rules in findsecbugs are clearly biased towards webservers. Some of them don't apply to launching the agent.

jeffret-b:
@jvz, thanks for the review! I've updated the PR for many of the findings and responded to some of your comments / questions. Please respond again or take another look.

jeffret-b:
Try building again.

jeffret-b:
With a close

PR #364, opened 2019-12-30
[JENKINS-60606] - Jenkinsfile: Use the standard infra.runMaven() step for Linux builds

oleg-nenashev:
I am not sure whether the original configuration really passes the settings file... CC @jglick @slide just in case there is some magic in ACI


oleg-nenashev:
``Parallel / linux` - Build duration according to BlueOcean:

- Before: 2m 42s
- After:  1m 49s

Jenkinsfile
```diff
@@ -14,7 +14,7 @@ parallel linux: {
         }
         stage('Build') {
             timeout(30) {
-                sh 'mvn -B -ntp -Dset.changelist clean install -Dmaven.test.failure.ignore -s settings-azure.xml -e'
```
jglick: So delete `settings-azure.xml`. But first you would need to call `retrieveMavenSettingsFile`. And that requires `pwd tmp: true`.

Jenkinsfile
```diff
@@ -14,7 +14,7 @@ parallel linux: {
...
+                infra.runMaven(['-Dset.changelist', 'clean', 'install', '-Dmaven.test.failure.ignore', '-e'])
```
jglick:
```suggestion
                infra.runMaven(['-Dset.changelist', 'clean', 'install', '-Dmaven.test.failure.ignore'])
```

(already included)

Jenkinsfile
jeffret-b: There seem to be required remaining pieces of @jglick's comment.

oleg-nenashev:
Resolved the merge conflict. Will merge after the build passes

jeffret-b (changes requested):
Have you resolved Jesse's concerns about "-ntp"?

jglick:
I think the stock function now uses `-ntp`. Anyway this does not seem to be contributing much of anything. We are no longer using a mirror.

jglick (commented):
Offers no functional benefit that I know of, just forces readers to look up another repository for the definition of a library function. We are already using the `maven`-labelled agent so this function basically does nothing new. At best this accomplishes nothing; at worst it gratuitously adds a call to `tool`.

jglick (changes requested):
Two calls to `tool`, actually (just checked), neither of which we need.

jeffret-b:
There doesn't seem to be any reason to retain this PR so I'm closing it.

PR #366, opened 2020-01-22
Add javax.websocket to pom.xml

jsoref:
IntelliJ configured to use adoptopenjdk openj9 1.8 complains about:
https://github.com/jenkinsci/remoting/blob/ca2fbc659c5b2a899bbce6c22d3a0ff87c21de58/src/main/java/hudson/remoting/Engine.java#L70-L76

This appears to make it happy.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,11 @@ THE SOFTWARE.
       <version>1.12</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.websocket</groupId>
+      <artifactId>javax.websocket-api</artifactId>
+      <version>1.1</version>
+    </dependency>
   </dependencies>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -182,6 +182,11 @@ THE SOFTWARE.
       <version>1.12</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.websocket</groupId>
+      <artifactId>javax.websocket-api</artifactId>
+      <version>1.1</version>
```
jvz: Might want to add `<scope>provided</scope>` since this is supposed to be bundled with the websocket implementation?

jeffret-b:
I'm not sure that makes a real difference but that's better to specify the scope.

PR #368, opened 2020-01-30
[JENKINS-60926] Fix "Install as a service" on Windows.

jeffret-b:
See [JENKINS-60926](https://issues.jenkins-ci.org/browse/JENKINS-60926).

A renaming of Engine.slaveName to Engine.agentName and making it
private broke a complicated interaction with the server and the slave-installer-module.
Rename just this one part back to get it working again.

The previously unknown usage is in [slave-installer-module](https://github.com/jenkinsci/slave-installer-module/blob/3f1fce8ddb1de41b280546755700c727311448b9/src/main/java/org/jenkinsci/modules/slave_installer/impl/InstallerGui.java#L142).

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -153,6 +153,9 @@ public class Engine extends Thread {
     @CheckForNull
     private URL hudsonUrl;
     private final String secretKey;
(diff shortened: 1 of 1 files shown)


jeffret-b:
Good idea. Thanks!

PR #369, opened 2020-02-05
Cleanup some test warnings and deprecated code

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Cleanup some of the warnings

### DIFF
--- a/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
+++ b/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
@@ -56,7 +56,8 @@ public class DiagnosedStreamCorruptionExceptionTest {
     @Test(timeout=3000)
     public void blockingStreamShouldNotPreventDiagnosis() throws Exception {
         FastPipedInputStream in = new FastPipedInputStream();
-        try (FastPipedOutputStream out = new FastPipedOutputStream(in)) {
+        try (FastPipedInputStream in = new FastPipedInputStream();
+            FastPipedOutputStream out = new FastPipedOutputStream(in)) {
             out.write(payload);
         }
 
--- a/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
+++ b/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
@@ -55,18 +55,17 @@ public class DiagnosedStreamCorruptionExceptionTest {
      */
     @Test(timeout=3000)
     public void blockingStreamShouldNotPreventDiagnosis() throws Exception {
-        FastPipedInputStream in = new FastPipedInputStream();
         try (FastPipedInputStream in = new FastPipedInputStream();
             FastPipedOutputStream out = new FastPipedOutputStream(in)) {
             out.write(payload);
+
+            ClassicCommandTransport ct = (ClassicCommandTransport)
+                    new ChannelBuilder("dummy",null)
+                        .withMode(Mode.BINARY)
+                        .withBaseLoader(getClass().getClassLoader())
+                        .negotiate(in, new NullOutputStream());
+
+            verify(ct);
         }
-
-        ClassicCommandTransport ct = (ClassicCommandTransport)
-                new ChannelBuilder("dummy",null)
-                    .withMode(Mode.BINARY)
-                    .withBaseLoader(getClass().getClassLoader())
-                    .negotiate(in, new NullOutputStream());
-
-        verify(ct);
     }
 }
--- a/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
+++ b/src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
@@ -56,8 +56,9 @@ public class DiagnosedStreamCorruptionExceptionTest {
     @Test(timeout=3000)
     public void blockingStreamShouldNotPreventDiagnosis() throws Exception {
         FastPipedInputStream in = new FastPipedInputStream();
-        FastPipedOutputStream out = new FastPipedOutputStream(in);
-        out.write(payload);
+        try (FastPipedOutputStream out = new FastPipedOutputStream(in)) {
+            out.write(payload);
+        }
 
         ClassicCommandTransport ct = (ClassicCommandTransport)
                 new ChannelBuilder("dummy",null)
--- a/src/test/java/hudson/remoting/TrafficAnalyzer.java
+++ b/src/test/java/hudson/remoting/TrafficAnalyzer.java
@@ -21,23 +21,24 @@ public class TrafficAnalyzer {
         File f = new File("/home/kohsuke/ws/hudson/investigations/javafx-windows-hang/out.log");
         DataInputStream fin = new DataInputStream(new FileInputStream(f));
         fin.readFully(new byte[4]); // skip preamble
-        ObjectInputStream ois = new ObjectInputStream(fin);
-        for (int n=0; ; n++) {
-            Command o = (Command)ois.readObject();
-            System.out.println("#"+n+" : "+o);
-            if (o instanceof RPCRequest) {
-                RPCRequest request = (RPCRequest) o;
-                System.out.print("  (");
-                boolean first=true;
-                for (Object argument : request.getArguments()) {
-                    if(first)   first=false;
-                    else        System.out.print(",");
-                    System.out.print(argument);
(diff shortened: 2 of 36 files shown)


src/test/java/org/jenkinsci/remoting/protocol/ProtocolStackImplTest.java
```diff
@@ -216,6 +216,7 @@ public void pipeBasicBackToBack() throws Exception {
         west.get().send(data);
         east.get().awaitByteContent(is(expected));
         assertThat(east.get().asByteArray(), is(expected));
+        //TODO hangs with close(null)
```
res0nance: This is somewhat concerning though as `close()` calls `close(null)` but only the latter causes hanging

jeffret-b: I took a quick look at this. Because of polymorphism there's an important change to what the code flow looks like it should be from static analysis (going to the declaration in the IDE) and the actual flow when running / debugging. The difference is specific to the tests in how it simulates different endpoints. It looks like changing the signature [here](https://github.com/jenkinsci/remoting/blob/master/src/test/java/org/jenkinsci/remoting/protocol/IOBufferMatcherLayer.java#L50) and overriding a different method could fix it, as long as all the test usages are updated.

I'm not sure just adding the TODO comments on the usages adds more value than just leaving it as it is. Fixing it may be the best way to go, though.

src/test/java/hudson/remoting/DiagnosedStreamCorruptionExceptionTest.java
```diff
@@ -56,8 +56,9 @@ private void verify(ClassicCommandTransport ct) throws IOException, ClassNotFoun
     @Test(timeout=3000)
     public void blockingStreamShouldNotPreventDiagnosis() throws Exception {
         FastPipedInputStream in = new FastPipedInputStream();
-        FastPipedOutputStream out = new FastPipedOutputStream(in);
-        out.write(payload);
+        try (FastPipedOutputStream out = new FastPipedOutputStream(in)) {
```
MarkEWaite: Shouldn't both the stream opening statements be inside the try with resources?
```suggestion
        try (FastPipedInputStream in = new FastPipedInputStream();
            FastPipedOutputStream out = new FastPipedOutputStream(in)) {
```

src/test/java/hudson/remoting/TrafficAnalyzer.java
```diff
@@ -21,23 +21,24 @@ public static void main(String[] args) throws Exception {
         File f = new File("/home/kohsuke/ws/hudson/investigations/javafx-windows-hang/out.log");
         DataInputStream fin = new DataInputStream(new FileInputStream(f));
```
MarkEWaite: Shouldn't the read also be included in a try with resources?
```suggestion
        try (DataInputStream fin = new DataInputStream(new FileInputStream(f))) {
```

src/test/java/hudson/remoting/TrafficAnalyzer.java
```diff
@@ -21,23 +21,24 @@ public static void main(String[] args) throws Exception {
...
                 }
-                System.out.println(")");
+                if (o.createdAt!=null)
+                    o.createdAt.printStackTrace(System.out);
             }
-            if (o.createdAt!=null)
-                o.createdAt.printStackTrace(System.out);
         }
     }
 }
```
MarkEWaite: If the earlier comment about including the input in a try with resources is correct and useful, then something like this will be needed
```suggestion
    }
}
```

src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
```diff
@@ -155,7 +155,7 @@ public void tearDown() {
     }
 
     @Theory
-    public void happyPath(Factory factory, boolean useNioHubServer, boolean useNioHubClient) throws Exception {
+    public void happyPath(Factory factory, boolean useNioHubServer, boolean useNioHubClient) throws Throwable {
```
MarkEWaite: I don't understand the motivation for this change.  Can you explain further why the throw was broadened from Exception to Throwable?  The [stackoverflow article](https://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch) indicates that Throwable is broader than Exception, but I don't understand why a caller would need to know that this Theory may throw expections which are Throwable but are not Exceptions.

src/test/java/org/jenkinsci/remoting/protocol/ProtocolStackImplTest.java
MarkEWaite: It seems like something else must be happening that causes the difference in behavior between `close()` and `close(null)`.

I assume that difference is not new behavior, but newly discovered behavior that has existed for an unknown period.  Is my assumption correct?

src/test/java/org/jenkinsci/remoting/protocol/impl/SSLEngineFilterLayerTest.java
```diff
@@ -558,8 +559,8 @@ public void sendingBiggerAndBiggerBatches(NetworkLayerFactory serverFactory, Net
 
         client.awaitByteContent(SequentialSender.matcher(amount));
 
-        client.close();
-        server.close();
+        client.close(null);
```
MarkEWaite: Based on the earlier comments in this pull request, I assumed that
```
client.close(null);
```

might have problems.  Yet it does not. Any idea of the differences between this working case and the earlier hanging case?

src/test/java/org/jenkinsci/remoting/protocol/ProtocolStackImplTest.java
jeffret-b: Thanks for reviewing @MarkEWaite! I think your comment and mine probably crossed. The difference in behavior is due to runtime polymorphism. It's specific to the test infrastructure for creating multiple in-process connection endpoints. I'm sure it's been around for a while.

It's probably pretty easy to correct if @res0nance (or someone else) would like to spend a few minutes on it.

src/test/java/org/jenkinsci/remoting/protocol/impl/SSLEngineFilterLayerTest.java
jeffret-b: Probably a different concrete implementation class at runtime. I haven't verified that, yet, though.

src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
res0nance: My IDE was complaining about it, I don't mind reverting it

jeffret-b: Reverting that would be better. I don't see that it needs to widened to Throwable.

jeffret-b:
@res0nance, here are three things I think would be worth doing in this changeset.

1) Change the "throws Throwable" back to "throws Exception". (I think we're agreed on that one.)
2) Since you are adding the try-with-resources, include the enhancements Mark proposed.
3) (Optional) Try to fix the `close(null)` problem instead of adding the TODO comments.

src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java
jeffret-b: Looks like you missed changing this one back.

jeffret-b: Now I see why you needed to change this from an Exception to a Throwable. This change [here](https://github.com/jenkinsci/remoting/pull/369/files#diff-9138943fb30c237d96a6976f2eab1105R40) changes the signature of this line [here](https://github.com/jenkinsci/remoting/blob/621b672f67ba595d9fd73f5fd748e0ea2e56cbe6/src/test/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerTest.java#L167) so that it now throws Throwable rather than Exception. Since it's test code, it should be okay changing the signature.

We've given you some bad advice about changing it back to Exception because we didn't understand how that other change ripples through.

This is why the build is failing.

res0nance: Oh well, It's all good.

The inheritance change is probably for the best so the class will advertise its exception and return type correctly

PR #370, opened 2020-02-06
[JENKINS-60926] Expose the agent name via a getter

jeffret-b:
Different approach to earlier change. Keep the renaming and add a getter for other modules to use.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -153,10 +153,7 @@ public class Engine extends Thread {
     @CheckForNull
     private URL hudsonUrl;
     private final String secretKey;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -997,6 +994,16 @@ private SSLSocketFactory getSSLSocketFactory()
      */
     static final int SOCKET_TIMEOUT = Integer.getInteger(Engine.class.getName()+".socketTimeout",30*60*1000);
 
+    /**
+     * Get the agent name associated with this Engine instance.
+     *
+     * @return the agent name.
```
oleg-nenashev:
```suggestion
     * @return the agent name.
     * @since TODO
```

PR #372, opened 2020-02-15
[WIP][JENKINS-61103] - improve handling of RemotingSystemException(InterruptedException) in RemoteClassLoader

thomasgl-orange:
See [JENKINS-61103](https://issues.jenkins-ci.org/browse/JENKINS-61103)

Adds a dynamic proxy over the `IClassLoader` proxy in `RemoteClassLoader`, which will handle `RemotingSystemException`s caused by an `InterruptedException`.  The overall goal is to avoid letting a remote interruption irremediably damage loading/initialization of some classes.

- first commit workarounds a compilation "error" in a test, wrongly reported by my IDE (totally unrelated to the rest of the PR).
- second commit "fixes" a `RemoteClassLoading` test case which was not actually failing when I was removing what it was (I think) supposed to test (the retry logic in `RemoteClassLoader#findClass`).
- third commit is the actual proposed change.
- fourth commit adds a `RemoteClassLoading` test case showing that class initialization can also be damaged by a failure to load a resource.

I went for a dynamic proxy to keep code short enough, but of course this could have been a full concrete implementation of `IClassLoader` as well. I honestly don't know if it matters (from a performance point of view maybe?), please let me know if it is not appropriate.

Regarding the specific logic in this proxy, to decide whether a `RemotingSystemException` deserves special handling, there is room for discussion. I've tried to keep it simple, but for instance it could be argued that `getResource*` methods don't deserve a retry unless called during some class initialization (can be implemented, we have the calling stack available in the caught exception, we can look for `<clinit>`).

Regarding the two example error stack traces reported in [JENKINS-61103](https://issues.jenkins-ci.org/browse/JENKINS-61103), note that I don't have a unit test to prove that this change would actually fix the one with the exception raised from the `proxy.fetch(name)` call at [RemoteClassLoader.java#L315](https://github.com/jenkinsci/remoting/blob/remoting-3.36.1/src/main/java/hudson/remoting/RemoteClassLoader.java#L315).  FYI, the only unit test I've found which covers this line is [PrefetchingTest#testInnerClass](https://github.com/jenkinsci/remoting/blob/remoting-3.36.1/src/test/java/hudson/remoting/PrefetchingTest.java#L190), and so far I've not been able to work out something similar with a forced interruption via `RemoteClassLoader.TESTING_CLASS_LOAD`.

Final note, I've not really tested the proposed changes on top of `master`, but rather on top of 3.36.1. And my tests have been pretty basic so far, it is not something I have yet deployed on any "real" Jenkins servers, so there might be side effects I've not yet discovered.

### DIFF
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -141,7 +141,7 @@ final class RemoteClassLoader extends URLClassLoader {
         } else {
             targetProxy = proxy;
         }
-        if (MAX_RETRIES_ON_REMOTING_INTERRUPTION > 0) {
+        if (RemoteInterruptedExceptionCatcher.shouldBeEnabled()) {
             // JENKINS-61103: wrap the target (IClassLoader) proxy in a dynamic proxy which will handle retries in case
             // of interruption exception from the remoting system during class loading/initialization.
             // It should take care of cases like JENKINS-36991, JENKINS-51854 and others.
@@ -1124,8 +1124,8 @@ final class RemoteClassLoader extends URLClassLoader {
     /**
      * An {@link InvocationHandler} for proxying calls to a target {@link IClassLoader} which catches
      * {@link RemotingSystemException} caused by an {@link InterruptedException} (or {@link InterruptedIOException}).
-     * When this happens, the target method is called again, until it successfully returns (or throws a different kind
-     * of exception), or a maximum number of retries has been reached. The current thread is then interrupted again.
+     * When this happens, the target method is called again, until it successfully returns, throws a different kind
+     * of exception, or a maximum number of retries has been reached. The current thread is then interrupted again.
      * <p>
      * The purpose of this proxy is to avoid interruption of class loading/initialization, because that's something a
      * class loader can't recover from. Later attempts at using an affected class will result in a {@link LinkageError},
@@ -1156,7 +1156,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         break;
                     } catch (InvocationTargetException x) {
                         Throwable e = x.getCause(); // get the actual exception thrown by the target method call
-                        if (retryCounter++ >= MAX_RETRIES_ON_REMOTING_INTERRUPTION || !shouldRetry(method, e)) {
+                        if (!shouldRetry(method, e, retryCounter++)) {
                             throw e;
                         }
                         // defer interruption, and try again
@@ -1180,14 +1180,23 @@ final class RemoteClassLoader extends URLClassLoader {
          *
          * @param method the target method
          * @param e the {@link Throwable} which has been previously thrown
+         * @param nbPreviousAttempts number of times we've already attempted this call
          * @return true if the method invocation should be retried
          */
-        private boolean shouldRetry(Method method, Throwable e) {
-            return e instanceof RemotingSystemException
+        private boolean shouldRetry(Method method, Throwable e, int nbPreviousAttempts) {
+            return nbPreviousAttempts < MAX_RETRIES_ON_REMOTING_INTERRUPTION
+                    && e instanceof RemotingSystemException
                     && (e.getCause() instanceof InterruptedException
                             || e.getCause() instanceof InterruptedIOException)
                     // fetchJar() is only called from PreloadJarTask, we're not actually loading/initializing classes
                     && !"fetchJar".equals(method.getName());
+        }
+
+        /**
+         * @return true if this dynamic proxy should be used at all
+         */
+        public static boolean shouldBeEnabled() {
+            return MAX_RETRIES_ON_REMOTING_INTERRUPTION > 0;
         }
     }
 
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -356,6 +356,11 @@ final class RemoteClassLoader extends URLClassLoader {
     static Runnable TESTING_CLASS_REFERENCE_LOAD;
 
     /**
+     * Intercept {@link RemoteClassLoader#findResource(String)} to allow unittests to be written.
+     */
+    static Runnable TESTING_RESOURCE_LOAD;
+
+    /**
      * Loads class from the byte array.
      * @param name Name of the class
      * @param bytes Bytes
@@ -422,6 +427,8 @@ final class RemoteClassLoader extends URLClassLoader {
                 URL u = f.toURL();
                 if (u!=null)    return u;
             }
+
+            if (TESTING_RESOURCE_LOAD != null) TESTING_RESOURCE_LOAD.run();
 
             long startTime = System.nanoTime();
 
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -441,7 +441,9 @@ final class RemoteClassLoader extends URLClassLoader {
                 if (u!=null)    return u;
             }
 
-            if (TESTING_RESOURCE_LOAD != null) TESTING_RESOURCE_LOAD.run();
+            if (TESTING_RESOURCE_LOAD != null) {
+                TESTING_RESOURCE_LOAD.run();
+            }
 
             long startTime = System.nanoTime();
 
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -32,8 +32,11 @@ import javax.annotation.Nonnull;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Proxy;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -130,10 +133,48 @@ final class RemoteClassLoader extends URLClassLoader {
         final Channel channel = RemoteInvocationHandler.unwrap(proxy);
         this.channel = channel == null ? null : channel.ref();
         this.underlyingProxy = proxy;
+        final IClassLoader targetProxy;
         if (channel == null || !channel.remoteCapability.supportsPrefetch() || channel.getJarCache()==null) {
-            proxy = new DumbClassLoaderBridge(proxy);
+            targetProxy = new DumbClassLoaderBridge(proxy);
+        } else {
+            targetProxy = proxy;
         }
-        this.proxy = proxy;
+        // JENKINS-61103: wrap the target (IClassLoader) proxy in a dynamic proxy which will handle retries in case of
+        // interrupt exception from the remoting system during class loading/initialization.
+        // It should take care of cases like JENKINS-36991, JENKINS-51854 and others.
+        this.proxy = (IClassLoader) Proxy.newProxyInstance(RemoteClassLoader.class.getClassLoader(),
+                new Class[] { IClassLoader.class },
+                (p, method, methodArgs) -> {
+                    Object result;
+                    boolean interrupted = false;
+                    try {
+                        while (true) {
+                            try {
+                                result = method.invoke(targetProxy, methodArgs);
+                                break;
+                            } catch (InvocationTargetException x) {
+                                Throwable e = x.getCause();
+                                if (e instanceof RemotingSystemException
+                                        // this really look like remote interruptions we should defer, to avoid screwing
+                                        // up class loading/initialization:
+                                        && (e.getCause() instanceof InterruptedException
+                                                || e.getCause() instanceof InterruptedIOException)
+                                        // this is only called from PreloadJarTask, we're not actually loading classes:
+                                        && !"fetchJar".equals(method.getName())) {
+                                    interrupted = true;
+                                } else {
+                                    throw e;
+                                }
+                            }
+                        }
+                    } finally {
+                        // honor deferred interruption
+                        if (interrupted) {
+                            Thread.currentThread().interrupt();
+                        }
+                    }
+                    return result;
+                });
     }
 
     /**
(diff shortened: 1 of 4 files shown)


thomasgl-orange:
Oops, push-forced to remove a leftover `System.out.println()`...

jeffret-b:
Thanks for the submission! I'll try to take a look at it in the next few days. The build succeeds on my machine.

jeffret-b:
I'll try re-opening the PR to kick the CI run again.

jeffret-b:
Build is good this time.

jeffret-b:
Note: It looks like the description swaps the changes that occurred in the third and fourth commits. The third commit is the change. The fourth is another test.

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -130,10 +133,48 @@ private RemoteClassLoader(@CheckForNull ClassLoader parent, @Nonnull IClassLoade
...
-            proxy = new DumbClassLoaderBridge(proxy);
+            targetProxy = new DumbClassLoaderBridge(proxy);
+        } else {
+            targetProxy = proxy;
         }
-        this.proxy = proxy;
+        // JENKINS-61103: wrap the target (IClassLoader) proxy in a dynamic proxy which will handle retries in case of
+        // interrupt exception from the remoting system during class loading/initialization.
+        // It should take care of cases like JENKINS-36991, JENKINS-51854 and others.
+        this.proxy = (IClassLoader) Proxy.newProxyInstance(RemoteClassLoader.class.getClassLoader(),
```
jeffret-b: My biggest concern with this approach is the deep nesting and how it impacts readability. There are a number of different options, but we should try something to improve readability.

My next concern is that I have been bitten in the past by performance degradation from reflection. Most of the time it's not enough to cause concern, but I'm a little worried about what it might do in this area.

I like how this handles it at a central, localized place, but I think it would be worth exploring if it can be done in a more readable fashion without reflection performance concerns using something different from a dynamic proxy here.

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -130,10 +133,48 @@ private RemoteClassLoader(@CheckForNull ClassLoader parent, @Nonnull IClassLoade
...
+                            } catch (InvocationTargetException x) {
+                                Throwable e = x.getCause();
+                                if (e instanceof RemotingSystemException
+                                        // this really look like remote interruptions we should defer, to avoid screwing
+                                        // up class loading/initialization:
+                                        && (e.getCause() instanceof InterruptedException
+                                                || e.getCause() instanceof InterruptedIOException)
+                                        // this is only called from PreloadJarTask, we're not actually loading classes:
+                                        && !"fetchJar".equals(method.getName())) {
+                                    interrupted = true;
```
jeffret-b: This looks like it could get blocked forever. It needs some kind of timeout or retry counter.

thomasgl-orange:
Thanks a lot @jeffret-b for the review. That's all valid points, and I will rework this PR. Flagging it "[WIP]" in the mean time.

> Note: It looks like the description swaps the changes that occurred in the third and fourth commits. The third commit is the change. The fourth is another test.

Right. The order in description was what I intended to submit (failing tests before the fix), but it looks like I forgot to reorder commits on this branch. Oh well, too late, I will change the description.

jeffret-b:
I wasn't worried about the out-of-order commits related to the description. I was just leaving a note for reviewers, including myself. It's probably better to mark them with a commit sha or a link than to rely on ordering.

I look forward to seeing your improved PR. This will be a nice improvement.

romanek-adam:
@thomasgl-orange what's the outlook for this PR? Do you think the review comments could be addressed soon? Could I help somehow to push this forward?

In any case thanks for the hard work!

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -423,6 +428,8 @@ public URL findResource(String name) {
                 if (u!=null)    return u;
             }
 
+            if (TESTING_RESOURCE_LOAD != null) TESTING_RESOURCE_LOAD.run();
```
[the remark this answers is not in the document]
thomasgl-orange: 3edc2be

src/main/java/hudson/remoting/RemoteClassLoader.java
thomasgl-orange: 85a19dee is an attempt at improving readability. The `InvocationHandler` is not lambda-defined anymore, it has comments, and the logic to decide whether there could be retries given a method and a caugh interruption is not buried in the middle of the `if` condition anymore (that's `RemoteClassLoader.RemoteInterruptedExceptionCatcher.shouldRetry(Method,Throwable)`).

This is still using a dynamic proxy though. I've not found time to work more on this and experiment alternative solutions, or add jmh benchmarks. So that's all I have for now, I thought it's better to share despite it not being an exhaustive answer to your concerns, because time flies. 

I'm not super worried about the impact of basing implementation on a dynamic proxy actually; it is "only" adding one more layer in calls to an other dynamic proxy (`RemoteInvocationHandler`), so at worst it's doubling an already existing penalty :)

But yes, would be good to explore perfs and alternatives, I just can't tell when I will be able to do it.

src/main/java/hudson/remoting/RemoteClassLoader.java
thomasgl-orange: I can't think of how it could happen for `InterruptedException`, but it probably can for `InterruptedIOException`, so yes, better safe than sorry. 

85a19dee adds a retries counter, with a configurable limit (2 by default; I even think 1 would be enough to handle common scenarios actually).

This also gives a convenient way to disable the retry logic and the whole dynamic proxy thing (set `MAX_RETRIES_ON_REMOTING_INTERRUPTION` to 0), for instance to check that some tests actually fail without it.

thomasgl-orange:
> what's the outlook for this PR? Do you think the review comments could be addressed soon?

Well, I've just pushed what I had so far, because it's been almost two weeks and I've not found time to do more. I want to (in particular I'm curious to see what some jmh benchmarks could show about the perfs impact of the dynamic proxy solution), but I really can't give a schedule.

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -130,10 +135,21 @@ private RemoteClassLoader(@CheckForNull ClassLoader parent, @Nonnull IClassLoade
...
         this.channel = channel == null ? null : channel.ref();
         this.underlyingProxy = proxy;
+        final IClassLoader targetProxy;
         if (channel == null || !channel.remoteCapability.supportsPrefetch() || channel.getJarCache()==null) {
-            proxy = new DumbClassLoaderBridge(proxy);
+            targetProxy = new DumbClassLoaderBridge(proxy);
+        } else {
+            targetProxy = proxy;
+        }
+        if (MAX_RETRIES_ON_REMOTING_INTERRUPTION > 0) {
```
jeffret-b: There's a readability refactoring here that would be nice. I was confused on reading this and may not be the only one.

I propose an Extract Method refactoring that looks something like this:
```suggestion
        if (shouldRetryOnRemotingInterruption()) {
```

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -1124,6 +1121,88 @@ private Object readResolve() throws ObjectStreamException {
         }
     }
 
+    /**
+     * An {@link InvocationHandler} for proxying calls to a target {@link IClassLoader} which catches
+     * {@link RemotingSystemException} caused by an {@link InterruptedException} (or {@link InterruptedIOException}).
+     * When this happens, the target method is called again, until it successfully returns (or throws a different kind
+     * of exception), or a maximum number of retries has been reached. The current thread is then interrupted again.
```
jeffret-b: nit: 

Maybe something like this:
```suggestion
     * When this happens, the target method is called again, until it successfully returns, throws a different kind
     * of exception, or a maximum number of retries has been reached. The current thread is then interrupted again.
```

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -1124,6 +1121,88 @@ private Object readResolve() throws ObjectStreamException {
...
+            boolean interrupted = false;
+            int retryCounter = 0;
+            try {
+                do {
+                    try {
+                        result = method.invoke(targetProxy, args);
+                        break;
+                    } catch (InvocationTargetException x) {
+                        Throwable e = x.getCause(); // get the actual exception thrown by the target method call
+                        if (retryCounter++ >= MAX_RETRIES_ON_REMOTING_INTERRUPTION || !shouldRetry(method, e)) {
```
jeffret-b: How about moving that first part (before the `||`) into the `shouldRetry()` method?

jeffret-b:
For anyone that sees this problem, is there any way you could test this PR and see if it makes a nice improvement? A bonus would be to see if there is any noticeable performance degradation.

src/main/java/hudson/remoting/RemoteClassLoader.java
thomasgl-orange: 505ad3b4

src/main/java/hudson/remoting/RemoteClassLoader.java
thomasgl-orange: 505ad3b4 - I've extracted it as `static RemoteInterruptedExceptionCatcher.shouldBeEnabled()`. This way, all references to `MAX_RETRIES_ON_REMOTING_INTERRUPTION` are in `RemoteInterruptedExceptionCatcher`.

jeffret-b: I like it.

thomasgl-orange: Regarding this `RemoteClassLoader` constructor, we could refactor a little bit more and extract more static methods, something like that:

```java

    private static IClassLoader enhanceDumbClassLoader(@CheckForNull Channel channel, @Nonnull IClassLoader proxy) {
        if (channel == null || !channel.remoteCapability.supportsPrefetch() || channel.getJarCache() == null) {
            return new DumbClassLoaderBridge(proxy);
        } else {
            return proxy;
        }
    }

    private static IClassLoader addRemoteInterruptedExceptionCatcher(@Nonnull IClassLoader proxy) {
        if (RemoteInterruptedExceptionCatcher.shouldBeEnabled()) {
            // JENKINS-61103: wrap the target (IClassLoader) proxy in a dynamic proxy which will handle retries in case
            // of interruption exception from the remoting system during class loading/initialization.
            // It should take care of cases like JENKINS-36991, JENKINS-51854 and others.
            return (IClassLoader) Proxy.newProxyInstance(RemoteClassLoader.class.getClassLoader(),
                    new Class[] { IClassLoader.class }, new RemoteInterruptedExceptionCatcher(proxy));
        } else {
            return proxy;
        }
    }

    private RemoteClassLoader(@CheckForNull ClassLoader parent, @Nonnull IClassLoader proxy) {
        super(new URL[0],parent);
        final Channel channel = RemoteInvocationHandler.unwrap(proxy);
        this.channel = channel == null ? null : channel.ref();
        this.underlyingProxy = proxy;
        this.proxy = addRemoteInterruptedExceptionCatcher(enhanceDumbClassLoader(channel, proxy));
    }
```

Or maybe the same but keeping an intermediate `targetProxy` variable:

```java
        final IClassLoader targetProxy = enhanceDumbClassLoader(channel, proxy);
        this.proxy = addRemoteInterruptedExceptionCatcher(targetProxy);
```

I don't really have a preference, you let me know if you have one.

jeffret-b:
Not sure what's going on with the build. Maybe a Windows infra issue. Let's kick it again and see if it's random.

src/main/java/hudson/remoting/RemoteClassLoader.java
thomasgl-orange: 505ad3b4

The expression in `shouldRetry` is becoming a bit long maybe, do you think it's time split into `if / else if` with intermediate `return false`?

src/main/java/hudson/remoting/RemoteClassLoader.java
jeffret-b: I thought about some of those as I was reading through it and think it could help clarify what's going on. I don't know that it's necessary but it could be nice.

src/main/java/hudson/remoting/RemoteClassLoader.java
jeffret-b: I would give that a try. It is getting kind of long and cumbersome.

jeffret-b:
Remaining areas where I'm concerned:

1) Can we determine this improves things before we release it? If not, we can probably still just release it as an improvement, but it would be nice to have some info.

2) It would be good to have a test for the retry maximum.

3) The CI failure. It builds fine on my machine. We need to get the infrastructure straightened out or whatever the problem is. I'll see if I can figure anything out in the next couple of days.

jeffret-b:
Try a rebuild.

jeffret-b:
Builds work again.

I've been busy but I'd like to get this integrated. Any comments on my latest?

src/test/java/hudson/remoting/ClassRemotingTest.java
```diff
@@ -187,12 +210,16 @@ public Object call() throws Exception {
...
+        private InterruptNthInvocation(int n) {
+            countDown = n;
+        }
         @Override
         public void run() {
-            invocationCount++;
-            if (invocationCount == 3) {
+            countDown--;
+            if (countDown == 0) {
+                System.out.println("## interrupt");
```
jeffret-b: I don't think we need to keep this println.

jeffret-b:
I tried tossing together a test for the retry maximum. Something isn't working quite the way I expected it to. It might just be a test artifact but I'm not sure yet. I'm more convinced that we need a test for retries and the maximum.

It may take me a while before I can get back to digging into that, but I'd be happy to review if someone can get something put together.

jeffret-b:
I've put a bunch of work into this issue the past few days. I'm more confident now that we need to address an interrupt in the loading of resources. That's a missing piece that can cause problems. Fixing it may help in a number of situations, though we lack good information on exactly what those situations are.

I have discovered that some of the existing tests aren't working correctly. Some pass without really testing the right thing. That made it more difficult to create reliable tests for a change like this. I've figured out how to make them work correctly, with some changes.

I've struggled, though, to get the tests to work correctly with this PR. There are some things going on here that complicate things more. It also has the potential to impact more behaviors of the remote class loader, which could possibly introduce other problems. I like this centralized approach, but the impacts to tests and other possible behaviors concern me.

I've been putting together another proposal, based upon cleaning up and modernizing the class and then doing another implementation of retries for resources, based on the existing ones. I've got it working along with several tests. I'll probably start putting together a couple of PRs for it in the coming days.

Any other proposals or ideas are welcome. It could be possible to continue with this approach with sufficient tests and improvements.

thomasgl-orange:
Closing in favor of #379.  Thanks @jeffret-b for the time you've spent reviewing this proposal, and implementing the alternative PR!

PR #376, opened 2020-04-16
Cleanup and modernization for RemoteClassLoader

jeffret-b:
Some simple modernization and cleanup for RemoteClassLoader. These changes are all refactorings with no change in behavior.

Once I get this approved and merged I've got a couple more cleanup PRs to process and then some actual improvements to RemoteClassLoader.

### DIFF
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -92,8 +92,8 @@ final class RemoteClassLoader extends URLClassLoader {
      */
     private /*mostly final*/ Channel.Ref channel;
 
-    private final Map<String,URLish> resourceMap = new HashMap<String,URLish>();
-    private final Map<String,Vector<URLish>> resourcesMap = new HashMap<String,Vector<URLish>>();
+    private final Map<String,URLish> resourceMap = new HashMap<>();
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -322,7 +323,6 @@ ClassReference toRef(ClassFile2 cf) {
                                     // but we need to remember to set the interrupt flag back on
                                     // before we leave this call.
                                     interrupted = true;
-                                    continue;   // JENKINS-19453: retry
```
varyvol, line 325: Isn't this a behaviour change?

jeffret-b: I don't think so. As IntelliJ says, "'continue' is unnecessary as the last statement in a loop ".

Because of the depth of the loops it's hard to read and the GH UI here doesn't make it any easier.

jvz: Looks like a redundant `continue` as it's the end of the block.

varyvol: Makes sense, thanks for the clarification!

PR #379, opened 2020-04-21
[JENKINS-61103] Retry on class resource load failures and introduce timeouts

jeffret-b:
See [JENKINS-61103](https://github.com/jenkinsci/remoting/pull/372). This is an alternative approach to #372.

There have been a couple of previous efforts to introduce retries into the RemoteClassLoader. These have reportedly resolved some situations, however we have continued to receive reports of class loading failures. As noted in JENKINS-61103 and the earlier PR, one area that isn't covered by a retry is resource loading as part of a class.

This PR does three things:
1. Fixes some of the tests so that they work correctly. Some of them weren't really testing what they claimed to. Some could be simplified a little.
2. Introduces timeouts and sleep to the retries. This should slow down further attempts that may quickly fail and provide a way to eventually terminate if the problem doesn't resolve. I add this to the two existing retry areas and to the new one. I arbitrarily chose 10 minutes as the total timeout time. It's much smaller than the previous value (infinite), but still a significant amount of time.
3. Adds retries to `findResource()` using the same pattern.

These changes only get involved in truly exceptional conditions. I've never been able to reproduce them directly. Other reports of class loading failures have similarly lacked for reproducibility. This has two major impacts:
1. It's very difficult to test how much this fix will impact the situation, how well it improve things, or whether it will correct any of the reported situations.
2. Any problems introduced will not have widespread effect. If the retry period is not ideal, it won't affect the great majority of situations. This makes me feel like it is worthwhile to try out this change to see how well it helps.

This change might help for [JENKINS-51854](https://issues.jenkins-ci.org/browse/JENKINS-51854) and [JENKINS-514910](https://issues.jenkins-ci.org/browse/JENKINS-514910).

### DIFF
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -294,17 +294,19 @@ final class RemoteClassLoader extends URLClassLoader {
                         return rcl.loadClassFile(name, proxy.fetch(name));
                     } catch (IOException x) {
                         throw new ClassNotFoundException(name, x);
-                    } catch (InterruptedException x) {
-                        // pretend as if this operation is not interruptible.
-                        // but we need to remember to set the interrupt flag back on
-                        // before we leave this call.
-                        interrupted = true;
-                        try {
-                            rcl.getClassLoadingLock(name).wait(RETRY_SLEEP_DURATION_MILLISECONDS);
-                        } catch (InterruptedException e) {
-                            // Not much to do if we can't sleep. Run through the tries more quickly.
+                    } catch (InterruptedException | RemotingSystemException x) {
+                        if (x instanceof InterruptedException | x.getCause() instanceof InterruptedException) {
+                            // pretend as if this operation is not interruptible.
+                            // but we need to remember to set the interrupt flag back on
+                            // before we leave this call.
+                            interrupted = true;
+                            try {
+                                rcl.getClassLoadingLock(name).wait(RETRY_SLEEP_DURATION_MILLISECONDS);
+                            } catch (InterruptedException e) {
+                                // Not much to do if we can't sleep. Run through the tries more quickly.
+                            }
+                            LOGGER.finer("Handling interrupt while loading remote class. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
                         }
-                        LOGGER.finer("Handling interrupt while loading remote class. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
                     }
                  }
                 throw new ClassNotFoundException("Could not load class " + name + " after " + MAX_RETRIES + " tries.");
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -72,8 +72,38 @@ import static java.util.logging.Level.WARNING;
 final class RemoteClassLoader extends URLClassLoader {
 
     private static final Logger LOGGER = Logger.getLogger(RemoteClassLoader.class.getName());
-    static int SLEEP_DURATION_MS = 100;
-    static int MAX_RETRIES = 10 * 60 * 1000 / SLEEP_DURATION_MS; // 10 minutes
+
+    /**
+     * Intercept {@link RemoteClassLoader#loadRemoteClass(String, Channel, ClassReference, RemoteClassLoader)} for unit tests.
+     * See JENKINS-6604 and similar issues.
+     * Should not be used for any other purpose.
+     */
+    static Runnable TESTING_CLASS_LOAD;
+    /**
+     * Intercept {@link RemoteClassLoader#prefetchClassReference(String, Channel)} for unit tests.
+     * Should not be used for any other purpose.
+     */
+    static Runnable TESTING_CLASS_REFERENCE_LOAD;
+    /**
+     * Intercept {@link RemoteClassLoader#findResource(String)} for unit tests.
+     * Should not be used for any other purpose.
+     */
+    static Runnable TESTING_RESOURCE_LOAD;
+
+    /**
+     * The amount of time to sleep before retrying an interrupted class load.
+     * This sleep keeps it from hammering the channel if there is a failure.
+     * The default value is 100 (ms).
+     */
+    static int RETRY_SLEEP_DURATION_MILLISECONDS = Integer.getInteger(RemoteClassLoader.class.getName() + "retrySleepDurationMilliseconds", 100);
+    /**
+     * The total number of retries for an interrupted class load.
+     * This makes the operation retry for an extended period of time but eventually timeout.
+     * Combined with the default value for RETRY_SLEEP_DURATION_MILLISECONDS this gives a default
+     * timeout of 10 minutes, which is much less than the former (infinite) retry but still a significant
+     * amount of time.
+     */
+    static int MAX_RETRIES = Integer.getInteger(RemoteClassLoader.class.getName() + "maxRetries", 10 * 60 * 1000 / RETRY_SLEEP_DURATION_MILLISECONDS);
 
     /**
      * Proxy to the code running on remote end.
@@ -277,7 +307,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         // before we leave this call.
                         interrupted = true;
                         try {
-                            rcl.getClassLoadingLock(name).wait(SLEEP_DURATION_MS);
+                            rcl.getClassLoadingLock(name).wait(RETRY_SLEEP_DURATION_MILLISECONDS);
                         } catch (InterruptedException e) {
                             // Not much to do if we can't sleep. Run through the tries more quickly.
                         }
@@ -367,7 +397,7 @@ final class RemoteClassLoader extends URLClassLoader {
                             // before we leave this call.
                             interrupted = true;
                             try {
-                                Thread.sleep(SLEEP_DURATION_MS);
+                                Thread.sleep(RETRY_SLEEP_DURATION_MILLISECONDS);
                             } catch (InterruptedException e) {
                                 // Not much to do if we can't sleep. Run through the tries more quickly.
                             }
@@ -392,14 +422,6 @@ final class RemoteClassLoader extends URLClassLoader {
         }
         return cr;
     }
-
-    /**
-     * Intercept {@link RemoteClassLoader#findClass(String)} to allow unit tests to be written.
-     * <p>
-     * See JENKINS-6604 and similar issues
-     */
-    static Runnable TESTING_CLASS_LOAD;
-    static Runnable TESTING_CLASS_REFERENCE_LOAD;
 
     /**
      * Loads class from the byte array.
@@ -459,6 +481,7 @@ final class RemoteClassLoader extends URLClassLoader {
     }
 
     @CheckForNull
+    @SuppressFBWarnings(value = "DLS_DEAD_LOCAL_STORE", justification = "False positive -- spotbugs mistake.")
     public URL findResource(String name) {
         // first attempt to load from locally fetched jars
         URL url = super.findResource(name);
@@ -467,39 +490,72 @@ final class RemoteClassLoader extends URLClassLoader {
             return url;
         }
 
+        boolean interrupted = false;
         try {
-            if (resourceMap.containsKey(name)) {
-                URLish f = resourceMap.get(name);
-                if (f == null) {
-                    return null;    // no such resource
+            int tries = 0;
+            while (tries < MAX_RETRIES) {
+                tries++;
+                try {
+                    if (resourceMap.containsKey(name)) {
+                        URLish f = resourceMap.get(name);
+                        if (f == null) {
+                            return null;    // no such resource
+                        }
+                        URL u = f.toURL();
+                        if (u != null) {
+                            return u;
+                        }
+                    }
+
+                    if (TESTING_RESOURCE_LOAD != null) {
+                        TESTING_RESOURCE_LOAD.run();
+                    }
+
+                    long startTime = System.nanoTime();
+
+                    ResourceFile r = proxy.getResource2(name);
+                    ResourceImageRef image = null;
+                    if (r != null) {
+                        image = r.image;
+                    }
+
+                    channel.resourceLoadingTime.addAndGet(System.nanoTime() - startTime);
+                    channel.resourceLoadingCount.incrementAndGet();
+                    if (image == null) {
+                        resourceMap.put(name, null);
+                        return null;
+                    }
+
+                    URLish res = image.resolveURL(channel, name).get();
+                    resourceMap.put(name, res);
+                    return res.toURL();
+                } catch (IOException | InterruptedException | ExecutionException e) {
+                    throw new Error("Unable to load resource " + name, e);
+                } catch (RemotingSystemException x) {
+                    if (x.getCause() instanceof InterruptedException) {
+                        // pretend as if this operation is not interruptible.
+                        // but we need to remember to set the interrupt flag back on
+                        // before we leave this call.
+                        interrupted = true;
+                        try {
+                            Thread.sleep(RETRY_SLEEP_DURATION_MILLISECONDS);
+                        } catch (InterruptedException e) {
+                            // Not much to do if we can't sleep. Run through the tries more quickly.
+                        }
+                        continue;   // JENKINS-19453: retry
+                    }
+                    throw x;
                 }
-                URL u = f.toURL();
-                if (u != null) {
-                    return u;
-                }
+
+                // no code is allowed to reach here
             }
-
-            long startTime = System.nanoTime();
-
-            ResourceFile r = proxy.getResource2(name);
-            ResourceImageRef image = null;
-            if (r != null) {
-                image = r.image;
+        } finally {
+            // process the interrupt later.
+            if (interrupted) {
+                Thread.currentThread().interrupt();
             }
-
-            channel.resourceLoadingTime.addAndGet(System.nanoTime() - startTime);
-            channel.resourceLoadingCount.incrementAndGet();
-            if (image == null) {
-                resourceMap.put(name, null);
-                return null;
-            }
-
-            URLish res = image.resolveURL(channel, name).get();
-            resourceMap.put(name, res);
-            return res.toURL();
-        } catch (IOException | InterruptedException | ExecutionException e) {
-            throw new Error("Unable to load resource " + name, e);
         }
+        throw new RuntimeException("Could not load resource " + name + " after " + MAX_RETRIES + " tries.");
     }
 
     /**
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -275,12 +275,7 @@ final class RemoteClassLoader extends URLClassLoader {
                 while (tries < MAX_RETRIES) {
                     tries++;
                     try {
-                        if (TESTING_CLASS_LOAD != null) {
-                            TESTING_CLASS_LOAD.run();
-                            if (Thread.currentThread().isInterrupted()) {
-                                throw new InterruptedException("loading was interrupted.");
-                            }
-                        }
+                        invokeClassLoadTestingHookIfNeeded();
 
                         if (c != null) {
                             return c;
@@ -322,6 +317,17 @@ final class RemoteClassLoader extends URLClassLoader {
         }
     }
 
+    private void invokeClassLoadTestingHookIfNeeded() throws InterruptedException {
+        // Testing support only.
+        if (TESTING_CLASS_LOAD != null) {
+            TESTING_CLASS_LOAD.run();
+            if (Thread.currentThread().isInterrupted()) {
+                // Otherwise the interrupt isn't recognized and the test doesn't work.
+                throw new InterruptedException("loading was interrupted.");
+            }
+        }
+    }
+
     private ClassReference prefetchClassReference(String name, Channel channel) throws ClassNotFoundException {
         ClassReference cr;
         cr = prefetchedClasses.remove(name);
@@ -339,9 +345,7 @@ final class RemoteClassLoader extends URLClassLoader {
                 while (tries < MAX_RETRIES) {
                     tries++;
                     try {
-                        if (TESTING_CLASS_REFERENCE_LOAD != null) {
-                            TESTING_CLASS_REFERENCE_LOAD.run();
-                        }
+                        invokeClassReferenceLoadTestingHookIfNeeded();
 
                         Map<String, ClassFile2> all = proxy.fetch3(name);
                         synchronized (prefetchedClasses) {
@@ -421,6 +425,13 @@ final class RemoteClassLoader extends URLClassLoader {
             channel.classLoadingPrefetchCacheCount.incrementAndGet();
         }
         return cr;
+    }
+
+    private void invokeClassReferenceLoadTestingHookIfNeeded() {
+        // Testing support only.
+        if (TESTING_CLASS_REFERENCE_LOAD != null) {
+            TESTING_CLASS_REFERENCE_LOAD.run();
+        }
     }
 
     /**
@@ -507,9 +518,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         }
                     }
 
-                    if (TESTING_RESOURCE_LOAD != null) {
-                        TESTING_RESOURCE_LOAD.run();
-                    }
+                    invokeResourceLoadTestingHookIfNeeded();
 
                     long startTime = System.nanoTime();
 
@@ -542,7 +551,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         } catch (InterruptedException e) {
                             // Not much to do if we can't sleep. Run through the tries more quickly.
                         }
-                        continue;   // JENKINS-19453: retry
+                        continue;
                     }
                     throw x;
                 }
@@ -556,6 +565,13 @@ final class RemoteClassLoader extends URLClassLoader {
             }
         }
         throw new RuntimeException("Could not load resource " + name + " after " + MAX_RETRIES + " tries.");
+    }
+
+    private void invokeResourceLoadTestingHookIfNeeded() {
+        // Testing support only.
+        if (TESTING_RESOURCE_LOAD != null) {
+            TESTING_RESOURCE_LOAD.run();
+        }
     }
 
     /**
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -306,6 +306,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         } catch (InterruptedException e) {
                             // Not much to do if we can't sleep. Run through the tries more quickly.
                         }
+                        LOGGER.finer("Handling interrupt while loading remote class. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
                     }
                  }
                 throw new ClassNotFoundException("Could not load class " + name + " after " + MAX_RETRIES + " tries.");
@@ -405,6 +406,7 @@ final class RemoteClassLoader extends URLClassLoader {
                             } catch (InterruptedException e) {
                                 // Not much to do if we can't sleep. Run through the tries more quickly.
                             }
+                            LOGGER.finer("Handling interrupt while fetching class reference. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
                             continue;   // JENKINS-19453: retry
                         }
                         throw x;
@@ -551,6 +553,7 @@ final class RemoteClassLoader extends URLClassLoader {
                         } catch (InterruptedException e) {
                             // Not much to do if we can't sleep. Run through the tries more quickly.
                         }
+                        LOGGER.finer("Handling interrupt while finding resource. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
                         continue;
                     }
                     throw x;
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -271,9 +271,7 @@ final class RemoteClassLoader extends URLClassLoader {
                 // and just retry until it succeeds, but in the end we set the interrupt flag
                 // back on to let the interrupt in the next earliest occasion.
 
-                int tries = 0;
-                while (tries < MAX_RETRIES) {
-                    tries++;
+                for (int tries = 0; tries < MAX_RETRIES; tries++) {
                     try {
                         invokeClassLoadTestingHookIfNeeded();
 
@@ -342,9 +340,7 @@ final class RemoteClassLoader extends URLClassLoader {
                 // and just retry until it succeeds, but in the end we set the interrupt flag
                 // back on to let the interrupt in the next earliest occasion.
 
-                int tries = 0;
-                while (tries < MAX_RETRIES) {
-                    tries++;
+                 for (int tries = 0; tries < MAX_RETRIES; tries++) {
                     try {
                         invokeClassReferenceLoadTestingHookIfNeeded();
 
@@ -494,7 +490,6 @@ final class RemoteClassLoader extends URLClassLoader {
     }
 
     @CheckForNull
-    @SuppressFBWarnings(value = "DLS_DEAD_LOCAL_STORE", justification = "False positive -- spotbugs mistake.")
     public URL findResource(String name) {
         // first attempt to load from locally fetched jars
         URL url = super.findResource(name);
@@ -505,9 +500,7 @@ final class RemoteClassLoader extends URLClassLoader {
 
         boolean interrupted = false;
         try {
-            int tries = 0;
-            while (tries < MAX_RETRIES) {
-                tries++;
+            for (int tries = 0; tries < MAX_RETRIES; tries++) {
                 try {
                     if (resourceMap.containsKey(name)) {
                         URLish f = resourceMap.get(name);
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -74,22 +74,26 @@ final class RemoteClassLoader extends URLClassLoader {
 
     private static final Logger LOGGER = Logger.getLogger(RemoteClassLoader.class.getName());
 
+    interface Interruptible {
+        void run() throws InterruptedException;
+    }
+
     /**
      * Intercept {@link RemoteClassLoader#loadRemoteClass(String, Channel, ClassReference, RemoteClassLoader)} for unit tests.
      * See JENKINS-6604 and similar issues.
      * Should not be used for any other purpose.
      */
-    static Runnable TESTING_CLASS_LOAD;
+    static Interruptible TESTING_CLASS_LOAD;
     /**
      * Intercept {@link RemoteClassLoader#prefetchClassReference(String, Channel)} for unit tests.
      * Should not be used for any other purpose.
      */
-    static Runnable TESTING_CLASS_REFERENCE_LOAD;
+    static Interruptible TESTING_CLASS_REFERENCE_LOAD;
     /**
      * Intercept {@link RemoteClassLoader#findResource(String)} for unit tests.
      * Should not be used for any other purpose.
      */
-    static Runnable TESTING_RESOURCE_LOAD;
+    static Interruptible TESTING_RESOURCE_LOAD;
 
     /**
      * The amount of time to sleep before retrying an interrupted class load.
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -239,11 +271,11 @@ along with the reference to the initiating ClassLoader (if the initiating ClassL
                 // and just retry until it succeeds, but in the end we set the interrupt flag
                 // back on to let the interrupt in the next earliest occasion.
 
-                while (true) {
+                int tries = 0;
+                while (tries < MAX_RETRIES) {
```
jvz, line 275: You really don't like `for` loops, do you? ;P

jeffret-b: I'm not sure why I used a while loop here instead of a for loop. Maybe I thought the while was more expressive of the intended early exit? Maybe because it was already a while? It probably would be better to convert to for loops.

jeffret-b:
> If this turns out to be helpful, perhaps this can be enhanced later to use a more generic retry strategy?

Could be. I'm trying to keep it understandable and not too much of a change. There are other patterns that could be used, including the approach in the earlier PR. I couldn't get that one to work out right, including with the tests.

I'd also like to investigate some retries on channel failures, some sort of automated reconnection. That seems to be an issue that comes up fairly frequently. Might be kind of complicated.

jeffret-b:
I'm hoping to get a few more reviews before proceeding on this, so please take a look.

src/main/java/hudson/remoting/RemoteClassLoader.java
jvz: I was just thinking back to the `for (foo = bar; foo != null; foo = foo.bar)` loop, and now I saw a normal for loop not using the syntax, so I thought you might have a thing :P

jvz:
For retrying, there's always https://github.com/resilience4j/resilience4j thought I don't know how heavy that is.

basil:
> For retrying, there's always https://github.com/resilience4j/resilience4j thought I don't know how heavy that is.

[Failsafe](https://jodah.net/failsafe/) is another option. Since we're on the subject, I love [Tenacity's documentation](https://tenacity.readthedocs.io/) on retrying. It so clearly and concisely explains the pros and cons of various retrying strategies.

jvz:
> > For retrying, there's always https://github.com/resilience4j/resilience4j thought I don't know how heavy that is.
> 
> [Failsafe](https://jodah.net/failsafe/) is another option. Since we're on the subject, I love [Tenacity's documentation](https://tenacity.readthedocs.io/) on retrying. It so clearly and concisely explains the pros and cons of various retrying strategies.

Failsafe looks really neat!

jeffret-b:
Thanks for the comments and reviews. As Matt mentioned earlier, I'm trying to keep this change relatively simple for now. Those libraries might be useful later. At a minimum, maybe we can start iterating towards more information on failures and reproducible cases.

thomasgl-orange:
I have done a few manual tests of a 4.4-SNAPSHOT built from this PR (with a patched Jenkins 2.222.1), using the procedure I had described in [JENKINS-61103](https://issues.jenkins-ci.org/browse/JENKINS-61103):
- disconnect the agent
- delete its jar cache (not sure it really matters, I've just assumed it would give me more opportunities to interrupt class loading while doing remoting stuff)
- reconnect it
- run a simple Maven job (here I'm running the "clean" target on the jenkinsci/remoting repository, but it doesn't matter)
- interrupt the job when it starts talking about the POM

After a few tries, I've reproduced a `LinkageError` caused by a `RemotingSystemException` of `InterruptedException`. Following attempts at running Maven job then fail with a `NoClassDefFoundError`, for the reasons described in t.

* Log of the interrupted build:
```
[...]
23:43:16  > git checkout -f 142a5ef53ef80df3bf7b0f10ca9e555238ec33e3 # timeout=10
23:43:18 Commit message: "Merge pull request #378 from jeffret-b/updateRCL"
23:43:18  > git rev-list --no-walk 142a5ef53ef80df3bf7b0f10ca9e555238ec33e3 # timeout=10
23:43:20 Parsing POMs
23:43:20 using global settings config with name FaaS network settings
23:43:20 Replacing all maven server entries not found in credentials list is true
23:43:23 Build was aborted
23:43:23 Aborted by Thomas De Grenier De La Tour
23:43:23 Started calculate disk usage of build
23:43:23 Finished Calculation of disk usage of build in 0 seconds
23:43:23 Started calculate disk usage of workspace
23:43:24 Stop LogSizeTimerTask
23:43:24 Finished Calculation of disk usage of workspace in  1 second
23:43:24 [WS-CLEANUP] Deleting project workspace...
23:43:25 [WS-CLEANUP] done
23:43:25 Finished: ABORTED
```

* Exception in the agent log:
```
Apr 22, 2020 11:43:23 PM hudson.remoting.UserRequest perform
WARNING: LinkageError while performing UserRequest:hudson.maven.MavenModuleSetBuild$PomParser@4480bf1e
java.lang.ExceptionInInitializerError
	at org.codehaus.plexus.DefaultPlexusContainer.<init>(DefaultPlexusContainer.java:182)
	at org.codehaus.plexus.DefaultPlexusContainer.<init>(DefaultPlexusContainer.java:168)
	at hudson.maven.MavenEmbedderUtils.buildPlexusContainer(MavenEmbedderUtils.java:166)
	at hudson.maven.MavenEmbedderUtils.buildPlexusContainer(MavenEmbedderUtils.java:159)
	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:110)
	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:137)
	at hudson.maven.MavenUtil.createEmbedder(MavenUtil.java:211)
	at hudson.maven.MavenModuleSetBuild$PomParser.invoke(MavenModuleSetBuild.java:1323)
	at hudson.maven.MavenModuleSetBuild$PomParser.invoke(MavenModuleSetBuild.java:1126)
	at hudson.FilePath$FileCallableWrapper.call(FilePath.java:3069)
	at hudson.remoting.UserRequest.perform(UserRequest.java:211)
	at hudson.remoting.UserRequest.perform(UserRequest.java:54)
	at hudson.remoting.Request$2.run(Request.java:368)
	at hudson.remoting.InterceptingExecutorService$1.call(InterceptingExecutorService.java:72)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
Caused by: hudson.remoting.RemotingSystemException: java.lang.InterruptedException
	at hudson.remoting.RemoteInvocationHandler.invoke(RemoteInvocationHandler.java:301)
	at com.sun.proxy.$Proxy5.fetch(Unknown Source)
	at hudson.remoting.RemoteClassLoader.loadRemoteClass(RemoteClassLoader.java:294)
	at hudson.remoting.RemoteClassLoader.loadWithMultiClassLoader(RemoteClassLoader.java:257)
	at hudson.remoting.RemoteClassLoader.findClass(RemoteClassLoader.java:216)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at org.eclipse.sisu.inject.Weak.concurrentKeys(Weak.java:89)
	at org.eclipse.sisu.inject.Weak.concurrentKeys(Weak.java:79)
	at org.eclipse.sisu.plexus.ClassRealmManager.<clinit>(ClassRealmManager.java:66)
	... 18 more
Caused by: java.lang.InterruptedException
	at java.lang.Object.wait(Native Method)
	at hudson.remoting.Request.call(Request.java:176)
	at hudson.remoting.RemoteInvocationHandler.invoke(RemoteInvocationHandler.java:288)
	... 27 more
```

* Log of the next build attempt:
```
[...]
23:45:04  > git checkout -f 142a5ef53ef80df3bf7b0f10ca9e555238ec33e3 # timeout=10
23:45:04 Commit message: "Merge pull request #378 from jeffret-b/updateRCL"
23:45:04  > git rev-list --no-walk 142a5ef53ef80df3bf7b0f10ca9e555238ec33e3 # timeout=10
23:45:04 Parsing POMs
23:45:04 using global settings config with name FaaS network settings
23:45:04 Replacing all maven server entries not found in credentials list is true
23:45:04 ERROR: Failed to parse POMs
23:45:04 java.io.IOException: Remote call on faas-tmp-thomas-21158-2ezsv225r failed
23:45:04 	at hudson.remoting.Channel.call(Channel.java:1004)
23:45:04 	at hudson.FilePath.act(FilePath.java:1069)
23:45:04 	at hudson.FilePath.act(FilePath.java:1058)
23:45:04 	at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.parsePoms(MavenModuleSetBuild.java:987)
23:45:04 	at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.doRun(MavenModuleSetBuild.java:691)
23:45:04 	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:504)
23:45:04 	at hudson.model.Run.execute(Run.java:1856)
23:45:04 	at hudson.maven.MavenModuleSetBuild.run(MavenModuleSetBuild.java:543)
23:45:04 	at hudson.model.ResourceController.execute(ResourceController.java:97)
23:45:04 	at hudson.model.Executor.run(Executor.java:428)
23:45:04 Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.eclipse.sisu.plexus.ClassRealmManager
23:45:04 	at org.codehaus.plexus.DefaultPlexusContainer.<init>(DefaultPlexusContainer.java:182)
23:45:04 	at org.codehaus.plexus.DefaultPlexusContainer.<init>(DefaultPlexusContainer.java:168)
23:45:04 	at hudson.maven.MavenEmbedderUtils.buildPlexusContainer(MavenEmbedderUtils.java:166)
23:45:04 	at hudson.maven.MavenEmbedderUtils.buildPlexusContainer(MavenEmbedderUtils.java:159)
23:45:04 	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:110)
23:45:04 	at hudson.maven.MavenEmbedder.<init>(MavenEmbedder.java:137)
23:45:04 	at hudson.maven.MavenUtil.createEmbedder(MavenUtil.java:211)
23:45:04 	at hudson.maven.MavenModuleSetBuild$PomParser.invoke(MavenModuleSetBuild.java:1323)
23:45:04 	at hudson.maven.MavenModuleSetBuild$PomParser.invoke(MavenModuleSetBuild.java:1126)
23:45:04 	at hudson.FilePath$FileCallableWrapper.call(FilePath.java:3069)
23:45:04 	at hudson.remoting.UserRequest.perform(UserRequest.java:211)
23:45:04 	at hudson.remoting.UserRequest.perform(UserRequest.java:54)
23:45:04 	at hudson.remoting.Request$2.run(Request.java:368)
23:45:04 	at hudson.remoting.InterceptingExecutorService$1.call(InterceptingExecutorService.java:72)
23:45:04 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
23:45:04 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
23:45:04 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
23:45:04 	at java.lang.Thread.run(Thread.java:748)
23:45:04 	Suppressed: hudson.remoting.Channel$CallSiteStackTrace: Remote call to faas-tmp-thomas-21158-2ezsv225r
23:45:04 		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1788)
23:45:04 		at hudson.remoting.UserRequest$ExceptionResponse.retrieve(UserRequest.java:356)
23:45:04 		at hudson.remoting.Channel.call(Channel.java:998)
23:45:04 		at hudson.FilePath.act(FilePath.java:1069)
23:45:04 		at hudson.FilePath.act(FilePath.java:1058)
23:45:04 		at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.parsePoms(MavenModuleSetBuild.java:987)
23:45:04 		at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.doRun(MavenModuleSetBuild.java:691)
23:45:04 		at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:504)
23:45:04 		at hudson.model.Run.execute(Run.java:1856)
23:45:04 		at hudson.maven.MavenModuleSetBuild.run(MavenModuleSetBuild.java:543)
23:45:04 		at hudson.model.ResourceController.execute(ResourceController.java:97)
23:45:04 		at hudson.model.Executor.run(Executor.java:428)
23:45:05 Started calculate disk usage of build
23:45:05 Finished Calculation of disk usage of build in 0 seconds
23:45:06 Started calculate disk usage of workspace
23:45:06 Finished Calculation of disk usage of workspace in 0 seconds
23:45:06 [WS-CLEANUP] Deleting project workspace...
23:45:06 [WS-CLEANUP] Skipped based on build state FAILURE
23:45:06 Finished: FAILURE
```

That's one difference with the approach I had taken in #372 (the dynamic proxy would have retried the interrupted `fetch` call - although, as mentioned in that PR, I had no unit test to prove it, because it was not obvious how to even reach that line).

Of course, this case can be fixed with your approach too. In `master`, this `RemotingSystemException` would come from here:
[RemoteClassLoader.java#L264](https://github.com/jenkinsci/remoting/blob/master/src/main/java/hudson/remoting/RemoteClassLoader.java#L264)
There is some try/catch/retry logic (that's one you've modified in this PR), but it only applies in case of plain `InterruptedException`, not in case of "`RemotingSystemException` of `InterruptedException`". Catching the right exception there might be enough.


One more other random thought regarding this PR vs #372: with this PR come proper unit tests of the retry logic itself. You can simulate repeated interruptions, because the try/catch/retry blocks are at high level in some `RemoteClassLoader` methods, "around" the test plugs (the `TESTING_CLASS_LOAD` and `TESTING_CLASS_REFERENCE_LOAD` runnables). In #372, I've not been able to find a way to do the same: all that these test plugs can do is set the interrupted flag before a remote class-loading operation, leading to one exception, but I couldn't test more than one retry this way, because it was implemented at a lower level. It would have required a different approach, and I had no clear intuition of what (something at the Channel level? one more layer of proxy?).

jeffret-b:
> There is some try/catch/retry logic (that's one you've modified in this PR), but it only applies in case of plain `InterruptedException`, not in case of "`RemotingSystemException` of `InterruptedException`". Catching the right exception there might be enough.

I wondered about why that catch was different than the others, but I had poor information as to why things might be different or where the failures occurred. I've just extended the catch in that area in my latest commit to also handle RemotingSystemException similar to how it's being done elsewhere. That may give some nice additional protection.

> In #372, I've not been able to find a way to do the same: all that these test plugs can do is set the interrupted flag before a remote class-loading operation, leading to one exception, but I couldn't test more than one retry this way, because it was implemented at a lower level.

I like the cleanness and thoroughness of your approach, but I've struggled with a couple of things with it. I never could get the tests or MAX_RETRIES to work with it. Given the level at which that approach behaves, I wasn't as confident in exactly what it was doing or how much it might apply to other situations. I ended up pursuing this approach because I had more confidence in what it was doing, partly because I was able to get decent tests around it.

thomasgl-orange:
> I've just extended the catch in that area in my latest commit to also handle RemotingSystemException similar to how it's being done elsewhere. That may give some nice additional protection.

Yes, 5c0bec8 should fix it.
I've not tried yet, but does it by chance allow to get rid of the [`throw new InterruptedException` in `invokeClassLoadTestingHookIfNeeded()`](https://github.com/jenkinsci/remoting/pull/379/files#diff-32d87b4ebdfc84c84c9202eee3200a4dR327)?

Reviewing the code, I find a few other call paths from `ClassLoader` public methods to the remoting proxy (`IClassLoader`) with no safe-guard (no try/catch/retry for `RemotingSystemExcpetion(InterruptedException)`), which could thus theoretically lead to similar errors:

* `findClass(String)` => `fetchFromProxy(String, Channel)` => `IClassLoader.fetch(String)`
  * [RemoteClassLoader.java#L218](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L218) => [RemoteClassLoader.java#L225](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L225)
  * only when `!channel.remoteCapability.supportsMultiClassLoaderRPC()`, so probably not worth fixing
* `findClass(String)` => `loadWithMultiClassLoader(String, Channel)` => `IClassLoader.fetch2(String)`
  * [RemoteClassLoader.java#L216](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L216) => [RemoteClassLoader.java#L249](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L249)
  * only when `!channel.remoteCapability.supportsPrefetch()`, so probably not worth fixing
* `findResources(String)` => `IClassLoader.getResources2(String)`
  * [RemoteClassLoader.java#L611](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L611)
  * this one, I think, deserves being safe-guard, in a way similar to `findResource(String)`

Also, I see that [in `findResource(String)`](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L538), there is no retry in case of `InterruptedException`, whereas there are retries for that exception [in `loadRemoteClass(...)`](https://github.com/jeffret-b/remoting/blob/5c0bec8/src/main/java/hudson/remoting/RemoteClassLoader.java#L297) (ie., in `findClass(String)`). It doesn't sound very coherent, what do you think? (To be clear, things where already like that before your PR, and were still like that with my PR too, I was just wondering if that's something you would consider cleaning up.)

And two last thoughts/questions:

* what about handling [`InterruptedIOException`](https://docs.oracle.com/javase/8/docs/api/java/io/InterruptedIOException.html)? That's also an exception which can be thrown when the thread is interrupted, thus similar to an `InterruptedException`, except it is not an `InterruptedException` because it is an `IOException` and Java is Java. I tend to think it makes sense to handle it in a similar way (I was doing that in #372 when it was the cause of a `RemotingSystemException`), but I've never been 100% sure it was right.
* an idea I did mention (but not implement) in #372 was, for resources loading (`findResource`/`findResources`), to only retry in case the call was part of some class initialization (`<clinit>` in the stack). Because the overall goal of this retry mechanism is to avoid errors which could leave the classloader in an unrecoverable state with half-loaded classes, but there's no arm in immediately interrupting resources loading in an other context. What do you think?

> I ended up pursuing this approach because I had more confidence in what it was doing, partly because I was able to get decent tests around it.

Sure, I get that, testability is a big plus here.

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -264,15 +294,22 @@ along with the reference to the initiating ClassLoader (if the initiating ClassL
...
+                            } catch (InterruptedException e) {
+                                // Not much to do if we can't sleep. Run through the tries more quickly.
+                            }
+                            LOGGER.finer("Handling interrupt while loading remote class. Current retry count = " + tries + ", maximum = " + MAX_RETRIES);
+                        }
                     }
-
-                    // no code is allowed to reach here
-                }
+                 }
```
thomasgl-orange, line 311:
```suggestion
                }
```

jeffret-b:
> I've not tried yet, but does it by chance allow to get rid of the [`throw new InterruptedException` in `invokeClassLoadTestingHookIfNeeded()`](https://github.com/jenkinsci/remoting/pull/379/files#diff-32d87b4ebdfc84c84c9202eee3200a4dR327)?

No, that's still needed.

>     * only when `!channel.remoteCapability.supportsMultiClassLoaderRPC()`, so probably not worth fixing

Yes, I see no need to do anything about those. I noticed at least one of those when I was putting my PR together but there's no reason to expend effort there.

>     * findResources(String)

I agree that we should do something about this one. I'll take a look.

(More responses to come.)

jeffret-b:
>     * `findResources(String)` => `IClassLoader.getResources2(String)`

This one should now be covered with the same pattern in [e6ddc55](https://github.com/jenkinsci/remoting/pull/379/commits/e6ddc55a3a38a997b364a4dcd78b0b0321c5a76e).

jeffret-b:
>     * an idea I did mention (but not implement) in #372 was, for resources loading (`findResource`/`findResources`), to only retry in case the call was part of some class initialization (`<clinit>` in the stack). Because the overall goal of this retry mechanism is to avoid errors which could leave the classloader in an unrecoverable state with half-loaded classes, but there's no harm in immediately interrupting resources loading in an other context. What do you think?

On the idea that it's easier to evaluate ideas like this with working code, I threw together an implementation. It's on another branch because I'm not sure what to do with it yet. So far I'm inclined to not merge it in and keep the retries for resource loading generally. When only catching the set of exceptions we're looking at, these are mostly involved with channel operations. We're not likely to catch ones that aren't involved in channel stuff anyway. Regular resource loading problems should continue to fail quickly. I'm not sure the special handling helps, though it probably doesn't hurt, either. See the [branch](https://github.com/jeffret-b/remoting/tree/retryRCLresourcesOnlyClinit) and the [commit diff](https://github.com/jeffret-b/remoting/commit/002a86e259bc79f8cbe6eaf279f7e619108819ce).

jeffret-b:
I'll look at the rest of the comments about the types of exceptions we're handling next week.

jeffret-b:
I added another commit with consistent exception checking for retries. As Thomas mentions earlier the behavior here has long been inconsistent. It's not clear that the same exceptions are exactly relevant in the same ways in each of the different places. There is value, though, in making the handling as consistent as possible. This set of exceptions seem like good ones to handle here with the retries. They will only be relevant in exceptional cases. Each of these seem like good candidates for retry, which will eventually time out. The time out is period is configurable. Given the difficulty in obtaining complete information about the failure cases, it's hard to know for certain which cases are needed where, but I think this set is reasonable.

jeffret-b:
@thomasgl-orange, thanks for all of your great input here and on the previous items. I think I've responded to all of the points you have raised. If I've missed any, please point them out again. I appreciate anything more you're able to review or suggest here.

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -281,6 +319,23 @@ along with the reference to the initiating ClassLoader (if the initiating ClassL
...
     }
 
+    private void invokeClassLoadTestingHookIfNeeded() throws InterruptedException {
+        // Testing support only.
+        if (TESTING_CLASS_LOAD != null) {
+            TESTING_CLASS_LOAD.run();
+            if (Thread.currentThread().isInterrupted()) {
+                // Otherwise the interrupt isn't recognized and the test doesn't work.
+                throw new InterruptedException("loading was interrupted.");
+            }
```
thomasgl-orange, line 329: nit: maybe use a different runnable hook, which would directly throw the exception you need for your test, instead of "hardcoding" the exception here?
We can imagine test cases which would involve setting the interrupted flag here, but without expecting this `InterruptedException` (for instance something which would try to get a `RemotingException` from the `proxy.fetch` call in `loadRemoteClass`).

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -449,38 +507,75 @@ public URL findResource(String name) {
...
+                    channel.resourceLoadingCount.incrementAndGet();
+                    if (image == null) {
+                        resourceMap.put(name, null);
+                        return null;
+                    }
+
+                    URLish res = image.resolveURL(channel, name).get();
+                    resourceMap.put(name, res);
+                    return res.toURL();
+                } catch (IOException | InterruptedException | ExecutionException e) {
```
thomasgl-orange, line 545: move `InterruptedException` to the next catch statement, so that it is handled by the retry logic

jeffret-b: This makes me reconsider the previous change to standardize the exceptions through a single method. Was there some reason InterruptedException was handled the way it was? -- But, I don't have much confidence that the previous exception handling was correct or very intentional. I still think the standardization is probably better.

jeffret-b:
I've added a few recent, notable improvements.

[9b505b3](https://github.com/jenkinsci/remoting/pull/379/commits/9b505b38c12f08b6b57224f1cdb56fee4107a084) This adds more standardization and consistency around the handling of exception types. There's also more of this in [3745555](https://github.com/jenkinsci/remoting/pull/379/commits/3745555d3f2d41351cba59fe77b01b5b96e6c0c6). I'm not certain this results in handling all of the desired exceptions correctly but I'm not convinced the previous behavior was correct. Consistency and standardization are generally better and if nothing else may help us surface any issues that still arise.

[4ef7ad3](https://github.com/jenkinsci/remoting/pull/379/commits/4ef7ad326a1b5da2f5ed4caa39c7d046bdc750c3) improves the clarity and handling of retries. This is based off a number of good suggestions from @thomasgl-orange. One thing this does is restore the capability for an infinite retry.

[d6985ed](https://github.com/jenkinsci/remoting/pull/379/commits/d6985ed50ae191608343fb929d3caede3ac0d4d2) stops managing the interrupted state after successful retries. The more I've gone over this code, this seems the right thing to do. I'm not seeing how it makes sense to reset the interrupted state after we've successfully managed the condition. It also cleans up the code noticeably. (Though the reformat may make it a little harder to review.)

src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -330,8 +332,17 @@ private void invokeClassLoadTestingHookIfNeeded() throws InterruptedException {
...
     }
 
-    private boolean shouldRetry(Throwable e) {
-        return e instanceof InterruptedException || (e instanceof RemotingSystemException
+    private boolean shouldRetry(Throwable e, int tries) {
+        return isRetryException(e) && hasMoreRetries(tries);
+    }
+
+    private boolean hasMoreRetries(int tries) {
+        return MAX_RETRIES <= 0 || tries <= MAX_RETRIES;
```
thomasgl-orange, line 327: I was more expecting `MAX_RETRIES=0` as a way to disable the retry mechanism.

Maybe something like this? `return MAX_RETRIES < 0 || tries <= MAX_RETRIES;`
- positive number: max number of retries
- zero: disable retries
- negative: infinite retries

thomasgl-orange:
> [d6985ed](https://github.com/jenkinsci/remoting/pull/379/commits/d6985ed50ae191608343fb929d3caede3ac0d4d2) stops managing the interrupted state after successful retries. The more I've gone over this code, this seems the right thing to do. I'm not seeing how it makes sense to reset the interrupted state after we've successfully managed the condition.

I have mixed feelings about this one.

There must be a reason why the thread got interrupted in the first place, and imho it would be wrong to ignore it simply because the thread happens to be currently loading classes and class loading is not the right time for being interrupted. It would mean that, for the same code remaining to execute after `interrupt()` got called on a thread, you could "randomly":
- either never get interrupted (no visible `InterruptedException`), because some classes were yet to be loaded first, and the `RemoteClassLoader` as swallowed the interruption,
- or get interrupted eventually, because all required classes were already loaded (or loading them didn't trigger an `InterruptedException` because they were in some cache), and thus the actual interruptible code eventually got executed with the thread still being in interrupted state.

It doesn't seem very coherent, and it's hard to tell what the consequences might be in practice.

On the other hand, if the original interruption was caused by the class loading process itself, then yes, once we've finally managed to load classes in a retry, there is no point setting the thread to interrupted state. One such case I can think of would be catching a `RemotingSystemException` of `SocketTimeoutException` (which is itself an `InterruptedIOException`): the interruption comes from the fact that we were loading classes over a network channel, so it's an "internal" cause, and in this case I agree it's not desirable to re-interrupt the thread after we've recovered.

thomasgl-orange:
>     1. It's very difficult to test how much this fix will impact the situation, how well it improve things, or whether it will correct any of the reported situations.

Not an definitive answer to this difficult question, but if it can help, I've wrote an automated test procedure which repeatedly (with a moving timing):
- launches a simple Maven job on a `DumbSlave` agent
- aborts its execution
- looks for a `LinkageError` trace in the agent log
- when found, launches the job again, to check that the agent class loader is effectively "corrupted" (and in practice, yes, the 2nd build will always fail)

It's here: https://github.com/thomasgl-orange/jenkins/commit/117aeb5d18c87089ff1f05814ac82bdca90ed329

So, it's similar to my manual test procedure (explained here in a comment somewhere, and in [JENKINS-61103](https://issues.jenkins-ci.org/browse/JENKINS-61103), but automated.

It's written as a `jenkins-test` JUnit test (of course, not one I would suggest for inclusion, but it was convenient to write that there). In my experiments, with the settings which are currently in the test (tuned for my 10 years old laptop I guess), the test is repeated 80 times, leading to (on average) 6 `LinkageError`.  The same with `remoting` `4.4-rc2787.5dbdde7d611d` (incremental build from this PR) instead of `4.2` never triggers any `LinkageError`, which confirm that this PR does improves something.

jeffret-b:
@thomasgl-orange , your thoughts about not promulgating the interrupt are consistent with mine. I'm not really sure it's a good idea, but if the the exception was handled by the retry and things were taken care of, it doesn't seem valuable to pass the interrupt on up. But, if it were a meaningful interrupt it would be meaningful to process. One consideration I see is that, especially with more consistency with what exceptions we're handling we're are setting that interrupted flag and then we're setting the thread as interrupted for things that weren't interrupts. It's hard to say.

Also, I'm very glad to see your test setups to demonstrate that these changes really do improve some situations. That's important confirmation that these changes are valuable. There are still some details to finalize, like the interrupt state but it would be good to get this in.

jeffret-b:
Thanks for everyone who has reviewed this and provided comments and suggestions. It's been very helpful and I'm convinced there are valuable improvements here.

Unfortunately, my time is becoming much more limited and I don't know when I'll be able to devote enough concentrated time to moving this forward, finalizing the remaining questions, and monitoring the results. I'll move it back to a draft state and see when I or someone else can get enough time to complete it.

thehustler088:
@jeffret-b @thomasgl-orange I am seeing similar issue in my environment for which I have opened new defect- https://issues.jenkins.io/browse/JENKINS-64182  can I test with draft changes? if yes, may i know the steps to build remoting jar with my jenkins.

jeffret-b:
@thehustler088, it would be great if you could test this out and see if it helps in your environment. I've resolved the conflicts and updated it to the latest master state.

You can obtain the agent.jar file for this incremental build at https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/remoting/4.7-rc2878.23d1ba3e3b46/

I pushed a companion PR in Jenkins core to incorporate this incremental in jenkinsci/jenkins#5054. Once it builds, it should produce an incremental that incorporates this special Remoting build. 

This is all experimental at this point.

jeffret-b:
@thehustler088, you can get an incrementals build of the Jenkins war at https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/jenkins-war/2.267-rc30573.4207aa6dd06a/

thehustler088:
Thanks @jeffret-b This build resolved the issues I am seeing earlier and working as expected in my environment when I abort running jobs. No exceptions / issues seen in remoting logs.

thehustler088:
@jeffret-b on some noes i see the below error: 
Nov 11, 2020 12:52:04 AM hudson.util.ProcessTree get
WARNING: Error while determining if vetoers exist
java.lang.InterruptedException
	at java.base/java.lang.Object.wait(Native Method)
	at hudson.remoting.Request.call(Request.java:177)
	at hudson.remoting.Channel.call(Channel.java:1000)
	at hudson.util.ProcessTree.get(ProcessTree.java:434)
	at hudson.Launcher$RemoteLauncher$KillTask.call(Launcher.java:1164)
	at hudson.Launcher$RemoteLauncher$KillTask.call(Launcher.java:1155)
	at hudson.remoting.UserRequest.perform(UserRequest.java:211)
	at hudson.remoting.UserRequest.perform(UserRequest.java:54)
	at hudson.remoting.Request$2.run(Request.java:376)
	at hudson.remoting.InterceptingExecutorService.lambda$wrap$0(InterceptingExecutorService.java:78)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)

jeffret-b:
Hmmm ... this PR includes some changes in how interrupts are handled. Looks like one now bubbles up in a different way. 

Do you see any indication that this causes a problem? (Besides the stack trace. Maybe we can just handle it differently if it doesn't cause any actual issues.)

thehustler088:
@jeffret-b I haven't seen any issues, all looks good except above traces on some nodes. does increasing max retries would avoid hitting these exceptions?

jeffret-b:
This sounds like a specific case that needs to be handled in the PR. I'll see if I can scrape together enough time in the next day or two to try and see how to handle it. I doubt increasing the max retries would change it, but you can always give it a try.

thehustler088:
@jeffret-b I running bunch of python scripts in background that does deployment and testing in my environment.

jeffret-b:
@thehustler088, I've gotten another incrementals build of the Jenkins war that may cover the follow-on issue: https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/jenkins-war/2.268-rc30582.2308bd563998/ . If you can give that a try and report on the results that would help.

thehustler088:
Sure @jeffret-b I may not able to load it today since weekend runs are in progress. I will load it on Sunday, I will let you the test results by Monday / Tuesday. Thanks!

thehustler088:
@jeffret-b I have loaded war file(https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/jenkins-war/2.268-rc30582.2308bd563998/ ) in my env, did not see any issues / traces  till now. I tried on 3 nodes(running ubuntu 18.04 LTS) by interrupting the running jobs. I may try on different nodes this week

thehustler088:
Is there way to make these draft changes(remoting jar) compatible with LTS versions, as I see some UI bugs with the current loaded build.?

jeffret-b:
@thehustler088, thanks for the testing! This gives more confidence that this change improves the situation, which will help push it forward.

The core build uses the incrementals capability, which requires it to be based on the latest state of the master branch. This can indeed bring in unstable changes. If you want to have a different version you can build  jenkinsci/jenkins#5054 on top of whatever branch you want. I don't know where I would host a build like that for you.

thehustler088:
@jeffret-b Thanks for the changes. I am trying to build piplines job for my new requirement with 2.268-rc30582.2308bd563998 and UI becomes unmanageable with active choice parameters. Is there a way to make remoting jar(which includes your draft changes) compatible with jenkins - Jenkins 2.263.1? I did not find a way to make it work.

Also, may I know when are these changes will be officially available with Jenkins?

jeffret-b:
@thehustler088, I cherry-picked the two relevant commits from jenkinsci/jenkins#5054 on top of the jenkinsci/jenkins branch `stable-2.263`. Then I built with skipping tests. I've uploaded the resulting war file to https://www.dropbox.com/s/1ez4t5pxp30q8tw/jenkins.war?dl=0 . Not sure if you can access that, but you can give it a try or build yourself.

I have no plan yet for merging these in. If you report success in your environment that gives more confidence. I could see about integrating these in sometime in January 2021.

thehustler088:
@jeffret-b Thanks much for building on stable branch. I have downloaded it, will load it sometime tomorrow. If this solves UI issues, I will load this build on multiple master servers.

thehustler088:
@jeffret-b I have been using this build for about week and I did not notice any issues with interrupting the running jobs. Also, UI is working fine with active choice parameters.

jeffret-b:
Thanks for testing that. The results are very promising. I'll try to get this moving forward early next year.

In the meantime, if anyone else wants to review or especially test, that would be great.

thehustler088:
@jeffret-b any progress on pushing the changes to official jenkins release

jeffret-b:
@thehustler088 , I've been busy, which may continue through the end of this month, but I want to get this done as soon as I can.

jeffret-b:
I'm moving this officially out of draft and work-in-progress. There are a number of reviews. Some people have tested and verified improvements using this reworked approach.

I'd like to get this merged and released soon. I appreciate any other reviews or tests. I'd like to get it merged in and ready for release within the next couple of days.

jeffret-b:
@thehustler088 , this is now released in Remoting 4.7. I'm not going to push it into the Jenkins packaging immediately. Any testing or use results you can share will help increase confidence and speed of getting it packaged into core.

I'm hoping to get it into a weekly Jenkins release within the next couple of releases.

Prabhu088:
@jeffret-b Thanks much for pushing the changes to official remoting 4.7. May I know what version of jenkins is compatible with remoting jar 4.7?  can I use with  Jenkins 2.263.1? If yes, please let me know what branches to be pulled in order to build Jenkins war 2.263.1 with Remoting 4.7

jeffret-b:
This Remoting version should work fine with older versions, especially including 2.263.x. Even older versions should still work fine, though they get little testing. Newer features may not work -- for example, Jenkins 2.217 is required for WebSocket capabilities.

Prabhu088:
@jeffret-b I have added 4.7 remoting jar in Jenkins 2.263.1, I will load and will let you know by end of this week

thehustler088:
I have loaded this in couple of my testing machine running ubuntu 18.04lts and cent os 7, did not see any issue / exceptions when I abort running jobs.

PR #383, opened 2020-06-09
Ping repeatedly from ping thread

cpt1gl0:
Ping repeatedly from ping thread before assuming error to avoid unnecessary node disconnects.

### DIFF
--- a/src/main/java/hudson/remoting/PingThread.java
+++ b/src/main/java/hudson/remoting/PingThread.java
@@ -48,9 +48,9 @@ import java.util.logging.Level;
  * @since 1.170
  */
 public abstract class PingThread extends Thread {
-    private static final long     DEFAULT_TIMEOUT      = TimeUnit.MINUTES.toMillis(1);
-    private static final long     DEFAULT_INTERVAL     = TimeUnit.MINUTES.toMillis(10);
-    private static final unsigned DEFAULT_MAX_TIMEOUTS = 4;
+    private static final long  DEFAULT_TIMEOUT      = TimeUnit.MINUTES.toMillis(1);
+    private static final long  DEFAULT_INTERVAL     = TimeUnit.MINUTES.toMillis(10);
+    private static final int   DEFAULT_MAX_TIMEOUTS = 4;
 
     private final Channel channel;
 
@@ -68,9 +68,9 @@ public abstract class PingThread extends Thread {
     /**
      * Tolerate max timeouts before assuming ping error.
      */
-    private final unsigned maxTimeouts;
+    private final int maxTimeouts;
 
-    public PingThread(Channel channel, long timeout, long interval, unsigned maxTimeouts) {
+    public PingThread(Channel channel, long timeout, long interval, int maxTimeouts) {
         super("Ping thread for channel "+channel);
         this.channel = channel;
         this.timeout = timeout;
@@ -121,7 +121,7 @@ public abstract class PingThread extends Thread {
     private void ping() throws IOException, InterruptedException {
         LOGGER.log(Level.FINE, "pinging {0}", channel.getName());
 
-        unsigned timeouts = 0;
+        int timeouts = 0;
 
         while (true) {
             Future<?> f = channel.callAsync(new Ping());
--- a/src/main/java/hudson/remoting/PingThread.java
+++ b/src/main/java/hudson/remoting/PingThread.java
@@ -48,6 +48,8 @@ import java.util.logging.Level;
  * @since 1.170
  */
 public abstract class PingThread extends Thread {
+    private final static int DEFAULT_MAX_TIMEOUTS = 3;
+
     private final Channel channel;
 
     /**
@@ -61,7 +63,8 @@ public abstract class PingThread extends Thread {
      */
     private final long interval;
 
-    private final static int DEFAULT_MAX_TIMEOUTS = 3;
+    private final int maxTimeouts;
+    private int timeouts = 0;
 
     public PingThread(Channel channel, long timeout, long interval, int maxTimeouts) {
         super("Ping thread for channel "+channel);
@@ -89,25 +92,20 @@ public abstract class PingThread extends Thread {
             while(true) {
                 long nextCheck = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(interval);
 
-                int timeouts = 0;
-                while (true) {
-                    try {
-                        ping();
-                        break;
-                    } catch (TimeoutException e) {
-                        if (++timeouts >= this.maxTimeouts) {
-                            onDead(e);
-                            break;
-                        } else {
-                            LOGGER.log( Level.WARNING,
-                                        "timeout {0}/{1} pinging {2}, retrying...",
-                                        new Object[]
-                                        {timeouts, this.maxTimeouts, channel.getName()} );
-                        }
-                    } catch (ExecutionException e) {
+                try {
+                    ping();
+                    this.timeouts = 0;
+                } catch (TimeoutException e) {
+                    if (++this.timeouts >= this.maxTimeouts) {
                         onDead(e);
-                        break;
+                    } else {
+                        LOGGER.log( Level.WARNING, "timeout {0}/{1} pinging {2}",
+                                    new Object[] {this.timeouts,
+                                                  this.maxTimeouts,
+                                                  channel.getName()} );
                     }
+                } catch (ExecutionException e) {
+                    onDead(e);
                 }
 
                 // wait until the next check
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/PingThread.java
```diff
@@ -121,16 +142,16 @@ private void ping() throws IOException, InterruptedException {
...
+                throw e;
             } catch (TimeoutException e) {
                 // get method waits "at most the amount specified in the timeout",
                 // so let's make sure that it really waited enough
             }
             remaining = end - System.nanoTime();
         } while(remaining>0);
 
-        onDead(new TimeoutException("Ping started at "+start+" hasn't completed by "+System.currentTimeMillis()));//.initCause(e)
+        throw new TimeoutException( "Ping started at " + start +
```
afalko, line 153: Maybe cleaner to do a String.format here

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -61,11 +63,14 @@
      */
     private final long interval;
 
-    public PingThread(Channel channel, long timeout, long interval) {
+    private final int maxTimeouts;
+
+    public PingThread(Channel channel, long timeout, long interval, int maxTimeouts) {
```
afalko, line 68: Instead of changing a public constructor, create a new one and set your default within the old one. I fear other plugins might use the current constructor you are replacing.

jeffret-b:
Try a rebuild after infrastructure issues.

jeffret-b:
I've wanted to look into making this ping thread more resilient. Changing it can be dangerous because there are some odd behaviors associated with it. 

Thanks for the submission. I'll try to find some time to take a look at it.

cpt1gl0:
Re-implemented repeated ping, addressing the mentioned issues and trying to be less invasive:

* Re-ping 4 times with timeout of 1 minute to mimic old behaviour
* Keep old constructors
* Moved retry logic to ping function
* Use string format instead of joining strings

jeffret-b:
Looks like there's a compilation failure.

jeffret-b:
Let's try a CI rebuild to see where we currently are with this.

jeffret-b:
Yes, there is a remaining compilation failure.

@cpt1gl0 , are you interested in getting back to finish this up?

res0nance:
@cpt1gl0 I hope you don't mind, I took the liberty of resolving the simple issues.

src/main/java/hudson/remoting/PingThread.java
jeffret-b: Looks like this has been resolved.

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -106,31 +120,45 @@ public void run() {
...
-                return;
-            } catch (ExecutionException e) {
-                if (e.getCause() instanceof RequestAbortedException)
-                    return; // connection has shut down orderly.
-                onDead(e);
-                return;
-            } catch (TimeoutException e) {
-                // get method waits "at most the amount specified in the timeout",
-                // so let's make sure that it really waited enough
+        while (true) {
```
jeffret-b, line 127: Seems like this would be better expressed as a `for` loop.

jeffret-b (commented):
The code looks good, but the overall proposal is incomplete. Changes to the ping thread need a Jira ticket, even if they seem to maintain the existing behavior. It would be good to have more information on what this solves and how it will be used.

It may still make sense to accept this change, but it would help to have better background information.

src/main/java/hudson/remoting/PingThread.java
cpt1gl0: Has been addressed in latest commit. In fact it was possible to entirely get rid of the inner loop.

cpt1gl0:
> The code looks good, but the overall proposal is incomplete. Changes to the ping thread need a Jira ticket, even if they seem to maintain the existing behavior. It would be good to have more information on what this solves and how it will be used.
> 
> It may still make sense to accept this change, but it would help to have better background information.

It is intended to solve issues where a node is disconnected because of a single failing ping.

jeffret-b:
@cpt1gl0, thanks for getting back to this. Builds are passing now so we're in better shape.

Could you clarify a few things about your experience with this change, please?

Are you observing a problem that is fixed by this PR? In other words, are you experiencing an issue "where a node is disconnected because of a single failing ping"? If so, is this PR correcting the problem?

(It will help in accepting this PR if we have a confirmed fix and more testing, though we can still consider it as an improvement without that.)

Could you submit a Jira about the problem this addresses and link it to this PR?

cpt1gl0:
@jeffret-b I just can tell you that at my company, where we use Jenkins as an integral part of our build environment, we experience problems with disconnecting nodes when the build system is under heavy load. Usually those disconnections are caused by failing pings from ping thread. From that perspective, I think it would make sense to make the ping thread more fault tolerant, and allow for single failing pings without disconnect.

jeffret-b:
@cpt1gl0, have you been able to try out your improvement, perhaps in a test environment, to see that it improves the situation? If we can get some real-world success with this change I'm less concerned about the risk of introducing it.

cpt1gl0:
@jeffret-b I can not give you any hard facts. Personally, I think it's a good idea to introduce that change, but it is for you to decide.

jeffret-b:
I'll see about getting this in some future release. If anyone else would like to review or even better try it out that be very appreciated.

jtnord:
> we experience problems with disconnecting nodes when the build system is under heavy load. Usually those disconnections are caused by failing pings from ping thread

Normally the ping thread is the symptom not the cause.
If you do sable the ping thread entirely (system property so easy to test) does that help you.  If not the changes here are likely to be beneficial.

olivergondza:
I agree it is desirable to have it confirmed that a thread reporting a ping failure will actually recover and later operates fine.

cpt1gl0:
> Normally the ping thread is the symptom not the cause.
A totally agree and I think this is not for discussion.

In our case usually a system under heavy load is the original cause, and yes, a failing ping thread is the symptom. For me it's not a question of cause and effect, but more about reasonable measures to take. Does it make sense to disconnect a node after a single ping failure? What's the gain? Why not be a bit more fault tolerant and leave the possibility for the system to recover without any failing builds? If pings keep failing the node would still be disconnected eventually and possible issues will not go unnoticed. Personally I would prefer a system which does not immediately chop off one of its legs because of a numb toe.

daniel-beck:
> without any failing builds

Not the topic, but still: Consider using Pipeline.

jeffret-b:
@cpt1gl0 , if you have any experience using this change that results in better stability or results, that would help to move it forward. I agree that it would be better to be more resilient.

felipecrs:
I don't know, but perhaps this could help https://issues.jenkins.io/browse/JENKINS-50730

jeffret-b:
> I don't know, but perhaps this could help https://issues.jenkins.io/browse/JENKINS-50730

@felipecrs In what way does that help? Or how is that issue related to the ping thread?

felipecrs:
@jeffret-b, I'm sorry, thinking about it now, I think it was a mistake from my side. Please disregard.

NorseGaud:
The PR seems to be solving a problem we see as well.

NorseGaud:
@cpt1gl0 Did you want me to open a jenkins ticket for this? Or could you?

jglick:
> Why not be a bit more fault tolerant and leave the possibility for the system to recover without any failing builds?

To expand on https://github.com/jenkinsci/remoting/pull/383#issuecomment-835891416, a Pipeline build should recover automatically despite a node disconnection. There are three cases:

1. The agent disconnects while running `sh` (or `bat` or `powershell`) but the agent machine itself is fine. The agent should reconnect automatically, and the `sh` step should proceed, including with any output printed during the outage, even if the forked process completed during that time.
2. Same but during some other “non-durable” step such as `checkout scm`. That step will fail, but you can arrange for the build to recover automatically
   ```groovy
   retry(conditions: [nonresumable()]) {
     node(…) {
       checkout scm
       // as before
     }
   }
   ```
3. The agent machine actually crashed. In that case of course this PR would be irrelevant, but mentioning for completeness; you can still arrange for the build to recover automatically (even across controller restarts as of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180):
   ```groovy
   retry(conditions: [agent(), nonresumable()]) {
     node(…) {
       // as before
     }
   }
   ```

Given these resilience modes, and the fact that you can already configure a longer ping timeout, I am not sure what is gained by retrying a failed ping.

henzef:
Is anyone still working on this? I think the connection issues I am seeing in my setup would be solved by this.

PR #81, opened 2016-04-13
[FIXED JENKINS-34213] Ensure that the unexporter cleans up whatever it can each sweep

stephenc:
@reviewbybees

PR #392, opened 2020-08-20
RemoteInvocationHandler.reportStats LogRecord’s could get mangled during serialization

jglick:
See https://github.com/jenkinsci/jenkins/pull/4621 for the original buglet. As of https://github.com/jenkinsci/jenkins/pull/4912, seems better to just avoid the problematic idiom.

### DIFF
--- a/src/main/java/hudson/remoting/RemoteInvocationHandler.java
+++ b/src/main/java/hudson/remoting/RemoteInvocationHandler.java
@@ -729,7 +729,7 @@ final class RemoteInvocationHandler implements InvocationHandler, Serializable {
                                             + "pressure."
                                     : "The all time average rate is %.1f±%.1f/sec. "
                                                     + "The 15 minute average rate is %.1f±%.1f/sec. "
-                                                    + "At the 95% confidence level both are above 100.0/sec. ",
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -719,63 +718,63 @@ private void reportStats() {
             if (m15Std > 1 && 100 < m15Avg - 2 * m15Std) {
                 if (tStd > 1 && 100 < tAvg - 2 * tStd) {
                     logger.log(Level.SEVERE,
```
jglick, line 720: These are some sort of exotic conditions with no test coverage that I know of, so just hoping it works.

jglick: see #81

jeffret-b:
Builds fine on my system. Let's see if we can get a build on CI.

jeffret-b (approved):
Looks good. Hopefully it will work well enough.

String.format() can be slow, but we shouldn't be doing it enough here to cause performance problems.

PR #393, opened 2020-08-21
[JENKINS-63118] delete /tmp/resource-xxx on exit

thomasgl-orange:
See [JENKINS-63118](https://issues.jenkins-ci.org/browse/JENKINS-63118)

In Util.makeResource(), call deleteDirectoryOnExit() on the temporary
directory, and not just on the resource file it contains.

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -73,7 +73,7 @@ public class Util {
             fos.write(image);
         }
 
-        deleteDirectoryOnExit(resource);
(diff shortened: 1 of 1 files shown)


jeffret-b:
Builds fine on my system. Let's see if we can get a build on CI.

MarkEWaite:
Builds fine on my Windows system and does not report any failures.

jeffret-b (approved):
The change looks good.

I haven't been able to verify that it actually fixes anything, though. The fix here doesn't quite match the originally submitted bug description. It doesn't break anything and should make some improvement so I'll go ahead and include it.

PR #397, opened 2020-10-13
Use Warning-NG instead of findbugs.

Vlatombe:
Should fix the build.


oleg-nenashev (commented):
We should rather move it to Warnings NG, see https://github.com/jenkinsci/jenkinsfile-runner/blob/master/Jenkinsfile for an example

jglick (approved):

PR #412, opened 2020-10-15
Upgrade bouncy castle

jeffret-b:
Upgrade both components together in one PR to preserve master as buildable.

Supersedes the dependabot PRs #404 and #400.


jvz (approved):
If you refactor this into a maven property, then dependabot will upgrade it properly in the future.

jeffret-b:
Let's try another build.

PR #417, opened 2020-10-16
Improve documentation, removing more instances of 'slave' term.

jeffret-b:
This cleans up a few more instances of incorrect terminology in documentation.

This only fully makes sense with jenkinsci/jenkins#35452, but I consider the two to be fully independent. Without that change in core, the file name in the documentation here doesn't exist, but that's a minor concern as things get updated and coordinated. Old and new agents work fine with or without the core change.


jeffret-b:
Kick a rebuild.

jvz (approved):

PR #422, opened 2020-11-03
Undeprecate KeyUtils.fingerprint

jglick:
Contrary to the conversation in #361, this method was already used in `remoting`, and it is used in https://github.com/jenkinsci/jenkins/pull/5044 since that was basically a copy & inline of this method.


jeffret-b (approved):

res0nance (approved):

jglick:
As a follow-up, probably we should redeprecate it, stop using it from core, and switch the core instance identity fingerprint to use, say, SHA-256.

PR #428, opened 2020-12-21
Collection of various minor cleanups.

jeffret-b:
Another round of minor cleanups across the Remoting project.

Correct misspellings. Enhance annotations. Remove some unused pieces that are not part of the public API.


src/main/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerFactory.java
```diff
@@ -92,7 +87,9 @@ public JnlpProtocolHandlerFactory(@Nonnull ExecutorService threadPool) {
      * @return {@code this} for method chaining.
      */
     public JnlpProtocolHandlerFactory withNioChannelHub(@CheckForNull NioChannelHub nioChannelHub) {
-        this.nioChannelHub = nioChannelHub;
+        /**
+         * The {@link NioChannelHub} to use or {@code null}
+         */
```
res0nance, line 92: Is this supposed to be removed with no effect?

jvz (approved):
Very nice cleanup! As a side note, it'd be really neat if we could extract the generic networking/buffering code into its own library (think something like a trimmed down Netty or Mina) as it could potentially allow for more reuse of that code (and more eyes testing it). On the other hand, we also have Stapler which shows that separating out libraries isn't always useful.

src/main/java/org/jenkinsci/remoting/engine/JnlpProtocolHandlerFactory.java
jeffret-b: @res0nance, as Matt comments, the nioChannelHub field isn't used. I haven't dug into the details of when this might have been used, though I'm not certain it ever was. The value is set on this line, but nothing ever uses it. As part of this PR, I've retained the public API, but removed the unused internal pieces.

jeffret-b:
@jvz, I don't think there's much value in a separate library containing the generic networking/buffering code. What would be nice is separating that code out (essentially removing it) and then replacing it with a proven, reliable messaging system. Remoting over Kafka is one such example.

PR #435, opened 2021-02-05
Remove hardcode of JKS for use other key stores

gulyaev13:
I suggest removing the hardcode with JKS Keystore to allow using other Keystores. For example, BCFKS provided in the FIPS version of Bouncy Castle.
More info here:
https://groups.google.com/g/jenkinsci-dev/c/YkEG_oSR1mw

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -483,7 +483,7 @@ public class Engine extends Thread {
                 char[] password = "password".toCharArray();
                 KeyStore store;
                 try {
-                    store = KeyStore.getInstance("JKS");
+                    store = KeyStore.getInstance(KeyStore.getDefaultType());
                 } catch (KeyStoreException e) {
                     throw new IllegalStateException("Java runtime specification requires support for JKS key store", e);
                 }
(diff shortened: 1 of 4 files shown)


jeffret-b:
At a preliminary look, this makes sense. I'll need to spend more time when I can to evaluate it.

jeffret-b:
Let's try a rebuild to see if we can get a clean build.

gulyaev13:
> Let's try a rebuild to see if we can get a clean build.

@jeffret-b, looks like everything is ok.

jeffret-b (approved):
Makes sense and works fine in my limited testing.

gulyaev13:
@jeffret-b , do you have any plans for release 4.7 with my changes?

jeffret-b:
Soon. At least within a week. Probably towards the start of this coming week.

I'm not sure about when it will actually be packaged into Jenkins though. Given the changes involved in this and everything else going on in Jenkins changes at the moment, it may sit a while. This Remoting release can be used with older or existing Jenkins installations. That can enable more testing or experience with these changes, which can in turn help motivate inclusion into the Jenkins war.

gulyaev13:
Thanks for your response.
I want to use the new version of remoting for build images with the JNLP agent. It isn't related to Jenkins Core.
https://github.com/jenkinsci/docker-agent

I'm creating the same docker agent images but based on our own base image.

jeffret-b:
@gulyaev13 , this is now released in Remoting 4.7. I'm not going to push it into the Jenkins packaging immediately. Any testing or use results you can share will help increase confidence and speed of getting it packaged into core.

PR #438, opened 2021-02-11
Undo the Dependabot proposed update for the Websocket library.

jeffret-b:
It causes related failures in Jenkins tests.
We can wait for the next update or try later to resolve the failures.
At this point, I just don't want to have this added complexity in this release.

(I could have just the revert feature, but since the change got in before other changes, I decided to do a regular PR to keep a record.)


jvz (approved):

jeffret-b:
Need to make sure we have a clean build.

PR #441, opened 2021-02-17
Disable stacktraces from ExportTable entries by default

res0nance:
Not certain if this is a good idea but the idea spawned from [JENKINS-63975](https://issues.jenkins.io/browse/JENKINS-63975) as stacktraces seem to take a considerable amount of memory usage in Jenkins and as far as I understand it, it is entirely for debugging.

Although I'm not certain how often this workflow is used in the real world, if its very rare this ever happens perhaps its better to disable by default.

### DIFF
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -81,7 +81,7 @@ final class ExportTable {
         /**
          * Where was this object first exported?
          */
-        @Nonnull
+        @CheckForNull
         final CreatedAt allocationTrace;
         /**
          * Where was this object unexported?
@@ -100,7 +100,7 @@ final class ExportTable {
             this.interfaces = interfaces.clone();
             this.object = object;
             this.objectType = object.getClass().getName();
-            this.allocationTrace = new CreatedAt();
+            this.allocationTrace = EXPORT_TRACES ? new CreatedAt() : null;
 
             table.put(id,this);
             reverse.put(object,this);
@@ -144,10 +144,11 @@ final class ExportTable {
                 reverse.remove(object);
 
                 object = null;
-                releaseTrace = new ReleasedAt(callSite);
-
+                if (EXPORT_TRACES) {
+                    releaseTrace = new ReleasedAt(callSite);
+                }
                 unexportLog.add(this);
-                while (unexportLog.size()>UNEXPORT_LOG_SIZE)
+                while (unexportLog.size() > UNEXPORT_LOG_SIZE)
                     unexportLog.remove(0);
             }
         }
@@ -168,8 +169,10 @@ final class ExportTable {
          */
         void dump(PrintWriter w) throws IOException {
             w.printf("#%d (ref.%d) : object=%s type=%s interfaces=%s%n", id, referenceCount, object, objectType, interfaceNames());
-            allocationTrace.printStackTrace(w);
-            if (releaseTrace!=null) {
+            if (allocationTrace != null) {
+                allocationTrace.printStackTrace(w);
+            }
+            if (releaseTrace != null) {
                 releaseTrace.printStackTrace(w);
             }
         }
@@ -424,8 +427,10 @@ final class ExportTable {
 
         if (!unexportLog.isEmpty()) {
             for (Entry<?> e : unexportLog) {
-                if (e.id==id)
-                    cause = new Exception("Object was recently deallocated\n"+Util.indent(e.dump()), e.releaseTrace);
+                if (e.id==id) {
+                    cause = new Exception("Object was recently deallocated\n" + Util.indent(e.dump()), e.releaseTrace);
+                    break;
+                }
             }
             if (cause==null) {
                 // If there is no cause available, create an artificial cause and use the last unexport entry as an estimated release time if possible
@@ -485,6 +490,8 @@ final class ExportTable {
      * @since 2.40
      */
     public static int UNEXPORT_LOG_SIZE = Integer.getInteger(ExportTable.class.getName()+".unexportLogSize",1024);
+
+    static boolean EXPORT_TRACES = Boolean.getBoolean(ExportTable.class.getName() + "exportTraces");
 
     private static final Logger LOGGER = Logger.getLogger(ExportTable.class.getName());
 }
--- a/src/main/java/hudson/remoting/ExportTable.java
+++ b/src/main/java/hudson/remoting/ExportTable.java
@@ -491,7 +491,7 @@ final class ExportTable {
      */
     public static int UNEXPORT_LOG_SIZE = Integer.getInteger(ExportTable.class.getName()+".unexportLogSize",1024);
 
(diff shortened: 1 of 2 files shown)


jeffret-b:
This looks like a good idea. I don't think we get much value from retaining these stack traces. I'll have to look at it more deeply.

basil:
While I have not used it myself, this infrastructure smells like what Brendan Gregg would call a [crisis tool](http://www.youtube.com/watch?v=fhBHvsi0Ql0&t=11m11s): a tool which you would need to get to root cause in a crisis situation and which you would not be able to easily install or enable in that same crisis situation. If such tools are not installed or enabled by default, then these crises often pass without getting to root cause. Installing or enabling the crisis tool after the fact may not help if the problem is not easily reproducible, or if inducing another failure for the purposes of root cause analysis is prohibitively expensive. The philosophy of debugging from first failure is best served by having such crisis tools installed and enabled by default.

res0nance:
> While I have not used it myself, this infrastructure smells like what Brendan Gregg would call a [crisis tool](http://www.youtube.com/watch?v=fhBHvsi0Ql0&t=11m11s): a tool which you would need to get to root cause in a crisis situation and which you would not be able to easily install or enable in that same crisis situation. If such tools are not installed or enabled by default, then these crises often pass without getting to root cause. Installing or enabling the crisis tool after the fact may not help if the problem is not easily reproducible, or if inducing another failure for the purposes of root cause analysis is prohibitively expensive. The philosophy of debugging from first failure is best served by having such crisis tools installed and enabled by default.

That was an informative watch, although in this case compared to having tools installed this has a cost associated in terms of memory which directly affects scalability. In some controllers StacktraceElement can be one of the largest consumers of memory. One of my instances has it pegged at the second highest consumer of memory. Depending on how useful this information is to remoting perhaps an argument can be made to disable it by default.

jeffret-b:
Yes, I'm sure this was intended as some sort of debugging tool in problematic situations. But, I don't know that there was really a justification for it or someone years ago thought it might be useful sometime. I'm not personally aware of it every being useful.

Most of the failures associated with Remoting these days originate from outside -- system, environment, or networking issues that cause the channel to close. Or weird plugin interactions that end up looking like something in Remoting. The stack traces  from inside rarely provide interesting information in these cases.

Unless we can come up with a meaningful scenario where these stack traces help, removing them would be better given the cost.

basil:
>I'm not personally aware of it ever being useful.

That doesn't necessarily mean this functionality is pointless, just that nobody has used it recently (which could be for any number of reasons). If this functionality is now pointless and has not been recently used, then it would indeed be clearly desirable to disable it by default or even to remove it entirely. If this functionality still has a legitimate purpose but has not been recently used, then making it harder to use during crisis situations would be a small regression in debuggability. Whether that small regression is worth the gain in scalability seems hard to quantify without fully understanding the original use case, something I don't personally understand and which isn't explained in this PR.

I think it would be helpful to research the original use case from which this functionality originated and to reason explicitly about whether or not that use case remains legitimate, regardless of whether or not anyone has used this functionality recently. Ideally a solution could be found that improves scalability without decreasing debuggability. It is unclear to me whether such a solution exists, but I think it would be worth thinking about. At minimum, going through this thought exercise would give us more clarity about the cost (to be weighed against the obvious scalability benefit).

If we have looked for such a solution but have not found one, then we have to choose which of the two is more important in this situation: scalability or debuggability. That decision is in part based on the cost-benefit calculus described above but also comes down to the community's [values](https://www.slideshare.net/bcantrill/platform-as-reflection-of-values-joyent-nodejs-and-beyond). I personally value debuggability over scalability, but I recognize that I am on the periphery of this community and that the community's values may not exactly match my own (which is fine).

jeffret-b:
This diagnostic capability was added 7 years ago in an attempt to gather more information for tracking down some specific information. I'm not seeing any indication that it was helpful in diagnosing the problem or that the problem has retained any importance. While it's hard to say that it no longer exists, I can't find any further reports. There are only 4 people watching it.

This PR does include a way to reintroduce the logging, if it is needed. Given the thin history of this capability, even that may be excessive.

Given the cost of this feature and the very dubious value, it's probably best to at least disable it.

jeffret-b (approved):
I'm inclined to go with this cleanup. I haven't seen cases where this information has been useful. If it does prove useful somewhere this change includes a flag to turn it back on.

jeffret-b:
Additional reviews and opinions on this one are definitely welcome. I've taken basil's into account and am still inclined to accept it as mentioned in my approval.

oleg-nenashev (commented):
It should be documented as a system property. No objections w.r.t disabling it by default, but maybe we need some means to enable it in runtime for the controller and agents. Otherwise debugging will become even more complicated

Will request reviews from @jenkinsci/core to get more feedback

src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -486,5 +491,7 @@ synchronized void dump(@Nonnull PrintWriter w) throws IOException {
      */
     public static int UNEXPORT_LOG_SIZE = Integer.getInteger(ExportTable.class.getName()+".unexportLogSize",1024);
 
+    static boolean EXPORT_TRACES = Boolean.getBoolean(ExportTable.class.getName() + "exportTraces");
```
jglick, line 494:
```suggestion
    static boolean EXPORT_TRACES = Boolean.getBoolean(ExportTable.class.getName() + ".exportTraces");
```

src/test/java/hudson/remoting/ExportTableTest.java
```diff
@@ -11,6 +11,7 @@
  */
 public class ExportTableTest extends TestCase {
     public void testDiagnosis() throws Exception {
+        ExportTable.EXPORT_TRACES = true;
```
jglick, line 14: Restore previous value in a `finally` block, or `@After` etc. (See `FlagRule` for JTH-based tests.)

jglick (approved):
Probably OK. If heap rather than CPU is the main issue, might be preferable to retain traces for recently created entries, but I am not sure if that would even be appropriate.

timja (commented):
lgtm apart from Jesse's comments

timja (approved):

jeffret-b:
Looks like this has received enough approvals so I marked it as ready-to-merge. Other reviews are still welcome. We'll plan on getting this into an upcoming release.

oleg-nenashev (approved):
Looks good

src/main/java/hudson/remoting/ExportTable.java
```diff
@@ -486,5 +491,7 @@ synchronized void dump(@Nonnull PrintWriter w) throws IOException {
      */
     public static int UNEXPORT_LOG_SIZE = Integer.getInteger(ExportTable.class.getName()+".unexportLogSize",1024);
 
+    static boolean EXPORT_TRACES = Boolean.getBoolean(ExportTable.class.getName() + ".exportTraces");
```
daniel-beck, line 494: Should be private?

res0nance: Its used in the test

PR #442, opened 2021-02-23
Various cleanups

jvz:
This cleans up a few areas I found via SonarLint:

* Migration of legacy collection types
* Set interrupted flag on thread after catching interrupt exceptions
* Don't use volatile references for thread safety purposes (would normally require an atomic field reference updater or other complexities)
* Use lambda logging
* Use try-with-resources
* Use `TimeUnit` instead of hard-coded time math

### DIFF
--- a/src/main/java/hudson/remoting/Channel.java
+++ b/src/main/java/hudson/remoting/Channel.java
@@ -53,7 +53,7 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.Date;
-import java.util.Hashtable;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -159,7 +159,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * Requests that are sent to the remote side for execution, yet we are waiting locally until
      * we hear back their responses.
      */
-    /*package*/ final Map<Integer,Request<? extends Serializable,? extends Throwable>> pendingCalls = new Hashtable<>();
+    /*package*/ final Map<Integer,Request<? extends Serializable,? extends Throwable>> pendingCalls =
+            Collections.synchronizedMap(new HashMap<>());
 
     /**
      * Remembers last I/O ID issued from locally to the other side, per thread.
@@ -171,7 +172,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * Records the {@link Request}s being executed on this channel, sent by the remote peer.
      */
     /*package*/ final Map<Integer,Request<?,?>> executingCalls =
-        Collections.synchronizedMap(new Hashtable<>());
+        Collections.synchronizedMap(new HashMap<>());
 
     /**
      * {@link ClassLoader}s that are proxies of the remote classloaders.
@@ -219,7 +220,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
     /**
      * Number of {@link Command} objects sent to the other side.
      */
-    private volatile long commandsSent;
+    private final AtomicLong commandsSent = new AtomicLong();
 
     /**
      * Number of {@link Command} objects received from the other side.
@@ -227,7 +228,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * When a transport is functioning correctly, {@link #commandsSent} of one side
      * and {@link #commandsReceived} of the other side should closely match.
      */
-    private volatile long commandsReceived;
+    private final AtomicLong commandsReceived = new AtomicLong();
 
     /**
      * Timestamp of the last {@link Command} object sent/received, in
@@ -241,7 +242,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
      * without telling us anything, the {@link SocketOutputStream#write(int)} will
      * return right away, and the socket only really times out after 10s of minutes.
      */
-    private volatile long lastCommandSentAt, lastCommandReceivedAt;
+    private final AtomicLong lastCommandSentAt = new AtomicLong();
+    private final AtomicLong lastCommandReceivedAt = new AtomicLong();
 
     /**
      * Timestamp of when this channel was connected/created, in
@@ -595,9 +597,9 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
         transport.setup(this, new CommandReceiver() {
             @Override
             public void handle(Command cmd) {
-                commandsReceived++;
+                commandsReceived.incrementAndGet();
                 long receivedAt = System.currentTimeMillis();
-                lastCommandReceivedAt = receivedAt;
+                lastCommandReceivedAt.set(receivedAt);
                 if (logger.isLoggable(Level.FINE)) {
                     logger.fine("Received " + cmd);
                 } else if (logger.isLoggable(Level.FINER)) {
@@ -764,8 +766,8 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             logger.fine("Send "+cmd);
 
         transport.write(cmd, cmd instanceof CloseCommand);
-        commandsSent++;
-        lastCommandSentAt = System.currentTimeMillis();
+        commandsSent.incrementAndGet();
+        lastCommandSentAt.set(System.currentTimeMillis());
     }
 
     /**
@@ -1001,9 +1003,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             return r.retrieve(this, UserRequest.getClassLoader(callable));
 
         // re-wrap the exception so that we can capture the stack trace of the caller.
-        } catch (ClassNotFoundException e) {
-            throw new IOException("Remote call on " + name + " failed", e);
-        } catch (Error e) {
+        } catch (ClassNotFoundException | Error e) {
             throw new IOException("Remote call on " + name + " failed", e);
         } catch (SecurityException e) {
             throw new IOException("Failed to deserialize response to " + request + ": " + e, e);
@@ -1184,7 +1184,7 @@ public class Channel implements VirtualChannel, IChannel, Closeable {
             // given tickets like JENKINS-20709 that talks about hangs, it seems
             // like a good defensive measure to periodically wake up to make sure
             // that the wait condition is still not met in case we don't call notifyAll correctly
-            wait(30*1000);
+            wait(TimeUnit.SECONDS.toMillis(30));
     }
 
     /**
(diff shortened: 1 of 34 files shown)


src/main/java/hudson/remoting/Channel.java
```diff
@@ -159,7 +159,8 @@
      * Requests that are sent to the remote side for execution, yet we are waiting locally until
      * we hear back their responses.
      */
-    /*package*/ final Map<Integer,Request<? extends Serializable,? extends Throwable>> pendingCalls = new Hashtable<>();
+    /*package*/ final Map<Integer,Request<? extends Serializable,? extends Throwable>> pendingCalls =
+            Collections.synchronizedMap(new HashMap<>());
```
jeffret-b, line 163: Any reason not to use ConcurrentHashMap here?

Actually, if you look at the usages of pendingCalls, I'm pretty certain it's thread safety is broken, before and after. Especially if we accept that the synchronized block near 1083 is important.

src/main/java/hudson/remoting/Channel.java
```diff
@@ -171,7 +172,7 @@
      * Records the {@link Request}s being executed on this channel, sent by the remote peer.
      */
     /*package*/ final Map<Integer,Request<?,?>> executingCalls =
-        Collections.synchronizedMap(new Hashtable<>());
+        Collections.synchronizedMap(new HashMap<>());
```
jeffret-b, line 175: Similar comment here. Why not ConcurrentHashMap?

And the thread safety appears broken both before and after.

(If we're going to change these, we probably ought to make sure the thread safety is solid, otherwise it's probably better not to change.)

src/main/java/hudson/remoting/FastPipedInputStream.java
```diff
@@ -175,6 +175,7 @@ public int read(@Nonnull byte[] b, int off, int len) throws IOException {
                     try {
                         buffer.wait(FastPipedOutputStream.TIMEOUT);
                     } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
```
jeffret-b, line 178: This is certainly considered good practice, but it will require good testing before changing the interrupted behavior in areas like this.

src/main/java/hudson/remoting/ImportedClassLoaderTable.java
```diff
@@ -37,7 +37,7 @@
  */
 final class ImportedClassLoaderTable {
     final Channel channel;
-    final Map<IClassLoader,ClassLoader> classLoaders = new Hashtable<>();
+    final Map<IClassLoader,ClassLoader> classLoaders = new ConcurrentHashMap<>();
```
jeffret-b, line 40: This change to ConcurrentHashMap makes sense.

jeffret-b (commented):
Most of these changes are clearly very nice code improvements.

I'm concerned about the changes to the threading model in `Channel`. Looks like it was already broken, but if we're going to change it, we need to make sure it works correctly.

I'm also concerned about potential impacts from changing how interrupts are handled. (I would say interrupt model, but I'm not sure there's really a model there.)

src/main/java/hudson/remoting/Channel.java
jvz: Code that called this was synchronizing on the collection itself which would require more refactoring to support properly. The synchronized views all use the collection as the lock which works as a drop in replacement for Vector and Hashtable for the most part.

src/main/java/hudson/remoting/Channel.java
jvz: In this particular case, there's no reason to wrap Hashtable with synchronizedMap as that just layers two equivalent locks on the same collection.

src/main/java/hudson/remoting/Channel.java
jeffret-b: Sometimes the calling code is synchronizing on the collection, but not always. Here you changed it from a Hashtable to a synchronizedMap. I'm not sure that helps and it might hurt.

I'm concerned about introducing more synchronization in an area where it isn't clear.

src/main/java/hudson/remoting/Channel.java
jeffret-b: I'm not understanding what you mean? If it doesn't need to be wrapped with synchronizedMap, why does your change propose to still do that?

jvz: Because you should only use one or the other.

src/main/java/hudson/remoting/Channel.java
jvz: To be more clear: you only need to synchronize on the collection to do atomic operations. You don't need to synchronize on it for get and put for example, but if you wanted to iterate over the entries, you should synchronize then to avoid concurrent modifications.

jeffret-b:
@jvz , I updated the thread-safety for Channel as I had proposed. Have you tested this proposed change? Especially related to the changes in interrupt handling.

jvz:
No I didn't do any manual testing here; I made changes based on my understanding of the Java memory model and how concurrency works in the JVM in general along with examining the call sites of the code I changed.

jvz:
Changes look good. We could try running this in one of our Jenkins setups for some time to see if it's ok or not.

jeffret-b:
I've got some test configurations I typically run through with a Remoting release or significant changes. They're pretty limited, but they verify some basic situations. They take a little bit of time to configure and run so I won't get around to doing it for a bit.

jeffret-b (approved):
This looks like it should be fine.

I'm still not entirely certain about the changes to interrupts but they make sense.

I've run it through some basic testing and it works fine. I've only got limited testing scenarios available so it may still have problems in some less resilient production environments. We'll probably have to proceed to a release to get any information about other scenarios.

src/main/java/hudson/remoting/ChunkedOutputStream.java
```diff
@@ -24,7 +24,9 @@
     private final OutputStream base;
 
     public ChunkedOutputStream(int frameSize, OutputStream base) {
-        assert 0<frameSize && frameSize<=Short.MAX_VALUE;
+        if (frameSize < 0 || frameSize > Short.MAX_VALUE) {
+            throw new IllegalArgumentException("Illegal frame size");
```
oleg-nenashev, line 28:
```suggestion
            throw new IllegalArgumentException("Illegal frame size: " + frameSize);
```

src/main/java/hudson/remoting/ClassFilter.java
```diff
@@ -228,14 +226,6 @@ public static void appendDefaultFilter(Pattern filter) throws ClassFilterExcepti
             return patterns;
         } catch (IOException ex) {
             throw new Error("Could not load user provided overrides for ClassFiltering from as "+prop+" does not exist or is not readable.",ex);
-        } finally {
-            if (br != null) {
-                try {
-                    br.close();
-                } catch (IOException ioEx) {
-                    LOGGER.log(Level.WARNING, "Failed to cleanly close input stream", ioEx);
```
oleg-nenashev, line 236: It changes the code behavior. Instead of a log warning, close operation failure will result in an `Error` being thrown

src/main/java/hudson/remoting/FastPipedOutputStream.java
```diff
@@ -166,6 +166,7 @@ public void write(@Nonnull byte[] b, int off, int len) throws IOException {
                     try {
                         buf.wait(TIMEOUT);
                     } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
```
oleg-nenashev, line 169: hm, why?

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -349,29 +350,15 @@ public JnlpAgentEndpoint resolve() throws IOException {
...
             s.setReuseAddress(true);
             SocketAddress sa = new InetSocketAddress(hostname, port);
             s.connect(sa, 5000);
+            return s.isConnected();
         } catch (IOException e) {
             LOGGER.warning(e.getMessage());
-        } finally {
-            if (s != null) {
-                if (s.isConnected()) {
-                    exitStatus = true;
```
oleg-nenashev, line 365: Change of the behavior. when IOException happens. New code will always return false in this case

oleg-nenashev (changes requested):
There are behavior changes in try-with-resources. Also, I am not sure `Thread.currentThread().interrupt()` is really needed in all cases

basil:
I concur with the existing feedback that has been left regarding the interrupt handling changes carrying a high risk of regression. Without the interrupt handling changes, this PR would get a straightforward approval from me; with them, I think this turns into a tricky problem of figuring out how to test the PR appropriately.

jeffret-b:
@jvz do you plan on getting back to this and resolving the concerns Oleg raised?

jvz:
Yeah, I'll clean this up sometime soon.

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -102,6 +102,7 @@ public void run() {
         } catch (InterruptedException e) {
             // use interruption as a way to terminate the ping thread.
             LOGGER.fine(getName()+" is interrupted. Terminating");
+            Thread.currentThread().interrupt();
```
jvz, line 105: Redundant since this is the last bit of code this thread will run.

src/main/java/hudson/remoting/JarCacheSupport.java
```diff
@@ -121,6 +121,7 @@ public void run() {
                 bailout(e);
 
                 LOGGER.log(Level.WARNING, String.format("Interrupted while resolving a jar %016x%016x", sum1, sum2), e);
+                Thread.currentThread().interrupt();
```
jvz, line 124: Safe as this is run by an `Executor`.

jeffret-b (approved):
Looks good. The latest changes reverts the risky, though probably correct, changes.

jeffret-b:
@oleg-nenashev , do you want to update your review on this? Otherwise I'll proceed with it in a day or two.

src/main/java/org/jenkinsci/remoting/protocol/cert/DelegatingX509ExtendedTrustManager.java
```diff
@@ -29,14 +29,12 @@
...
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.X509ExtendedTrustManager;
-import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 
 /**
  * An {@link X509ExtendedTrustManager} that delegates to a runtime mutable delegate {@link X509ExtendedTrustManager}.
  *
  * @since 3.0
  */
-@IgnoreJRERequirement // TODO We override some methods in Java 7, so remove this ignore when baseline is Java 7
```
jglick, line 39: And the rest in #479.

PR #443, opened 2021-03-22
JENKINS-65046:Dynamic Class Loading Fails In Reflection

saptarshide007:
After updating from java 8 to java 11 getting the below error:-
Exception in thread "main" java.lang.IllegalArgumentException: object is not an instance of declaring class
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:566)
at hudson.remoting.Launcher.addClasspath(Launcher.java:129)

On analysing the remoting source code found that the hudson.remoting.Launcher.addClasspath invokes URLClassLoader and passes the instance using ClassLoader.getSystemClassLoader().
However in java 11, this seems like an illegal operation as ClassLoader.getSystemClassLoader() does not get the instance of  URLClassLoader due to which we are getting the error "object is not an instance of declaring class".

To fix this issue what updated the addClasspath method to invoke the URLClassLoader using a new instance of the class.

Reference:-
GitIssues:-
https://github.com/redhat-developer/rsp-server/issues/274
https://github.com/redhat-developer/rsp-server/issues/275

Jira:-
https://issues.jenkins.io/browse/JENKINS-65046

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -150,7 +150,10 @@ public class Launcher {
         $addURL.setAccessible(true);
 
         for(String token : pathList.split(File.pathSeparator))
-            $addURL.invoke(ClassLoader.getSystemClassLoader(),new File(token).toURI().toURL());
+        { 
+        	URL[] urls= {new File(token).toURI().toURL()};
+        	$addURL.invoke(URLClassLoader.newInstance(urls),urls[0]);
+        }
 
         // fix up the system.class.path to pretend that those jar files
         // are given through CLASSPATH or something.


jeffret-b (commented):
This makes sense given the reports in Jenkins and the changes reported in this area as part of Java 11.

Have you tested this? In both Java 8 and 11 environments?

saptarshide007:
> This makes sense given the reports in Jenkins and the changes reported in this area as part of Java 11.
> 
> Have you tested this? In both Java 8 and 11 environments?

saptarshide007:
> This makes sense given the reports in Jenkins and the changes reported in this area as part of Java 11.
> 
> Have you tested this? In both Java 8 and 11 environments?

saptarshide007:
Yes I did, works in both 8 and 11.

jeffret-b:
Great! Thanks for that. I don't have use cases handy for that.

jeffret-b (approved):

oleg-nenashev (changes requested):
The previous behavior is definitely unreliable, there is no guarantee that the System Classloader will be the URL classloader for any JDK. That said, after the switch to a new classloader instance there is no need in reflection at all. `URL`s  can be passed directly to the URL Classloader constructor.

I am also not sure what is the expected behavior. A new classloader instance is created for each loop iteration, and then URLs are added to that. Then this new classloader instance is just ignored and later garbage collected. I have no idea how this code would take any effect after the change.

@saptarshide007 could you please confirm that you have actually tested the code with the classpath parameter which really injects a class to be consumed?

saptarshide007:
> The previous behavior is definitely unreliable, there is no guarantee that the System Classloader will be the URL classloader for any JDK. That said, after the switch to a new classloader instance there is no need in reflection at all. `URL`s can be passed directly to the URL Classloader constructor.
> 
> I am also not sure what is the expected behavior. A new classloader instance is created for each loop iteration, and then URLs are added to that. Then this new classloader instance is just ignored and later garbage collected. I have no idea how this code would take any effect after the change.
> 
> @saptarshide007 could you please confirm that you have actually tested the code with the classpath parameter which really injects a class to be consumed?

@oleg-nenashev, thank you for the analysis. Till java 8, AppClassLoader/SystemCLassLoader(sun.misc.Launcher.AppClassLoader) was child of URLclassloader. The addUrl method of the URLclassLoader is protected, and thus reflection was used to access this method. This code made sense till then.
From java 9 the AppClassLoader(jdk.internal.loader.ClassLoaders.AppClassLoader) no longer is a child of URLclassLoader. That is the reason why we are not able to invoke the addUrl method of URLClassLoader by AppClassLoader.
Yes I agree that the, solution that is shared will not load the dependencies, as you said  the new classloader instance is just ignored and later garbage collected.
It did work in my case, but I probably need to check why it worked. I some how missed to dig deep into it. However, this will definitely not work.
We need to somehow add the URL dependencies to the class loader. As per my understanding we can achieve this in 2 way, either we have write our own classloader or we can use agents(java.lang.instrument,https://cgjennings.ca/articles/java-9-dynamic-jar-loading/).
Kindly hold on to this PR, will try to fix this and update the same here. Also kindly share any suggestions/feedback on the same.

jglick:
See #468. AFAICT this PR just makes `-cp` not work at all, so it should be closed.

jeffret-b:
I'm going to go ahead and close this one. If you want you can reopen it and fix things.

PR #445, opened 2021-04-09
[JENKINS-63520] Sequence the startup and receiving to avoid a deadlock.

jeffret-b:
See [JENKINS-63520](https://issues.jenkins.io/browse/JENKINS-63520).

Some users have reported a rare thread deadlock when starting an inbound TCP agent. There are no known reports of this occurring at other times. The reporter included a stack track showing information about the deadlock. It occurs when the agent ProtocolStack is still initializing and a message is received on the networking layer.

Personally, I've never observed the problem and have no way of testing it. I've tested normal behavior and this fix has worked fine.

This fix delays the processing of any received messages until after the ProtocolStack has fully initialized.

- [n/a] Link to relevant pull requests, esp. upstream and downstream changes
- [n/p] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
@@ -37,6 +37,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.jenkinsci.remoting.protocol.IOHub;
 import org.jenkinsci.remoting.protocol.NetworkLayer;
+import org.jenkinsci.remoting.protocol.ProtocolStack;
 import org.jenkinsci.remoting.util.ByteBufferQueue;
 import org.jenkinsci.remoting.util.IOUtils;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/remoting/protocol/impl/BIONetworkLayer.java
```diff
@@ -230,6 +231,15 @@ public synchronized void run() {
          */
         @Override
         public void run() {
+            while (!ProtocolStack.isStarted().get()) {
```
jvz, line 234: You could always use a CountDownLatch that counts down when transitioning to started. Then this method can call await rather than busy-waiting with sleep.

jeffret-bg: Good idea. That's an improved implementation.

jeffret-b:
Looks like we lost the agent mid-build. Let's try again.

PR #391, opened 2020-07-23
Add protection against missing trailing slashes when using websocket.

nzdjb:
Currently, if -webSocket is set and -url does not have a trailing slash, connection will fail with an UnresolvedAddressException. This can prove confusing to diagnose and resolve.

This PR adds a check for a trailing slash and adds it if missing.

PR #454, opened 2021-05-18
Define `hudsonUrl` in WebSocket mode to fix agent installers

jglick:
I noticed in CloudBees CI trying to launch a shared agent in WebSocket mode from the Java Web Start button an error indicating that the equivalent of https://github.com/jenkinsci/slave-installer-module/blob/8ab32a0616d147adef0d27fb51a359796e069eeb/src/main/java/org/jenkinsci/modules/slave_installer/impl/InstallerGui.java#L78 was throwing an NPE, despite the Jenkins root URL being configured. I could not reproduce a similar stack trace using an OSS agent connection, though the installer menu was not created either, so maybe the exception was just swallowed somewhere, or printed to the agent log which I forgot to check. Anyway, it seems clear from code that this `getHudsonUrl` method would have worked in TCP but not WebSocket mode.

Also cleaning up a trailing slash check introduced in #391.

Tested in context by building Jenkins core with this, then building CBCI operations center & connected controller with that, running both, hooking them up, and going through two scenarios:

* static agent on the CC
* shared agent on the OC

In both cases, used the JWS launch button and got a **File** from the agent installer (systemd in my case) as expected.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -30,6 +30,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.Socket;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


jeffret-b (approved):

jglick:
Note that the agent installer system only appears in GUI mode, which is only enabled when launched via JWS, which is not something we generally encourage. [Discussion in JEP-230](https://github.com/jenkinsci/jep/blob/master/jep/230/README.adoc#marking-slave-installer-as-detached)

oleg-nenashev (approved):
LGTM, thanks @jglick !

PR #455, opened 2021-05-18
Custom connection headers

agentgonzo:
Add the -connectionHeaders flag to add additional custom headers to the connection request.

These additional headers are then used when connecting to jenkins either via JNLP or websockets. 

Use-case: You have a reverse proxy in front of jenkins that requires custom headers to authenticate/route correctly.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -158,6 +158,7 @@ public class Engine extends Thread {
     private final String secretKey;
     private final String agentName;
     private boolean webSocket;
+    private Map<String, String> connectionHeaders;
     private String credentials;
     private String protocolName;
     private String proxyCredentials = System.getProperty("proxyCredentials");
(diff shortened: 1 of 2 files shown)


agentgonzo:
@jglick @oleg-nenashev @abayer - I could do with some additional guidance of how to test this, as I couldn't find any existing tests. 

our primary use-case is to support headers in websockets, but it made sense to add them to the JNLP connection as well.

oleg-nenashev (approved):

jglick:
> it made sense to add them to the JNLP connection as well

Does it? The only thing the HTTP connection for TCP inbound agents (please do not say _JNLP_ unless you really mean it) does is to get a host & port, which is public information (an unauthenticated endpoint) and generally static. The actual TCP connection uses a custom protocol, so you could not interpret any custom headers in a reverse proxy—if you could even use a reverse proxy with a TCP agent at all.

I would limit the addition to WebSocket agents unless and until there is a use case for doing something with TCP agents, which might look very different.

jeffret-b (changes requested):
As Jesse mentions, let's not add the headers to the TCP protocol, as they don't make any sense there. Unless there is an identified use case for that.

agentgonzo:
Thanks @jglick @jeffret-b . Can you take a look at #458 please?

PR #457, opened 2021-05-18
Reference new tcp-agent-listener endpoint

daniel-beck:
Downstream of https://github.com/jenkinsci/jenkins/pull/5495

It is unclear to me what the compatibility requirements of remoting are. This is a very simple patch that assumes remoting can be older (~minimum version defined in core), but never newer.

Also noting two TODOs where further cleanup can be done once https://github.com/jenkinsci/jenkins/pull/5494 is released or developer documentation migrated away from the wiki.

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
@@ -377,8 +377,8 @@ public class JnlpAgentEndpointResolver extends JnlpEndpointResolver {
     @Nonnull
     private URL toAgentListenerURL(@Nonnull String jenkinsUrl) throws MalformedURLException {
         return jenkinsUrl.endsWith("/")
-                ? new URL(jenkinsUrl + "tcpSlaveAgentListener/")
-                : new URL(jenkinsUrl + "/tcpSlaveAgentListener/");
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -377,8 +377,8 @@ private boolean isPortVisible(String hostname, int port) {
     @Nonnull
     private URL toAgentListenerURL(@Nonnull String jenkinsUrl) throws MalformedURLException {
         return jenkinsUrl.endsWith("/")
-                ? new URL(jenkinsUrl + "tcpSlaveAgentListener/")
-                : new URL(jenkinsUrl + "/tcpSlaveAgentListener/");
+                ? new URL(jenkinsUrl + "tcp-agent-listener/")
+                : new URL(jenkinsUrl + "/tcp-agent-listener/");
```
jglick, line 381: Merge conflict with #454.

jglick (changes requested):
So this will fail when run against older cores. Generally Remoting is intended to be bidirectionally compatible, so you would need to try the new endpoint but fall back to the older one.

jglick:
https://github.com/jenkinsci/jenkins/pull/5495 was closed so I think this is obsolete.

PR #458, opened 2021-05-19
Allow setting of custom websocket headers

agentgonzo:
Add the -websocketHeader flag to add additional custom headers to the connection request.

These additional headers are then used when connecting to jenkins via websockets.

Use-case: You have a reverse proxy in front of jenkins that requires custom headers to authenticate/route correctly.

Supercedes #455 

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -351,6 +351,11 @@ public class Engine extends Thread {
         this.webSocket = webSocket;
     }
 
+    /**
+     * Sets map of custom websocket headers. These headers will be applied to the websocket connection to Jenkins.
+     *
+     * @param webSocketHeaders a map of the headers to apply to the websocket connection
+     */
     public void setWebSocketHeaders(@Nonnull Map<String, String> webSocketHeaders) {
         this.webSocketHeaders = webSocketHeaders;
     }
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -158,6 +158,7 @@ public class Engine extends Thread {
     private final String secretKey;
     private final String agentName;
     private boolean webSocket;
+    private Map<String, String> webSocketHeaders;
     private String credentials;
     private String protocolName;
     private String proxyCredentials = System.getProperty("proxyCredentials");
@@ -350,6 +351,10 @@ public class Engine extends Thread {
         this.webSocket = webSocket;
     }
 
+    public void setWebSocketHeaders(@Nonnull Map<String, String> webSocketHeaders) {
+        this.webSocketHeaders = webSocketHeaders;
+    }
+
     /**
      * If set, connect to the specified host and port instead of connecting directly to Jenkins.
      * @param tunnel Value. {@code null} to disable tunneling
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -300,6 +307,8 @@ public Engine createEngine() {
                 headlessMode ? new CuiListener() : new GuiListener(),
                 urls, args.get(0), agentName, directConnection, instanceIdentity, new HashSet<>(protocols));
         engine.setWebSocket(webSocket);
+        if(webSocketHeaders!=null)
+            engine.setWebSocketHeaders(webSocketHeaders);
```
jglick, line 311: I do not suppose we need to imitate KK’s unusual formatting in new code, but does not matter!

jglick (approved):
Can I assume this has been tested in context somehow?

src/main/java/hudson/remoting/jnlp/Main.java
agentgonzo: I tend to favour consistency over correctness. Given there are a bunch of similar lines above/below it, I'd preder to keep that block consistent (either mimicking KK's style or replacing it all)

src/main/java/hudson/remoting/Engine.java
```diff
@@ -350,6 +351,10 @@ public void setWebSocket(boolean webSocket) {
         this.webSocket = webSocket;
     }
 
+    public void setWebSocketHeaders(@Nonnull Map<String, String> webSocketHeaders) {
```
oleg-nenashev, line 354: Javadoc would be nice, because this endpoint might be used on both server and client sides by API users

src/main/java/hudson/remoting/jnlp/Main.java
oleg-nenashev: We can refactor it in a follow-up if anyone has time

oleg-nenashev (approved):
Thanks @agentgonzo ! Looks good. 


src/main/java/hudson/remoting/Engine.java
agentgonzo: @oleg-nenashev added.

jeffret-b:
Try another build.

src/main/java/hudson/remoting/jnlp/Main.java
jeffret-b: We've been trying to move Remoting code to follow a more standard formatting style. But, this particular area hasn't been updated yet, so I suppose it's okay to follow the local formatting.

jeffret-b:
We need a successful build, or a viable explanation for why we can ignore the failures.

jvz:
That looks like an infra issue of some sort. It's a maven error.

jeffret-b:
Probably. I'm still not comfortable with merging if we don't have a
successful build.

jvz:
Right. Let's wait for the infra to repair itself first.

abayer:
I've tested this manually and it works - the headers get added to the websocket connection properly.

abayer:
(I also reran the build and that worked, so woo)

agentgonzo:
Great, thanks @abayer. Definitely worth getting a green build before merging.

agentgonzo:
I can also confirm from manual testing that the headers get added to the websocket connection and it works as expected.

jeffret-bg (approved):

PR #468, opened 2021-07-06
[JENKINS-64831] Deprecate the `-cp` option to `Launcher`

jglick:

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -144,9 +144,10 @@ public class Launcher {
     public String proxyCredentials = null;
 
     @Option(name="-cp",aliases="-classpath",metaVar="PATH",
-            usage="add the given classpath elements to the system classloader.")
+            usage="add the given classpath elements to the system classloader. (DEPRECATED)")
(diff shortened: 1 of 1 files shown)


basil (approved):

jeffret-b (approved):

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -144,9 +144,10 @@ public void setTextMode(boolean b) {
     public String proxyCredentials = null;
 
     @Option(name="-cp",aliases="-classpath",metaVar="PATH",
-            usage="add the given classpath elements to the system classloader.")
+            usage="add the given classpath elements to the system classloader. (DEPRECATED)")
```
ryancurrah, line 147: Should this `docker-inbound-agent` launch script be updated to not use `-cp`? https://github.com/jenkinsci/docker-inbound-agent/blob/master/jenkins-agent#L121

jglick: No, that is the `--classpath` / `-cp` option to the `java` launcher. This PR is about an application option, like

```bash
java -jar agent.jar -cp …
```

PR #471, opened 2021-07-14
Instrumentation feature

Aki-7:
## Proposal to introduce remoting instrumentation feature.

Related to GSoC '21 [Remoting Monitoring project](https://www.jenkins.io/projects/gsoc/2021/projects/remoting-monitoring/).

In the Remoting Monitoring project, we are trying to observe the behavior of remoting modules using OpenTelemetry. And users will be able to visualize the remoting behavior like this.

image

To collect more accurate and valuable data, we want to introduce a new feature to observe the remoting module.

### What I did

1. Adjust Launcher's `-cp` flag for Java9+
  - It doesn't behave as Java's `-cp` flag in Java9+, but somehow valuable.
  - In Java9+, we use a new URLClassLoader to load and set it to context class loader.
2. Introduce `LauncherInstrumentationListener` and `EngineInstrumentationListener`
  - They issue an event. e.g. `onRunWithStdinStdout`
3. Enable to load user-defined `LauncherInstrumentationListener` and `EngineInstrumentationListener`
  - Users have to pass the JAR file using the `-cp` flag
    - the JAR file will include user-defined listener classes, which inherit `LauncherInstrumentationListenerAdapter` or `EngineInstrumentationListenerAdapter`
  - Users have to set System properties.
    - value is the comma-separated canonical names of the listener classes users created. 

Command example
```
java \
-Dhudson.remoting.Engine.engineInstrumentationListenerCanonicalNames=user.defined.engine.instrumentation.listener \
-Dhudson.remoting.Launcher.launcherInstrumentationListenerCanonicalNames=user.defined.launcher.instrumentation.listener \
-jar agent.jar -jnlpUrl $JENKINS_ROOT/computer/$NODE_NAME/jenkins-agent.jnlp \
-workDir $WORK_DIR -cp user-defined-instrumentations.jar
``` 

Some references regarding the `-cp` flag.
https://github.com/jenkinsci/remoting/pull/468
https://github.com/jenkinsci/remoting/pull/443

### I would appreciate any negative or positive feedback!

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.Socket;
@@ -135,6 +136,7 @@ public class Engine extends Thread {
     public final EngineListener listener;
 
     private final EngineListenerSplitter events = new EngineListenerSplitter();
+    private final EngineInstrumentationListenerSplitter instrumentation = new EngineInstrumentationListenerSplitter();
     /**
      * To make Jenkins more graceful against user error,
      * JNLP agent can try to connect to multiple possible Jenkins URLs.
@@ -288,6 +290,7 @@ public class Engine extends Thread {
      *               This method can be used for testing startup logic.
      */
     /*package*/ void startEngine(boolean dryRun) throws IOException {
+        loadEngineInstrumentationListeners();
         LOGGER.log(Level.INFO, "Using Remoting version: {0}", Launcher.VERSION);
         @CheckForNull File jarCacheDirectory = null;
         
@@ -330,6 +333,37 @@ public class Engine extends Thread {
         if (!dryRun) {
             this.start();
         }
+    }
+
+    /*package*/boolean loadEngineInstrumentationListeners() {
+        boolean success = true;
+        String listenerCanonicalNamesProp = System.getProperty("hudson.remoting.Engine.engineInstrumentationListenerCanonicalNames", "");
+        if (listenerCanonicalNamesProp.trim().equals("")) return success;
+
+        String[] listenerCanonicalNames = listenerCanonicalNamesProp.split(":");
+        for (String listenerName : listenerCanonicalNames) {
+            try {
+                Class<?> clazz = Thread.currentThread().getContextClassLoader().loadClass(listenerName);
+                Object listener = clazz.getDeclaredConstructor(null).newInstance();
+                if (listener instanceof EngineInstrumentationListener) {
+                    instrumentation.add((EngineInstrumentationListener) listener);
+                    LOGGER.log(Level.INFO, String.format("Loaded an external engine instrumentation listener (%s)", listenerName));
+                } else {
+                    LOGGER.log(Level.WARNING, String.format("Failed to use %s. Not an instance of %s", listenerName, EngineInstrumentationListener.class.getCanonicalName()));
+                    success = false;
+                }
+            } catch (ClassNotFoundException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to load %s", listenerName), e);
+                success = false;
+            } catch (NoSuchMethodException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to get constructor of %s", listenerName), e);
+                success = false;
+            } catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to instantiate %s", listenerName), e);
+                success = false;
+            }
+        }
+        return success;
     }
 
     /**
@@ -492,6 +526,7 @@ public class Engine extends Thread {
     @Override
     @SuppressFBWarnings(value = "HARD_CODE_PASSWORD", justification = "Password doesn't need to be protected.")
     public void run() {
+        instrumentation.onStart(this, webSocket);
         if (webSocket) {
             runWebSocket();
             return;
@@ -661,6 +696,7 @@ public class Engine extends Thread {
                 }
                 this.protocolName = "WebSocket";
                 events.status("Connected");
+                instrumentation.onConnected();
                 ch.get().join();
                 events.status("Terminated");
                 if (noReconnect) {
@@ -810,6 +846,7 @@ public class Engine extends Thread {
                     }
 
                     events.status("Connected");
+                    instrumentation.onConnected();
                     channel.join();
                     events.status("Terminated");
                 } finally {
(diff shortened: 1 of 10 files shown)


src/main/java/hudson/remoting/Launcher.java
```diff
@@ -147,16 +148,28 @@ public void setTextMode(boolean b) {
...
+            System.setProperty("java.class.path",System.getProperty("java.class.path")+File.pathSeparatorChar+pathList);
+        } else {
+            // Java 9 or later
+            List<URL> urls = new ArrayList<>();
+            for (String token : pathList.split(File.pathSeparator))
+                urls.add(new File(token).toURI().toURL());
+
+            ClassLoader currentThreadClassLoader = Thread.currentThread().getContextClassLoader();
+            URLClassLoader urlClassLoader = new URLClassLoader(urls.toArray(new URL[]{}), currentThreadClassLoader);
+            Thread.currentThread().setContextClassLoader(urlClassLoader);
```
jglick, line 171: Do not rely on `Thread.contextClassLoader` please.

If you want to offer a flag allowing custom listener code to be loaded, do not do it this way. First of all, stop using system properties; introduce a real option with documentation. Then have it somehow specify the location of a JAR and/or class name(s) to load from it in a fresh class loader.

(Nicest would be to use SezPoz but adding a new library dep for this purpose is likely overkill, and anyway it would potentially clash with the copy loaded from `jenkins-core.jar`.)

src/main/java/hudson/remoting/Engine.java
```diff
@@ -332,6 +335,37 @@ public synchronized void startEngine() throws IOException {
...
+                    success = false;
+                }
+            } catch (ClassNotFoundException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to load %s", listenerName), e);
+                success = false;
+            } catch (NoSuchMethodException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to get constructor of %s", listenerName), e);
+                success = false;
+            } catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
+                LOGGER.log(Level.WARNING, String.format("Failed to instantiate %s", listenerName), e);
```
jglick, line 362: KISS

```suggestion
            } catch (Exception e) {
                LOGGER.log(Level.WARNING, "Failed to load " + listenerName, e);
                success = false;
```

src/main/java/hudson/remoting/Engine.java
```diff
@@ -332,6 +335,37 @@ public synchronized void startEngine() throws IOException {
...
+            try {
+                Class<?> clazz = Thread.currentThread().getContextClassLoader().loadClass(listenerName);
+                Object listener = clazz.getDeclaredConstructor(null).newInstance();
+                if (listener instanceof EngineInstrumentationListener) {
+                    instrumentation.add((EngineInstrumentationListener) listener);
+                    LOGGER.log(Level.INFO, String.format("Loaded an external engine instrumentation listener (%s)", listenerName));
+                } else {
+                    LOGGER.log(Level.WARNING, String.format("Failed to use %s. Not an instance of %s", listenerName, EngineInstrumentationListener.class.getCanonicalName()));
+                    success = false;
+                }
```
jglick, line 354: Java’s built-in CCEs in newer versions are better than any message you come up with.

```suggestion
                instrumentation.add((EngineInstrumentationListener) listener);
                LOGGER.log(Level.INFO, "Loaded an external engine instrumentation listener ({0})", listenerName);
```

jglick (changes requested):
Do not rely on `-cp`.

jglick:
I would actually suggest considering a different design altogether. Use the existing https://github.com/jenkinsci/remoting/blob/dd0a1a6a283469bb668c24dffc132c61ec148fb3/src/main/java/hudson/remoting/EngineListener.java#L41-L67 and/or https://github.com/jenkinsci/remoting/blob/dd0a1a6a283469bb668c24dffc132c61ec148fb3/src/main/java/hudson/remoting/Channel.java#L643-L692 (adding some `default` methods as needed for finer-grained or more informative events) and otherwise do nothing in `remoting`. Rather, make a Jenkins plugin which injects listener code into the agent when it connects. There are various examples of this already. This would make it much easier to distribute various listener implementations (just use the update center); there is no need to adjust the startup arguments for `agent.jar`; the Remoting code stays cleaner. The only downside is that you would not get access to details of what is going on before the connection is established.

src/main/java/hudson/remoting/Launcher.java
jglick: For example, define a hook as a JAR whose manifest has a `Main-Class` entry giving the FQN of a class implementing whichever listener interface makes sense, with a `public` no-arg constructor. Then define a CLI arg

```bash
-listener /path/to/impl.jar
```

When specified, the hook would be loaded in its own private `URLClassLoader` parented to that of `agent.jar`.

jglick:
> you would not get access to details of what is going on before the connection is established

I see [this survey](https://docs.google.com/presentation/d/1QKu3JI5eeE5VxtaUlsnw9kmTZLYvxYhBJrIzvYXSB4U/edit#slide=id.ge5bb69c10f_1_200) but it is not obvious to me that the extra bit of diagnostics possible with the current approach is worth making the distribution and configuration of the listener much more awkward. Typically the problem is sporadic disconnection; if you cannot connect at all, there are better ways to diagnose that directly, not using general telemetry.

Not sure what “robust against agent restart” is about. With the plugin approach, if the agent restarts, Jenkins will see a new `Channel`, and send the listener to it again. So that should not be an issue.

Aki-7:
Thank you so much for your review and advice!
I'm very grateful for your positive suggestions!

I'll provide a short answer for
> Not sure what “robust against agent restart” is about.

before going to bed.

> Typically the problem is sporadic disconnection

I agree with this and I thought the plugin approach would work well but I found that, in the UNIX agent, UnixSlaveRestarter restarts the agent with `execv` when disconnection happens.

code: https://github.com/jenkinsci/jenkins/blob/95e85495ca6b26ba61ff4a1bf18d181d5e218bc9/core/src/main/java/jenkins/slaves/restarter/UnixSlaveRestarter.java#L50-L63 
issue in my repo: https://github.com/jenkinsci/remoting-opentelemetry-plugin/issues/65

So if we take the plugin approach, the monitoring program will be killed every time sporadic disconnection happens. And we cannot collect the trace of the next connection and metrics after the sporadic disconnection. I think this information can be useful for troubleshooting and that's what I mean by "not robust against agent restart".

jeffret-b:
Generally, I agree with Jesse's comments and recommendations, particularly about keeping it simple. Launching the agent is already complicated enough -- it's important not to further complicate that. If the feature is valuable we need to implement reasonable support for it.

While I understand users would like to have diagnostics about what happens before a connection is completed, I doubt they would actually find anything useful there. I wouldn't prioritize that as a high-value feature or sacrifice other improvements for that.

Aki-7:
Thank you so much for a lot of feedback!
I'll try to things more simple!

Aki-7:
I'll close this PR, thank you so much for your feedback!

context: https://github.com/jenkinsci/remoting-opentelemetry-plugin/issues/65#issuecomment-886361757

oleg-nenashev:
Sorry for not following up on the discussion

PR #476, opened 2021-08-25
Add a warning about re-using agent names

0xDAFE:
This PR updates the documentation regarding the inbound agent secret. The way it is currently implemented is a potential pitfall for Incident Response teams and not documented anywhere, as discussed in https://issues.jenkins.io/browse/SECURITY-2425.

### DIFF
--- a/docs/inbound-agent.md
+++ b/docs/inbound-agent.md
@@ -20,6 +20,8 @@ If you use a different mechanism than clicking the "Launch" button, this page co
 The long string of hex digits is a secret key that the client needs to establish the connection. 
 Depending on how you connect you may also need the agent name, "AgentName" in this example, or the (misnamed) jnlpUrl, "https://myjenkins.example.com/computer/AgentName/jenkins-agent.jnlp".
 
+**Security Warning**: The secret key is actually not created randomly but implemented as a MAC of the agent name. This means that even when deleting and creating a new agent, the secret will not change if the agent name stays the same. If you assume a secret is compromised, it is therefore crucial to avoid re-using the compromised agent's name.
+
 ### "Launch" button
 This is the only launch mechanism that actually uses JNLP.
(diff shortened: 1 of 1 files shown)


docs/inbound-agent.md
```diff
@@ -20,6 +20,8 @@ If you use a different mechanism than clicking the "Launch" button, this page co
 The long string of hex digits is a secret key that the client needs to establish the connection. 
 Depending on how you connect you may also need the agent name, "AgentName" in this example, or the (misnamed) jnlpUrl, "https://myjenkins.example.com/computer/AgentName/jenkins-agent.jnlp".
 
+**Security Warning**: The secret key is actually not created randomly but implemented as a MAC of the agent name. This means that even when deleting and creating a new agent, the secret will not change if the agent name stays the same. If you assume a secret is compromised, it is therefore crucial to avoid re-using the compromised agent's name.
+
```
daniel-beck, line 24: This reads like an obvious "patch" to the existing documentation (admonition format, self-contained paragraph). This would be better off integrated with existing docs. How about this?

```suggestion
The long string of hex digits is a secret key that the client needs to establish the connection. 
This key is based on the agent name and a per-instance secret.
Depending on how you connect you may also need the agent name, "AgentName" in this example, or the (misnamed) jnlpUrl, "https://myjenkins.example.com/computer/AgentName/jenkins-agent.jnlp".

Note that the secret key will always be the same for a given agent name on the same Jenkins instance.
If the secret key of an agent is compromised, do not reuse the agent name on the affected Jenkins instance.

```

Given this warns about something that wasn't claimed anywhere, it also seems a bit much to emphasize it so strongly.

jglick: I would say _per-controller_; not sure if readers will understand what an _instance_ is.

jglick (approved):

daniel-beck (approved):

jglick:
https://issues.jenkins.io/browse/JENKINS-70168 FTR

PR #481, opened 2021-09-16
Remove setting of null security manager

timja:
Running an agent on Java 17:

```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by hudson.remoting.jnlp.Main (file:/Users/timja/Downloads/agent.jar)
WARNING: Please consider reporting this to the maintainers of hudson.remoting.jnlp.Main
WARNING: System::setSecurityManager will be removed in a future release
```

### DIFF
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -246,17 +246,6 @@ public class Main {
      * Main without the argument handling.
      */
     public static void _main(String[] args) throws IOException, InterruptedException, CmdLineException {
-        // see http://forum.java.sun.com/thread.jspa?threadID=706976&tstart=0
-        // not sure if this is the cause, but attempting to fix
-        // https://hudson.dev.java.net/issues/show_bug.cgi?id=310
-        // by overwriting the security manager.
-        try {
-            System.setSecurityManager(null);
-        } catch (SecurityException e) {
-            // ignore and move on.
-            // some user reported that this happens on their JVM: http://d.hatena.ne.jp/tueda_wolf/20080723
-        }
-
         // if we run in Mac, put the menu bar where the user expects it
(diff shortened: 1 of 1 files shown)


basil (approved):

jglick (approved):
All of these links are dead.

jeffret-b (approved):

oleg-nenashev (approved):
Maybe not a chore, but whatever. Thanks for the patch!

oleg-nenashev:

timja:
@jeffret-b can we ship this?

jeffret-b:
@timja , sure, that's fine.

AckermannC:
I think this change is causing an issue when trying to install the jenkins agent using java webstart.
See this issue: https://issues.jenkins.io/browse/JENKINS-67000

MarkEWaite:
> I think this change is causing an issue when trying to install the jenkins agent using java webstart. See this issue: https://issues.jenkins.io/browse/JENKINS-67000

Based on https://stackoverflow.com/questions/19487407/impact-of-system-setsecuritymanagernull, I think that this was the right change to make.  Disabling applet security checks seems like a very risky choice.

AckermannC:
> > I think this change is causing an issue when trying to install the jenkins agent using java webstart. See this issue: https://issues.jenkins.io/browse/JENKINS-67000
> 
> Based on https://stackoverflow.com/questions/19487407/impact-of-system-setsecuritymanagernull, I think that this was the right change to make. Disabling applet security checks seems like a very risky choice.

I get it, security is important to you and you don't think time should be wasted on fixing old technology... You are very outspoken on this in case 67000.
But there is still the launch-button offered in jenkins and in some environments it makes the work much easier.
If there is a more secure way to fix this great. It just would be nice if this would be implemented in the next version,

PR #489, opened 2021-11-04
Integrate SECURITY-2458 changes

daniel-beck:
Context:
https://www.jenkins.io/doc/upgrade-guide/2.303/#SECURITY-2458
https://www.jenkins.io/doc/developer/security/remoting-callables/

Probably debatable whether we want to keep the new behavior to allow `RoleChecker#check(RoleSensitive)`; probably makes sense to amend this and just allow all remoting callables to ignore the role check (not just some). For now, this needs merging so that the released change is on `master` though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@ THE SOFTWARE.
 
   <groupId>org.jenkins-ci.main</groupId>
   <artifactId>remoting</artifactId>
-  <version>4.11.1</version>
+  <version>${revision}${changelist}</version>
 
   <name>Jenkins remoting layer</name>
   <url>https://github.com/jenkinsci/remoting</url>
@@ -58,11 +58,11 @@ THE SOFTWARE.
     <connection>scm:git:git://github.com/jenkinsci/remoting.git</connection>
     <developerConnection>scm:git:ssh://git@github.com/jenkinsci/remoting.git</developerConnection>
     <url>https://github.com/jenkinsci/remoting</url>
-    <tag>remoting-4.11.1</tag>
+    <tag>${scmTag}</tag>
   </scm>
 
   <properties>
-    <revision>4.12</revision>
+    <revision>4.11.2</revision>
     <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
     <build.type>private</build.type>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <revision>4.11.2</revision>
+    <revision>4.12</revision>
     <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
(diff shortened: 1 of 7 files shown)


jglick:
Can you resolve the merge conflict?

pom.xml
```diff
@@ -62,7 +62,7 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <revision>4.12</revision>
+    <revision>4.11.2</revision>
```
jglick, line 65: revert please

jglick (approved):

jglick:
Still needs merge conflict resolved.

jglick (approved):

jeffret-b (approved):

src/main/java/hudson/remoting/RequiredRoleCheckerWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jenkinsci.remoting.Role;
+import org.jenkinsci.remoting.RoleChecker;
+import org.jenkinsci.remoting.RoleSensitive;
+
+import java.util.Collection;
+
+/**
+ * Implementation that can tell whether a {@code #check} method was invoked.
+ */
+/* package-private */ class RequiredRoleCheckerWrapper extends RoleChecker {
```
jvz, line 35: This sort of security state checking/reporting seems like a pattern that should be replicated in future patches. I love this idea!

jvz (approved):

daniel-beck:
@jeffret-b @oleg-nenashev  

Looks like [I messed up the staging](https://issues.jenkins.io/browse/JENKINS-67227) because I forgot this is signed.

Proposal to clean this up towards 2.319.1 (post-RC) and the next weekly release, both of which use remoting 4.11.1:

We could 
1. restore the `v4.11.x` branch,
2. point a [remoting profile in the `jenkins-infra/release` repo](https://github.com/jenkins-infra/release/blob/master/profile.d/components/remoting) to that branch,
3. set the MBPL on release.ci up to use that branch (it currently only checks master),
4. and build it.

I expect we'd get 4.11.2 fairly easily and could still integrate that, assuming the pipeline works as expected.

Thoughts?

CC @cathychan @timja

jeffret-b:
Can't we just release 4.12 and have people use that if they care? Is there any reason we need a 4.11.2 specifically?

We could include #490 or not in that next release.

daniel-beck:
2.319.1 was RC last week, release next week. Ideally we'd not integrate random changes that happen to be on master in LTS.

jeffret-b:
If you can get the build environment to work that way, it's fine.

daniel-beck:
Thanks! Will look into that tomorrow.

daniel-beck:
Had to mess with the branch a bit because conflict resolution happened there. Too bad it was squashed, but we cannot do anything about that now...

PR #490, opened 2021-11-06
Allow all remoting callables to bypass role check

daniel-beck:
This removes the need to support `RoleChecker#check(RoleSensitive)`

Downstream PR: https://github.com/jenkinsci/jenkins/pull/5901

Corresponding docs update: https://github.com/jenkins-infra/jenkins.io/pull/4707

----

During development of the [SECURITY-2458 security hardening](https://www.jenkins.io/doc/book/security/controller-isolation/required-role-check/), we noticed some `Callable`s defined in remoting would be rejected; but they couldn't perform a real role check, because application-level roles aren't known to remoting.

Additionally, I thought it might be useful to allow calling `RoleChecker#check(RoleSensitive, Role...)` without any `Role`, to indicate that a `Callable` doesn't wish to limit where it can execute. Unfortunately this didn't work for `RPCRequest` because it would prevent classloading, as that's what exposes classloaders. Beyond that, there was some pushback about allowing explicit role checks without any roles in user code, so much so that [the developer docs](https://www.jenkins.io/doc/developer/security/remoting-callables/#mandatory-role-checks) say we'll suspend distribution of any plugin that does this (which isn't really helpful).

For 2.303.3 and 2.319, we ended up with a kind of hybrid approach: Some `Callable`s did the `RoleChecker#check(RoleSensitive)`, others were explicitly exempted from the role check requirement. And because of the former, we added support for `RoleChecker#check(RoleSensitive)` to core -- while documenting that nobody should use it.

----

This PR cleans this up: It restores the expectation that all user-defined `Callable`s perform a non-empty role check with an actual role. `Callable`s defined in remoting, unaware of application-level roles, do not need to perform a real role check. `InternalCallable` represents that that.

**Please squash-merge this PR.**

### DIFF
--- a/src/main/java/hudson/remoting/ChannelBuilder.java
+++ b/src/main/java/hudson/remoting/ChannelBuilder.java
@@ -48,16 +48,9 @@ public class ChannelBuilder {
     private static final Logger LOGGER = Logger.getLogger(ChannelBuilder.class.getName());
     private static /* non-final for Groovy */ boolean CALLABLES_CAN_IGNORE_ROLECHECKER = Boolean.getBoolean(ChannelBuilder.class.getName() + ".allCallablesCanIgnoreRoleChecker");
 
-    private static final Set<String> REMOTING_CALLABLES = new HashSet<>();
     private static final Set<String> SPECIFIC_CALLABLES_CAN_IGNORE_ROLECHECKER = new HashSet<>();
 
     static {
-        REMOTING_CALLABLES.add(RemoteInvocationHandler.RPCRequest.class.getName());
-        REMOTING_CALLABLES.add(RemoteInvocationHandler.UserRPCRequest.class.getName());
-        REMOTING_CALLABLES.add(PingThread.Ping.class.getName());
-        REMOTING_CALLABLES.add(Channel.SetMaximumBytecodeLevel.class.getName());
-        REMOTING_CALLABLES.add(Channel.IOSyncer.class.getName());
-
         final String propertyName = ChannelBuilder.class.getName() + ".specificCallablesCanIgnoreRoleChecker";
         final String property = System.getProperty(propertyName);
         if (property != null) {
@@ -312,8 +305,8 @@ public class ChannelBuilder {
             return false;
         }
 
-        if (REMOTING_CALLABLES.contains(callable.getClass().getName())) {
-            LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is a remoting built-in callable");
+        if (callable instanceof InternalCallable) {
+            LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is a remoting built-in callable allowed to bypass the role check");
         }
 
         if (SPECIFIC_CALLABLES_CAN_IGNORE_ROLECHECKER.contains(callable.getClass().getName())) {
--- a/src/main/java/hudson/remoting/ChannelBuilder.java
+++ b/src/main/java/hudson/remoting/ChannelBuilder.java
@@ -307,6 +307,7 @@ public class ChannelBuilder {
 
         if (callable instanceof InternalCallable) {
             LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is a remoting built-in callable allowed to bypass the role check");
+            return false;
         }
 
         if (SPECIFIC_CALLABLES_CAN_IGNORE_ROLECHECKER.contains(callable.getClass().getName())) {
--- a/src/main/java/hudson/remoting/ChannelBuilder.java
+++ b/src/main/java/hudson/remoting/ChannelBuilder.java
@@ -312,7 +312,7 @@ public class ChannelBuilder {
             return false;
         }
 
-        if (REMOTING_CALLABLES.contains(callable.getClass().getName()))) {
+        if (REMOTING_CALLABLES.contains(callable.getClass().getName())) {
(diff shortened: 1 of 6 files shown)


jglick:
> When updating the remoting dependency in core to include this change

Is there a matching draft PR using an incremental dep?

src/main/java/hudson/remoting/ChannelBuilder.java
```diff
@@ -48,9 +48,16 @@
     private static final Logger LOGGER = Logger.getLogger(ChannelBuilder.class.getName());
     private static /* non-final for Groovy */ boolean CALLABLES_CAN_IGNORE_ROLECHECKER = Boolean.getBoolean(ChannelBuilder.class.getName() + ".allCallablesCanIgnoreRoleChecker");
 
+    private static final Set<String> REMOTING_CALLABLES = new HashSet<>();
     private static final Set<String> SPECIFIC_CALLABLES_CAN_IGNORE_ROLECHECKER = new HashSet<>();
 
     static {
+        REMOTING_CALLABLES.add(RemoteInvocationHandler.RPCRequest.class.getName());
```
jglick, line 55: Maybe we can have all of these extend some internal (restricted) base class which implements the empty `checkRoles`? Would avoid the need for this enumeration.

src/main/java/hudson/remoting/PingThread.java
```diff
@@ -164,7 +164,6 @@ public Void call() throws IOException {
         @Override
         public void checkRoles(RoleChecker checker) throws SecurityException {
             // this callable is literally no-op, can't get any safer than that
-            // TODO: Apply `checker.check(this)` after release and make this type private again
```
jglick, line 167: …which would allow `private` to be added back.

src/main/java/hudson/remoting/InternalCallable.java
```diff
@@ -0,0 +1,34 @@
...
+ * THE SOFTWARE.
+ */
+package hudson.remoting;
+
+/**
+ * For remoting internal use only: {@link hudson.remoting.Callable}s implement
+ * this marker interface to be allowed to bypass the required role check of
+ * {@link hudson.remoting.RequiredRoleCheckerWrapper}.
+ *
+ * @since TODO
```
jglick, line 31:
```suggestion
```

Not an API, so no need for `@since`.

And add `@Restricted(NoExternalUse.class)`.

src/main/java/hudson/remoting/InternalCallable.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+/* package */ interface InternalCallable<V, T extends Throwable> extends Callable<V, T> {
+}
```
jglick, line 34: So why not implement `checkRoles` here? (`default` method)

src/main/java/hudson/remoting/InternalCallable.java
daniel-beck: I suggest keeping `@since`, may be useful even for internal stuff.

jglick: Generally prefer not to include it, since `git blame` is your friend for tracking behavioral changes if you are already inside the source tree. Not important either way.

src/main/java/hudson/remoting/ChannelBuilder.java
```diff
@@ -299,21 +299,14 @@ public void check(@NonNull RoleSensitive subject, @NonNull Collection<Role> expe
...
-            LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is an RPCRequest");
-            return false;
-        }
-
-        if (callable instanceof PingThread.Ping) {
-            // TODO Post-release, remove special treatment of this type and have it call RoleChecker#check
-            LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is a PingThread.Ping");
-            return false;
+        if (callable instanceof InternalCallable) {
+            LOGGER.log(Level.FINE, () -> "Callable " + callable.getClass().getName() + " is a remoting built-in callable allowed to bypass the role check");
```
jglick, line 309:
```suggestion
            LOGGER.fine(() -> callable.getClass().getName() + " is a remoting built-in callable allowed to bypass the role check");
```

src/main/java/hudson/remoting/RemoteInvocationHandler.java
```diff
@@ -851,7 +851,8 @@ private void onChannelTermination(Channel channel) {
      *
      * @see UserRPCRequest
      */
-    static class RPCRequest extends Request<Serializable,Throwable> implements DelegatingCallable<Serializable,Throwable> {
+    static class RPCRequest extends Request<Serializable,Throwable> implements DelegatingCallable<Serializable,Throwable>, InternalCallable<Serializable, Throwable> {
```
jglick, line 854:
```suggestion
    private static class RPCRequest extends Request<Serializable,Throwable> implements DelegatingCallable<Serializable,Throwable>, InternalCallable<Serializable, Throwable> {
```

jglick (approved):
Looks nicer indeed.

src/main/java/hudson/remoting/RemoteInvocationHandler.java
daniel-beck: Wasn't me: `UserRequest` uses this in https://github.com/jenkinsci/remoting/blob/92ecf07540c9e2d8ef3c76f8008afcc8a160e505/src/main/java/hudson/remoting/UserRequest.java#L199

jeffret-b (approved):

PR #495, opened 2021-12-17
Update dependencies, exclude JSR 305

basil:
- Bump Hamcrest from 1.3 to 2.2
- Bump Guava from 30.0.1-jre to 31.0.1-jre
- Bump Mockito from 1.10.19 to 4.2.0 (required various test updates)
- Exclude JSR 305 in favor of JCIP annotations (matching recent cores)
- Bump `access-modifier` from 1.25 to 1.27


oleg-nenashev (approved):
I am not 100% sure about using jcip-annotations, but I do not have better suggestions. It seems to be a common approach in this case

src/main/java/hudson/remoting/AsyncFutureImpl.java
```diff
@@ -87,7 +85,7 @@ public synchronized V get() throws InterruptedException, ExecutionException {
 
     @Override
     @CheckForNull
-    public synchronized V get(@Nonnegative long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+    public synchronized V get(/* @javax.annotation.Nonnegative */ long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
```
jeffret-b, line 88: We should just remove it rather than comment it out.

src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
```diff
@@ -236,8 +235,7 @@ public int readable() {
      * @return Number of bytes we can write to the buffer.
      *         If the buffer is closed, may return the value beyond the limit (JENKINS-37514)
      */
-    @Nonnegative
-    public int writable() {
+    /* @javax.annotation.Nonnegative */ public int writable() {
```
jeffret-b, line 238: And here we should just remove it also. Especially since it pushes the useful stuff so far right.

jglick:
Looks right; did you check if this snapshot can be included in core without causing POM problems?

basil:
> I am not 100% sure about using `jcip-annotations`, but I do not have better suggestions. It seems to be a common approach in this case.

Yes, this was the approach taken in jenkinsci/jenkins#4604. I am being consistent with that change here.

> did you check if this snapshot can be included in core without causing POM problems?

I did, and it can.

src/main/java/hudson/remoting/AsyncFutureImpl.java
basil: Feedback that would have also applied to jenkinsci/jenkins@5674de12353df95d1c5f0baaa699636ed47b7284. I have done as you suggested.

src/main/java/org/jenkinsci/remoting/nio/FifoBuffer.java
basil: Ditto

basil:
Any further action items here? I hope to merge this sometime this week if there are no additional changes to be made.

jtnord (commented):
this is including (shading) a lot more libraries that before - is this intentional?

### this pr

```
[INFO] --- maven-shade-plugin:3.2.4:shade (default) @ remoting ---
[INFO] Including org.glassfish.tyrus.bundles:tyrus-standalone-client-jdk:jar:1.18 in the shaded jar.
[INFO] Including args4j:args4j:jar:2.33 in the shaded jar.
[INFO] Including org.jenkins-ci:constant-pool-scanner:jar:1.2 in the shaded jar.
[INFO] Including com.github.spotbugs:spotbugs-annotations:jar:4.5.0 in the shaded jar.
[INFO] Including net.jcip:jcip-annotations:jar:1.0 in the shaded jar.
[INFO] Including org.kohsuke:access-modifier-annotation:jar:1.27 in the shaded jar.
[INFO] Including org.jenkins-ci:annotation-indexer:jar:1.15 in the shaded jar.
[WARNING] access-modifier-annotation-1.27.jar, annotation-indexer-1.15.jar, args4j-2.33.jar, constant-pool-scanner-1.2.jar, jcip-annotations-1.0.jar, remoting-4.12-SNAPSHOT.jar, spotbugs-annotations-4.5.0.jar, tyrus-standalone-client-jdk-1.18.jar define 1 overlapping resource:
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```


### master@a9938ac

```
[INFO] --- maven-shade-plugin:3.2.4:shade (default) @ remoting ---
[INFO] Including org.glassfish.tyrus.bundles:tyrus-standalone-client-jdk:jar:1.18 in the shaded jar.
[INFO] Including args4j:args4j:jar:2.33 in the shaded jar.
[INFO] Including org.jenkins-ci:constant-pool-scanner:jar:1.2 in the shaded jar.
[WARNING] args4j-2.33.jar, constant-pool-scanner-1.2.jar, remoting-4.12-SNAPSHOT.jar, tyrus-standalone-client-jdk-1.18.jar define 1 overlapping resource:
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```

basil:
Of course not and thank you. Fixed in the latest revision.

jtnord (approved):

jeffret-b (approved):
Thanks for the cleanup

PR #507, opened 2022-02-04
[JENKINS-67000] Restore setting of null security manager

jglick:
https://issues.jenkins.io/browse/JENKINS-67000

Reverts #481 pending a more nuanced fix.

### DIFF
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -246,6 +246,14 @@ public class Main {
      * Main without the argument handling.
      */
     public static void _main(String[] args) throws IOException, InterruptedException, CmdLineException {
+        // TODO skip this on Java 17+ (e.g. io.jenkins.lib.versionnumber.JavaSpecificationVersion) as it prints a warning
+        // otherwise needed for JavaWebStart agents (JENKINS-67000)
+        try {
+            System.setSecurityManager(null);
+        } catch (SecurityException e) {
+            // ignore
+        }
+
         // if we run in Mac, put the menu bar where the user expects it
         System.setProperty("apple.laf.useScreenMenuBar", "true");
 


basil:
As a temporary hotfix, OK.

timja (approved):
Tbh once we drop Java 8 we can just revert this I think

jglick:
> once we drop Java 8 we can just revert this

Only if we remove JWS support as well. You can run IcedTea on Java 11.

jglick:
Will leave it to someone who knows how to use the release bot to merge & release this. (Ironically, I think we could use regular JEP-229 for this component if it were not for the signature, which is only used by…JavaWebStart.)

timja:
> > once we drop Java 8 we can just revert this
> 
> Only if we remove JWS support as well. You can run IcedTea on Java 11.

I doubt people are going to the hassle of setting that up just for this

timja:
FTR we don't support it on Java 11 even with iced tea: https://github.com/jenkinsci/jenkins/blob/e27b310065b3c036b5fc9d123f1d1d99d3058c00/core/src/main/java/hudson/slaves/JNLPLauncher.java#L262-L265

MarkEWaite:
@jeffret-b when you have reviewed this and decide to merge it, let me know and I can assist with the release machinery.

MarkEWaite (approved):

jglick:
> we don't support it on Java 11 even with iced tea

https://github.com/jenkinsci/jenkins/pull/3766 indeed, which seems incorrect, not least because it is checking the version of Java running on the _controller_ which is irrelevant in this context.

:shrug: if we do not care to support IcedTea/OpenWebStart users on Java 11+, fine with me but ought to be a conscious decision mentioned in @MarkEWaite’s https://github.com/jenkinsci/jep/pull/382. At the same time we should consider removing the JAR signature here, which (AFAIK) serves no function in non-JWS usages of `remoting.jar` and just complicates infrastructure; and we can remove `slave-installer` and its implementations: https://github.com/jenkinsci/jep/blob/master/jep/230/README.adoc#marking-slave-installer-as-detached

jeffret-b:
We really should remove the Launch button and all support for WebStart. They cause more confusion and bad behavior than they help. I've tried a couple of times in the past, but encountered some opposition based on unknown usage of the feature. The low level of interest in this regression indicates that this would be an overall gain and would negatively impact few people. We definitely should not carry WS forward into Java 11, even though it's possible. There's also no reason why we need to wait for Java 8 support termination before we take this step.

jeffret-b (approved):
Since this was an accidental regression, it's fine to restore this operation.

Ultimately, we really should remove it.

jeffret-b:
Let's give this one the 24 hour waiting period for any additional comments or concerns.

jeffret-b:
@MarkEWaite , can you perform a release when you have a chance?

MarkEWaite:
> @MarkEWaite , can you perform a release when you have a chance?

Starting the release build now.

MarkEWaite:
@jeffret-b could you publish the changelog from [GitHub releases](https://github.com/jenkinsci/remoting/releases)?

PR #511, opened 2022-02-14
Clean up SpotBugs exclusions

basil:
Removes no-longer-necessary suppressions, and moves the still-necessary suppressions from `src/spotbugs/excludeFilter.xml` to `@SuppressFBWarnings` annotations in the relevant code (where they are more likely to be noticed and fixed) if there were a reasonable number of them.

### DIFF
--- a/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
+++ b/src/main/java/hudson/remoting/AbstractByteArrayCommandTransport.java
@@ -1,6 +1,5 @@
 package hudson.remoting;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.jenkinsci.remoting.util.AnonymousClassWarnings;
 
 import java.io.ByteArrayOutputStream;
@@ -18,7 +17,6 @@ import java.util.logging.Logger;
  * @author Kohsuke Kawaguchi
  * @since 2.13
  */
-@SuppressFBWarnings(value = "SIC_INNER_SHOULD_BE_STATIC_ANON", justification = "TODO needs triage")
 public abstract class AbstractByteArrayCommandTransport extends CommandTransport {
     protected Channel channel;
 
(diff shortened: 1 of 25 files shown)


jeffret-b (commented):
I'm not convinced about accepting all of the ones that say, "TODO needs triage". I'm not sure that changing things around to introduce that into new places provides much benefit.

basil:
Thanks for the review, Jeff. If this review is blocking, could you please give me a concrete action item that is needed in order to move forward? If this review is non-blocking, please explicitly state as much.

basil:
The stated goal of this PR is to

> move the still-necessary suppressions from `src/spotbugs/excludeFilter.xml` to `@SuppressFBWarnings` annotations in the relevant code (where they are more likely to be noticed and fixed)

@jeffret-b Any guidance on how this PR could be adjusted to satisfy this objective while addressing your concerns?

jeffret-b:
I've been busy and will be even more so for the next few days. 

But, I don't think the goal is satisfied by the PR as it currently stands. Some of these spotbugs issues that are marked for triage might be more likely to be noticed but that doesn't make them more likely to be fixed. Oleg tried some years back with some of them and ended up giving up. I tried a while back with some. Some of them would require changes to the remoting design and public API, both of which are problematic.

So, remove all the ones that say "TODO needs triage" and leave them in the includes file. The remaining work could still have some use.

(I closed Oleg's old PR where he tried to fix some of these some time ago for the same reason. Making it more visible wasn't doing anything to help it get fixed.)

jeffret-b (changes requested):
See my other comment about "TODO needs triage".

basil:
> There are differences between remoting and core. Remoting has had a number of pretty thorough passes at driving down existing spotbugs warnings as far as possible.

Core has recently, as well.

basil:
Has the feedback from [this review](https://github.com/jenkinsci/remoting/pull/511#pullrequestreview-881098946) and [this review](https://github.com/jenkinsci/remoting/pull/511#pullrequestreview-887779302) been addressed?

jeffret-b:
Been away from computer and then trying to catch up again.

jeffret-b (approved):
Looks fine

jeffret-b:
Let's go ahead and give it the 24 hours for any additional comments before I merge it.

PR #513, opened 2022-02-22
JENKINS-67928 Prevent infinite loop in case of closed SSL connection

gulyaev13:
**Background**
I've found increased CPU load at my Jenkins instance.
The root cause of CPU load was a thread that is stuck in an infinite loop in SSLEngineFilterLayer.java
The remote agent was already terminated by kubernetes-plugin and the SSL connection is closed, but the thread is still working because can't exit from the while loop.

**Solution**
Set _done_ flag in case of receiving `SSLEngineResult.Status.CLOSED` from `sslEngine.unwrap(tempBuffer, appBuffer)`.
This change is allowed to return execution flow to `NIONetworkLayer.ready()` where the thread will handle ClosedChannelException.

**Libraries versions**
Jenkins core - 2.319.3
kubernetes - 1.31.3
remoting - 4.11.2

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/impl/SSLEngineFilterLayer.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/impl/SSLEngineFilterLayer.java
@@ -348,6 +348,8 @@ public class SSLEngineFilterLayer extends FilterLayer {
                     switch (result.getStatus()) {
                         case BUFFER_UNDERFLOW:
                     /* we need more data */
+                        case CLOSED:
+                    /* connection is already closed */
                             done = true;
                             break;
                         case BUFFER_OVERFLOW:
@@ -367,8 +369,6 @@ public class SSLEngineFilterLayer extends FilterLayer {
                                 next().onRecv(appBuffer);
                                 appBuffer.clear();
                             }
-                            break;
-                        case CLOSED:
                             break;
                     }
                     handshakeStatus = sslEngine.getHandshakeStatus();


timja:
Have you been able to test this yourself and see if it works?

No experience with this protocol

gulyaev13:
@jglick, @timja, I've patched the remoting version via custom-war-packager and tested change at my local instance.

```yaml
war:
  groupId: "org.jenkins-ci.main"
  artifactId: "jenkins-war"
  source:
    version: 2.319.3
libPatches:
  - groupId: "org.jenkins-ci.main"
    artifactId: "remoting"
    source:
      git: https://github.com/gulyaev13/remoting.git
      commit: b2a29ed6e5445e89d2aaf5de5fa08afd2ef46e5c
```

CPU utilization with remoting 4.11.2:
Снимок экрана 2022-02-28 в 19 17 37

CPU utilization with my patch for remoting:
Снимок экрана 2022-02-28 в 19 59 24

As you can see there are several peaks in the same moments of time, but with the patch, CPU resources are released and don't freeze at any value for a long time. It means that a fix solves the issue.
Also, I don't find any problems with common functionality - TCP/NIO transport works as expected.

timja:
@jeffret-b any chance you can take a look?

jeffret-b:
This needs a tracking ticket in Jira. Could you please create one, @gulyaev13 ?

jeffret-b (approved):
This makes sense. If the result is CLOSED, there should be no reason to continue processing. The testing provided by @gulyaev13 is hopeful. At the worst, it shouldn't harm any regular situations. And it should help in some cases of a closed connection.

Approving as long as we have a Jira tracking ticket connected to it in case someone finds a problem based upon some other scenario.

timja:
This PR is now ready for merge, after ~24 hours, we will merge it if there's no negative feedback.

Thanks!

PR #514, opened 2022-02-27
JENKINS-66446 Agent reconnects on Java 11 and websocket

timja:
Extracts the logic from https://github.com/jenkinsci/remoting/pull/295 which wasn't re-used in the websocket protocol

See https://issues.jenkins.io/browse/JENKINS-66446

To reproduce:

1. Use Java 11 on the agent
2. Create a Permanent agent with websocket protocol enabled
3. Restart the controller
4. It should fail to reconnect the agent with a `NoClassDefFoundError: jenkins/slaves/restarter/JnlpSlaveRestarterInstaller`
5. Repeat steps with this patch, reconnects just fine

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -683,10 +683,21 @@ public class Engine extends Thread {
                     }
                     TimeUnit.SECONDS.sleep(10);
                 }
-                events.onReconnect();
+                reconnect();
             }
         } catch (Exception e) {
             events.error(e);
+        }
+    }
+
+    private void reconnect() {
+        try {
+            events.status("Performing onReconnect operation.");
+            events.onReconnect();
+            events.status("onReconnect operation completed.");
+        } catch (NoClassDefFoundError e) {
+            events.status("onReconnect operation failed.");
+            LOGGER.log(Level.FINE, "Reconnection error.", e);
         }
     }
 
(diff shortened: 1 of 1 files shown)


jglick:
Ought to be possible to reproduce this in a functional test in core I guess?

jglick (approved):
It is correct to apply the same logic to WS mode as TCP mode. The original fix seems wrong. If you look at the stack trace, it seems that `JnlpSlaveRestarterInstaller` is _supposed_ to be running in this situation but is failing because it relies on remote class loading, and by the time we get here the channel is dead and it is too late to do that. You can see from https://javadoc.jenkins.io/jenkins/slaves/restarter/SlaveRestarter.html#restart-- that code is prepared for this possibility, but it is not working currently. I suspect https://github.com/jenkinsci/jenkins/pull/3333 inadvertently broke this by causing `JnlpSlaveRestarterInstaller` to no longer be loaded on the agent during _connection_ yet still sending over a `EngineListenerAdapter` which requires `JnlpSlaveRestarterInstaller` to be loaded (even if unused) during _disconnection_; and perhaps the Java 11 class linkage system behaves differently here, due to the “nest” system? (The inner class ought to refer only to its parent `static` nested class but apparently on Java 11 it refers to the outermost class as well?)

At any rate, I think this patch is OK, though I would probably log at `WARNING` rather than `FINE`, and the bug is in core.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -683,13 +683,24 @@ public void closeRead() throws IOException {
...
     }
 
+    private void reconnect() {
+        try {
+            events.status("Performing onReconnect operation.");
+            events.onReconnect();
+            events.status("onReconnect operation completed.");
+        } catch (NoClassDefFoundError e) {
+            events.status("onReconnect operation failed.");
+            LOGGER.log(Level.FINE, "Reconnection error.", e);
```
timja, line 700:
```suggestion
            LOGGER.log(Level.WARNING, "Reconnection error.", e);
```

basil (approved):
This PR is now ready for merge. We will merge it after approximately 24 hours if there is no negative feedback. Please see the [merge process documentation](https://github.com/jenkinsci/jenkins/blob/master/docs/MAINTAINERS.adoc#merge-process) for more information about the merge process. Thanks!

jeffret-b (approved):
Looks like it should help in this situation also. We've seen less reports of the problem of the other scenarios since the earlier change.

MarkEWaite (approved):

PR #531, opened 2022-05-05
Don't set null SecurityManager on Java 17

offa:
Sets a null Security Manager on Java < 17 only.

- #507 
- #481
- [JENKINS-67000](https://issues.jenkins.io/browse/JENKINS-67000)
- [JENKINS-67913](https://issues.jenkins.io/browse/JENKINS-67913)

--------------------------------

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -181,6 +181,11 @@ THE SOFTWARE.
       <version>9.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.10</version>
+    </dependency>
   </dependencies>
 
   <repositories>
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -249,6 +249,7 @@ public class Main {
      * Main without the argument handling.
      */
     public static void _main(String[] args) throws IOException, InterruptedException, CmdLineException {
+        // TODO skip this on Java 17+ as it prints a warning, otherwise needed for JavaWebStart agents (JENKINS-67000, JENKINS-67913)
         if (JavaSpecificationVersion.forCurrentJVM().isOlderThan(new VersionNumber("17"))) {
             try {
                 System.setSecurityManager(null);
--- a/src/main/java/hudson/remoting/jnlp/Main.java
+++ b/src/main/java/hudson/remoting/jnlp/Main.java
@@ -31,6 +31,9 @@ import hudson.remoting.EngineListener;
 import hudson.remoting.FileSystemJarCache;
 import hudson.remoting.Util;
 import java.util.Map;
+
+import hudson.util.VersionNumber;
+import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
 import org.jenkinsci.remoting.engine.WorkDirManager;
 import org.jenkinsci.remoting.util.PathUtils;
 import org.kohsuke.args4j.Argument;
@@ -246,12 +249,12 @@ public class Main {
      * Main without the argument handling.
      */
     public static void _main(String[] args) throws IOException, InterruptedException, CmdLineException {
-        // TODO skip this on Java 17+ (e.g. io.jenkins.lib.versionnumber.JavaSpecificationVersion) as it prints a warning
-        // otherwise needed for JavaWebStart agents (JENKINS-67000)
-        try {
-            System.setSecurityManager(null);
-        } catch (SecurityException e) {
-            // ignore
+        if (JavaSpecificationVersion.forCurrentJVM().isOlderThan(new VersionNumber("17"))) {
+            try {
+                System.setSecurityManager(null);
+            } catch (SecurityException e) {
+                // ignore
+            }
         }
 
         // if we run in Mac, put the menu bar where the user expects it


pom.xml
```diff
@@ -181,6 +181,11 @@ THE SOFTWARE.
       <version>9.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
```
jglick, line 186: Probably fine but we will see when this is integrated into core whether there is any clash with the version used by core. There are also integration tests in core which `java -jar agent.jar` so any shading/packaging issues should be smoked out.

src/main/java/hudson/remoting/jnlp/Main.java
```diff
@@ -246,12 +249,12 @@ public static void main(String[] args) throws IOException, InterruptedException
      * Main without the argument handling.
      */
     public static void _main(String[] args) throws IOException, InterruptedException, CmdLineException {
-        // TODO skip this on Java 17+ (e.g. io.jenkins.lib.versionnumber.JavaSpecificationVersion) as it prints a warning
-        // otherwise needed for JavaWebStart agents (JENKINS-67000)
```
jglick, line 250: I think you want to keep some comments explaining the context, specifically https://github.com/jenkinsci/remoting/pull/507#issuecomment-1030341067

jglick (approved):

basil (commented):
My understanding is that avoiding the call to `System.setSecurityManager(null)` on Java 17+ will eliminate the warning but will also remove support for Java Web Start agents, as it did on Java 11 in [JENKINS-67000](https://issues.jenkins.io/browse/JENKINS-67000). In other words, since Java Web Start remains an option offered in the GUI on all Java versions (including Java 17+), this PR will just reintroduce JENKINS-67000 (but for Java 17 rather than Java 11).

Just because Java 17+ support is in preview does not mean we encourage the reintroduction of bugs. I think that in order to move forward with this PR we would either need to formally deprecate and then remove the Java Web Start feature when running on Java 17 (including removing the option in the GUI). Alternatively, we could find a way to implement the Web Start feature on Java 17 without the call to `System.setSecurityManager(null)`.

I have not investigated the latter yet, but certainly retaining support for this feature would not break any existing use cases. I am not particularly against the former, but there are at least some users for whom this is a production use case, so we would need to communicate with them in advance through a blog post and/or deprecation warning to ensure they are not surprised by a sudden change.

timja:
JEP-236 says that we will be removing support as of Java 11.
https://github.com/jenkinsci/jep/tree/master/jep/236#library-updates

basil:
> JEP-236 says that we will be removing support as of Java 11. https://github.com/jenkinsci/jep/tree/master/jep/236#library-updates

I had no part in the writing of JEP-236 and will not comment on it. My existing comments above remain. It is fine to deprecate and subsequently remove a feature so long as there is a cogent justification, this is communicated to users, and the removal happens at all levels (including GUI). In this case, deprecation has not yet been done, there is no communication plan, and the GUI remains, so it is premature to remove implementation support for this feature.

pom.xml
offa: Core uses the same version (1.10) according to a quick GH search.

offa:
Thanks @basil, I didn't have this on my radar. Since it's no longer shipped I expected no one to actually continue using it.

However, this complicates things … and I honestly have no idea how to continue here.

src/main/java/hudson/remoting/jnlp/Main.java
offa: Indeed, added.

jglick:
> Java Web Start remains an option offered in the GUI on all Java versions (including Java 17+)

As noted in https://github.com/jenkinsci/remoting/pull/507#issuecomment-1030341067 the version check is done on the controller, which is somewhat off base: if you run the controller on Java 8 you would still be offered the option to use JWS for agents, which would now break (I suppose) if your agent happened to use a 17 JRE and you had downloaded IcedTea or similar. A pretty obscure scenario but some users might be affected.

Options:
* Take the opportunity to formally remove JWS support for good. That means cleaning up some other stuff in `remoting` (GUI mode, JAR signature) and `jenkins` (link to launch via `javaws`), deprecating `*slave-installer-module` and removing these from the WAR, and properly announcing this as a breaking change with a migration guide.
* Somehow detect that the agent is being run from `javaws` (is there a system property you can introspect?) and retain the nulling out of the `SecurityManager` even on 17. For now this will only print a console warning—which is not visible from a JWS GUI launch anyway. Kicks the can down the road a bit, since a subsequent Java version will likely throw an error.

MarkEWaite:
As a related aside, https://www.jenkins.io/doc/administration/requirements/upgrade-java-guidelines/#java-web-start states that Java Web Start is no longer available with Jenkins when the controller is running Java 11.  Not a blog post, but it has been publicly stated for a while

basil:
bq. Take the opportunity to formally remove JWS support for good. That means cleaning up some other stuff in remoting (GUI mode, JAR signature) and jenkins (link to launch via javaws), deprecating *slave-installer-module and removing these from the WAR, and properly announcing this as a breaking change with a migration guide.

Yeah, I agree that all of these are required for a proper removal. But the very minimum is to remove it from the GUI. All the rest is code cleanup that can and should be done, but breaking the implementation without removing the link from the GUI feels wrong to me, since that is exposing users to broken functionality.

timja:
@basil it’s only shown to users running on Java 8 to my knowledge

basil:
A point already addressed in https://github.com/jenkinsci/remoting/pull/531#issuecomment-1118836708 AFAICT.

jglick (changes requested):
I think we may as well proceed with #532.

jglick:
Up to you. Until it is approved and merged, the discussion on how best to proceed remains open.

offa:
Ok, so I'll leave open here for now. Feel free to close at any time (especially once the other PR is ready / merged).

PR #532, opened 2022-05-06
Removing JavaWebStart support

jglick:
As yet untested. Supersedes #531 (thus also #481 & #507).

PR #536, opened 2022-05-12
Push changes as part of release

timja:
ref https://github.com/jenkinsci/remoting/pull/532#issuecomment-1125014644

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -326,12 +326,6 @@ THE SOFTWARE.
           </execution>
         </executions>
       </plugin>
-      <plugin>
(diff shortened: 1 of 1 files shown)


jglick (commented):
I would delete the whole `maven-release-plugin` config section—use defaults from the parent POM. If you wish to use trusted.ci to release, fine, but if it has some special procedure other than `mvn -B release:{prepare,perform}` with adequate environmental credentials to GitHub & Artifactory, then it is responsible for passing corresponding `-D…` arguments to the command line itself.

jglick (approved):

MarkEWaite (approved):

PR #537, opened 2022-05-16
[JENKINS-68542] [WebSocket] Capture cookie returned by server and reuse for reconnection attempts

Vlatombe:
[JENKINS-68542](https://issues.jenkins.io/browse/JENKINS-68542)

This follows a similar logic to what is done for the [TCP path](https://github.com/jenkinsci/remoting/blob/6f32e99562f734e66f024aa44b777a3fdd6e9292/src/main/java/hudson/remoting/Engine.java#L1103-L1109).

* Build the header map before entering the infinite loop.
* Capture the cookie as processing the initial response and put it in the header map
* Further reconnection attempts use the updated header map.

Tested manually by launching a controller, then attaching an agent using websockets.
* Then start a second instance of the same agent => it gets rejected.
* Then start again the second instance by injecting the cookie through a system property => the second instance connects, the first instance gets disconnected. It tries to reconnect but gets rejected (I'm on mac, and it uses https://github.com/jenkinsci/jenkins/blob/e27b310065b3c036b5fc9d123f1d1d99d3058c00/core/src/main/java/jenkins/slaves/restarter/UnixSlaveRestarter.java#L24 bypassing this logic)

The initial use case that triggered this fix was a windows agent getting disconnected and unable to reconnect due to the controller not seeing it has disconnected. It doesn't use any `SlaveRestarter` implementation, so it should just reuse the same cookie.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -546,22 +546,23 @@ public class Engine extends Thread {
     @SuppressFBWarnings(value = {"REC_CATCH_EXCEPTION", "URLCONNECTION_SSRF_FD"}, justification = "checked exceptions were a mistake to begin with; connecting to Jenkins from agent")
     private void runWebSocket() {
         try {
+            String localCap = new Capability().toASCII();
+            final Map<String, List<String>> addedHeaders = new HashMap<>();
+            addedHeaders.put(JnlpConnectionState.CLIENT_NAME_KEY, Collections.singletonList(agentName));
+            addedHeaders.put(JnlpConnectionState.SECRET_KEY, Collections.singletonList(secretKey));
+            addedHeaders.put(Capability.KEY, Collections.singletonList(localCap));
+            if (webSocketHeaders != null) {
+                for (Map.Entry<String, String> entry : webSocketHeaders.entrySet()) {
+                    addedHeaders.put(entry.getKey(), Collections.singletonList(entry.getValue()));
+                }
+            }
             while (true) {
                 AtomicReference<Channel> ch = new AtomicReference<>();
-                String localCap = new Capability().toASCII();
                 class HeaderHandler extends ClientEndpointConfig.Configurator {
                     Capability remoteCapability = new Capability();
                     @Override
                     public void beforeRequest(Map<String, List<String>> headers) {
-                        headers.put(JnlpConnectionState.CLIENT_NAME_KEY, Collections.singletonList(agentName));
-                        headers.put(JnlpConnectionState.SECRET_KEY, Collections.singletonList(secretKey));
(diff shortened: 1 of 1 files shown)


Vlatombe:
Noting that this is only used if there is no [`SlaveRestarter`](https://github.com/jenkinsci/jenkins/blob/632033d6ada7c04bbc77ae710dc0a4635fdf9baa/core/src/main/java/jenkins/slaves/restarter/SlaveRestarter.java) implementation registered. Otherwise the whole process starts again and a new cookie is generated.
The customer problem was reported with windows agents, not running with winsw, since I didn't see any mention of it in the log sequence.

aHenryJard (commented):
Do you know if this part of code is already covered by unit test or ATH or if it's possible to write one ?

Vlatombe:
There are end-to-end tests in https://github.com/jenkinsci/jenkins/blob/5f9100286be16c434e973efc453a0b35ceee9342/test/src/test/java/jenkins/agents/WebSocketAgentsTest.java however the specific use case here I was not able to reproduce on my own. Requires some proxy in the middle that would close the connection only on agent side, while keeping it open on the controller side.

timja (approved):

Vlatombe:
Yeah, I thought of removing the installer dynamically as well from extensions list. This would likely cause the two agent instances to continuously flip. In any case this is not a realistic use case.

jglick:
> this is not a realistic use case

I actually never figured out the justification for the cookie system to exist. We already have an agent secret, so it does not seem like it adds any level of security to require that cookies match during a reconnection. I would think it is the responsibility of whatever service / script / package / container / etc. that launches the agent process to only attempt to run one copy at a time, and if due to some race condition a second copy tries to connect while there is still an open connection from the first copy, the desired behavior is unclear but you may as well throw out the first connection under the assumption it has been superseded. And if you _are_ using a native agent restarter, my understanding from your testing notes is that it makes a fresh connection and asks for a new cookie anyway? May be one of those features that was actually helpful a decade ago under different conditions and has just been maintained because no one wanted to bother proposing its deletion.

Vlatombe:
https://github.com/jenkinsci/jenkins/commit/2524af13872897c6ab16001c71765b8477b9ab2e & https://github.com/jenkinsci/jenkins/commit/71cbe0cc7c601c04509faa618b23194335288fee for context.

Actually if you were to delete this code, and if for some reason there is a second copy of the agent starting, you would end up bouncing infinitely between the two copies. I prefer the current behaviour where the second copy clearly gets rejected.

jglick:
Actually seems to date back further, FTR: https://github.com/jenkinsci/jenkins/commit/68ed742227891a3f716e4e479388c36876bb935a https://issues.jenkins.io/browse/JENKINS-5055

PR #542, opened 2022-05-24
[JENKINS-68542] Rename the header used for connection cookie

Vlatombe:
Follows-up #537, needed for https://github.com/jenkinsci/jenkins/pull/6576

The previous value was not a valid HTTP header name and could get
filtered by reverse proxies such as nginx, causing the previous fix not to work when using such setup.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -110,9 +110,9 @@ public class Engine extends Thread {
     public static final String REMOTING_MINIMUM_VERSION_HEADER = "X-Remoting-Minimum-Version";
 
     /**
-     * The property name for the cookie name key for websockets.
+     * The header name to be used for the connection cookie when using websockets.
      */
-    public static final String WEBSOCKET_COOKIE_KEY = "Connection-Cookie";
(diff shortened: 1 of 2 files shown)


jglick (commented):
This might break compatibility for TCP agents: someone using a different version of Remoting than is bundled in core would not see agents reconnect properly. Of course this is mainly an issue for hand-managed Windows agents (since on most platforms the agent restarter will kick in and override this system IIUC), and we generally recommend (but do not require) that you align Remoting versions, so it may be OK to do this with a release note.

Vlatombe:
I had similar thoughts. In the end, for the purpose of backporting this to current LTS, it may be better to keep backward compatibility.

Vlatombe:
> Change WS code on both sides to use a different HTTP header name. Makes code a bit more complex since we would need to special-case this header rather than simply map all HTTP headers to connection properties.

Ended up doing this. Actually not all http headers were mapped so I think it keeps the solution simple enough.

jglick (approved):

jglick (approved):

PR #554, opened 2022-06-29
Ignore reflection exception for Java > 8

lgemeinhardt:
Ignore reflection exception for Java > 8 to make selenium plugin usable for Java 11.

Jenkins moves to Java 11 👍 but the jenkins plugin (https://plugins.jenkins.io/selenium/) does not work with Java 11 (see https://support.cloudbees.com/hc/en-us/articles/360031995252-Selenium-plugin-issue-in-Java-11-SSH-connection-closed) 👎 

This fix catches the reflection exception and enables the selenium plugin again 👍 

Sure a better fix, would be to update somehow the selenium plugin, but this is not maintained anymore and the more important, it uses an official path (see below). So this fix may enables more plugins to work with Java 11. The path is the following:

hudson.plugins.selenium.process.SeleniumProcessUtils calls hudson.slaves.Channels.newJVM (jenkins-core)
hudson.slaves.Channels.newJVM calls hudson.remoting.Launcher (using the -cp argument and this fails for Java > 8)

- [x] Link to relevant issues in GitHub or Jira (linked to support.cloudbees.com)

### DIFF
--- a/src/main/java/hudson/remoting/Launcher.java
+++ b/src/main/java/hudson/remoting/Launcher.java
@@ -150,8 +150,13 @@ public class Launcher {
         Method $addURL = URLClassLoader.class.getDeclaredMethod("addURL", URL.class);
         $addURL.setAccessible(true);
 
-        for(String token : pathList.split(File.pathSeparator))
-            $addURL.invoke(ClassLoader.getSystemClassLoader(),new File(token).toURI().toURL());
+        try {
+            for(String token : pathList.split(File.pathSeparator))
+                $addURL.invoke(ClassLoader.getSystemClassLoader(),new File(token).toURI().toURL());
+        } catch (IllegalArgumentException ex) {
+            // looks like Java > 8 ... ignore it ...
+            LOGGER.log(Level.WARNING, "Cannot do reflection - ignore it...", ex);
+        }
 
         // fix up the system.class.path to pretend that those jar files
         // are given through CLASSPATH or something.


Vlatombe:
I don't know the selenium plugin, but if it calls the remoting launcher using `-cp` option, it probably has its reasons. Ignoring it on java > 8 will likely cause surprises downstream, it doesn't seem to be the right way to deal with the problem.

Going for the least surprise, if that option is known not to work on Java > 8, it should fail fast with a clear error message (prompting the caller to stop using that option).

lgemeinhardt:
@Vlatombe I could understand your concerns (non fail fast could hide issues). 
But Java > 8 will fail (because the system classloader is not an URLClassLoader anymore) without this fix (see below - before state) and the selenium plugin uses the -cp (and it's working) 👍 (see below - after state)
I assume the ```System.setProperty("java.class.path",System.getProperty("java.class.path")+File.pathSeparatorChar+pathList);``` does the classpath setting for Java > 8.

before state (w/o the patch):
```
Starting Selenium Grid
[jenkins_working_dir] $ E:\Java\jdk-11.0.15+10\bin\java -cp E:\jenkins_working_dir\war\WEB-INF\lib\remoting-4.13.2.jar hudson.remoting.Launcher -cp E:\jenkins_working_dir\plugins\selenium\WEB-INF\lib\selenium-server-standalone-3.141.59.jar;E:\jenkins_working_dir\plugins\selenium\WEB-INF\lib\htmlunit-driver-standalone-2.33.3.jar -connectTo localhost:35785
Jun 28, 2022 1:22:09 PM hudson.remoting.Launcher addClasspath
WARNING: [JENKINS-64831] Passing -cp to hudson.remoting.Launcher is deprecated and may not work at all on Java 9+
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by hudson.remoting.Launcher (file:/E:/jenkins_working_dir/war/WEB-INF/lib/remoting-4.13.2.jar) to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of hudson.remoting.Launcher
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Exception in thread "main" java.lang.IllegalArgumentException: object is not an instance of declaring class
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at hudson.remoting.Launcher.addClasspath(Launcher.java:154)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.spi.MethodSetter.addValue(MethodSetter.java:47)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.spi.StringOptionHandler.parseArguments(StringOptionHandler.java:19)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.CmdLineParser.parseArgument(CmdLineParser.java:520)
	at hudson.remoting.Launcher.main(Launcher.java:292)

```

after state (locally patched):
```
Starting Selenium Grid
[jenkins_working_dir] $ E:\Java\jdk-11.0.15+10\bin\java -cp E:\jenkins_working_dir\war\WEB-INF\lib\remoting-4.13.2.jar hudson.remoting.Launcher -cp E:\jenkins_working_dir\plugins\selenium\WEB-INF\lib\selenium-server-standalone-3.141.59.jar;E:\jenkins_working_dir\plugins\selenium\WEB-INF\lib\htmlunit-driver-standalone-2.33.3.jar -connectTo localhost:10932
Jun 29, 2022 2:08:53 PM hudson.remoting.Launcher addClasspath
WARNING: [JENKINS-64831] Passing -cp to hudson.remoting.Launcher is deprecated and may not work at all on Java 9+
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by hudson.remoting.Launcher (file:/E:/jenkins_working_dir/war/WEB-INF/lib/remoting-4.13.2.jar) to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of hudson.remoting.Launcher
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Jun 29, 2022 2:08:53 PM hudson.remoting.Launcher addClasspath
WARNING: Cannot do reflection - ignore it...
java.lang.IllegalArgumentException: object is not an instance of declaring class
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at hudson.remoting.Launcher.addClasspath(Launcher.java:155)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.spi.MethodSetter.addValue(MethodSetter.java:47)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.spi.StringOptionHandler.parseArguments(StringOptionHandler.java:19)
	at io.jenkins.remoting.shaded.org.kohsuke.args4j.CmdLineParser.parseArgument(CmdLineParser.java:520)
	at hudson.remoting.Launcher.main(Launcher.java:302)

Starting Selenium Grid
channel started
Jun 29, 2022 2:09:03 PM hudson.plugins.selenium.HubLauncher call
OFF: Grid hub starting with log level INFO
Jun 29, 2022 2:09:03 PM hudson.plugins.selenium.HubLauncher call
OFF: Grid Hub preparing to start on port 4444
Jun 29, 2022 2:09:05 PM org.seleniumhq.jetty9.util.log.Log initialized
INFO: Logging initialized @12493ms to org.seleniumhq.jetty9.util.log.Slf4jLog
Jun 29, 2022 2:09:06 PM org.seleniumhq.jetty9.server.Server doStart
...
```

jglick:
> a better fix, would be to update somehow the selenium plugin, but this is not maintained anymore

https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/

lgemeinhardt:
@jglick Thanks for your comment and I know that the problem could be fixed in the plugin, but I the plugins uses some (non deprecated) hudson.slaves.Channels.newJVM method (from the jenkins-core package) and this calls (https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/Channels.java#L226) the hudson.remoting.Launcher (using the -cp argument and this fails for Java > 8) 👎 

Looks that a fix within the plugin is may not that trivial (and need a alternative for the jenkins-core call) 👎  
But I could clearly understand, that some catch and log code is not the good solution (would reject them normally too), but do you think there is any way to get this fixed in the Launcher (may using some other way)?

basil:
I think you identified the path forward in https://github.com/jenkinsci/remoting/pull/554#issuecomment-1169983969: adopt the plugin and remove the offending caller rather than ignoring the call in Remoting. If this call was not needed in the first place, great! If this call was needed for some reason, that is something the plugin maintainer can discuss with the core developers and I am sure we could find a path forward. In the absence of someone stepping up to adopt this plugin, I plan to mark it deprecated within the next month or two, as it clearly does not work on Java 11.

lgemeinhardt:
@basil Thanks for your comment and I know that the problem could be fixed in the plugin, but I the plugins uses some (non deprecated and official) hudson.slaves.Channels.newJVM method (from the **jenkins-core package**) and this calls (https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/Channels.java#L226) the hudson.remoting.Launcher (using the -cp argument and this fails for Java > 8) 😕 

Looks that a fix within the plugin is may not that trivial (and need a alternative for the jenkins-core call) 😕 In addition all other callers of the hudson.slaves.Channels.newJVM method will fail too, if they use the classpath...

jglick:
Apparently `Channels.newJVM` needs to be deprecated too. Was forgotten.

jglick:
> do you think there is any way to get this fixed in the Launcher

No. The `-cp` argument is fundamentally wrong. Any plugin relying on this needs to be redesigned a bit. It could use a regular agent launch (`java -jar agent.jar` etc.) and then at runtime, if and when needed, create a new `URLClassLoader` with some libraries and load stuff from it reflectively.

lgemeinhardt:
@jglick Thanks for the clarification. 
I had the impression (after reading https://github.com/jenkinsci/remoting/commit/946aee295f88c70fb1e69c9ae70d2fe514e92c0d) that the deprecation was triggered by the Java > 8 issue. 
But now knowing, that the deprecation was triggered by a “design issue”, it makes perfectly sense to leave the code like this and use the situation (Jenkins Java 11 force until September 2022) to get rid of this code 😊

One last question, before I will close this PR (because it makes no sense, knowing this background) – does it makes sense to check if `hudon.slaves.Channels` could be “fixed” (by converting the deprecated -cp to a "normal" classpath) here https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/Channels.java#L226 (`vmb.args().add("-cp").add(classpath);`) with code like this `java.util.Arrays.stream(classpath.toString().split(java.io.File.pathSeparator)).forEach(arg -> vmb.classpath().add(arg));`? 
If so, I will check, if it works and file a PR for it.

jglick:
A `Channels.newJVM` overload taking no classpath-related arguments would be fine, if there are some plugins which would actually use it. (I am frankly not sure what this method is even for. Smells like a weird design that should be avoided.) The point is that the whole concept of injecting a classpath entry into the agent process is misguided.

lgemeinhardt:
@jglick Thanks for the quick reply. 
I opened a PR for the new `Channels.newJVM` change (after I tested it locally) https://github.com/jenkinsci/jenkins/pull/6723. 
What I see from the selenium plugin (the only user of this I saw so far, but this was onyl, because it did not work 👎 ) is that this is used to start the selenium grid as external process and communicate from Jenkins with it, to update the UI. 
Not sure if there are more plugins using a similar way, but may "Hadoop", because it is an example here https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/Channels.java#L197.

PR #558, opened 2022-07-07
JENKINS-68954 Agent with remoting 4.14+ hangs up or show response time >5s

bulanovk:
https://issues.jenkins.io/browse/JENKINS-68954 amends #523

Replace AtmostOneThreadExecutor with  Executors.newFixedThreadPool to avoid performance degradation and agents hang up

PR #523, opened 2022-04-04
[JENKINS-65873] Throttle anonymous class warnings to a single thread

Vlatombe:
Split from #505.

Anonymous class check is done async and is not a critical process, so it shouldn't use more than one thread at a time.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #259, opened 2018-03-07
[JENKINS-49994] Infrastructure for warning about serialization of anonymous inner classes

jglick:
[JENKINS-49994](https://issues.jenkins-ci.org/browse/JENKINS-49994)

@reviewbybees

PR #505, opened 2022-01-28
[JENKINS-65873] Use at most one thread to check classes

Vlatombe:
The current implementation reuses the ExecutorService from the channel, which is unbounded.

With a lot of classes to check on agent initialization, this can cause random errors like `java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached`

As there may be multiple factors at play here, this is adding a separate thread pool to perform class check. Backed by system property `org.jenkinsci.remoting.util.AnonymousClassWarnings.useSeparateThreadPool`.

Round 2: Since this doesn't seem to be the only factor, I have also added the system property `hudson.remoting.Engine.threadPoolSize`. Given an integer, it will replace the default unbounded cached thread pool by a fixed size based thread pool. Example `-Dhudson.remoting.Engine.threadPoolSize=10`. In case an OOM is raised, the current number of threads should be dumped in the stacktrace directly.

[JENKINS-65873](https://issues.jenkins.io/browse/JENKINS-65873)

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #559, opened 2022-07-07
[JENKINS-65873] [JENKINS-68954] Optimize `AnonymousClassWarnings`

jglick:
Amending #259, reverting #523, superseding #558,  and probably fixing [JENKINS-68954](https://issues.jenkins.io/browse/JENKINS-68954) and fixing [JENKINS-65873](https://issues.jenkins.io/browse/JENKINS-65873) more thoroughly and simply.

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
+++ b/src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
@@ -26,11 +26,8 @@ package org.jenkinsci.remoting.util;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.remoting.AtmostOneThreadExecutor;
 import hudson.remoting.Channel;
 
-import hudson.remoting.DaemonThreadFactory;
-import hudson.remoting.NamingThreadFactory;
 import java.io.IOException;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
@@ -38,7 +35,6 @@ import java.net.URL;
 import java.security.CodeSource;
 import java.util.Map;
 import java.util.WeakHashMap;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.logging.Logger;
 
@@ -50,8 +46,6 @@ public class AnonymousClassWarnings {
 
     private static final Logger LOGGER = Logger.getLogger(AnonymousClassWarnings.class.getName());
     private static final Map<Class<?>, Boolean> checked = new WeakHashMap<>();
-
-    private final static ExecutorService THREAD_POOL =  new AtmostOneThreadExecutor(new NamingThreadFactory(new DaemonThreadFactory(), AnonymousClassWarnings.class.getSimpleName()));
 
     /**
      * Checks a class which is being either serialized or deserialized.
@@ -69,7 +63,7 @@ public class AnonymousClassWarnings {
         } else {
             // May not call methods like Class#isAnonymousClass synchronously, since these can in turn trigger remote class loading.
             try {
-                THREAD_POOL.submit(() -> doCheck(clazz));
+                channel.executor.submit(() -> doCheck(clazz));
             } catch (RejectedExecutionException x) {
                 // never mind, we tried
             }
--- a/src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
+++ b/src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
@@ -53,8 +53,8 @@ public class AnonymousClassWarnings {
      */
     public static void check(@NonNull Class<?> clazz) {
         synchronized (checked) {
-            if (checked.containsKey(clazz)) {
-                return; // fast path
+            if (checked.put(clazz, true) != null) {
+                return;
             }
         }
         Channel channel = Channel.current();
@@ -85,18 +85,11 @@ public class AnonymousClassWarnings {
     private static void warn(@NonNull Class<?> c, String kind) {
         String name = c.getName();
         String codeSource = codeSource(c);
-        // Need to be very defensive about calling anything while holding this lock, lest we trigger class loading-related deadlocks.
-        boolean doWarn;
-        synchronized (checked) {
-            doWarn = checked.put(c, true) == null;
-        }
-        if (doWarn) {
-            if (codeSource == null) {
-                LOGGER.warning("Attempt to (de-)serialize " + kind + " class " + name + "; see: https://jenkins.io/redirect/serialization-of-anonymous-classes/");
-            } else {
-                // most easily tracked back to source using javap -classpath <location> -l '<name>'
-                LOGGER.warning("Attempt to (de-)serialize " + kind + " class " + name + " in " + codeSource + "; see: https://jenkins.io/redirect/serialization-of-anonymous-classes/");
-            }
+        if (codeSource == null) {
+            LOGGER.warning("Attempt to (de-)serialize " + kind + " class " + name + "; see: https://jenkins.io/redirect/serialization-of-anonymous-classes/");
+        } else {
+            // most easily tracked back to source using javap -classpath <location> -l '<name>'
+            LOGGER.warning("Attempt to (de-)serialize " + kind + " class " + name + " in " + codeSource + "; see: https://jenkins.io/redirect/serialization-of-anonymous-classes/");
         }
     }
 


Vlatombe:
This would essentially revert #523, which is fine for me since this was meant as a mitigation for [JENKINS-65873](https://issues.jenkins.io/browse/JENKINS-68954). But we know now that this is due to a JDK issue, it seems okay to me to just come back to the previous implementation.

Vlatombe (approved):

jglick:
OK, first I have seen that issue. There are a lot of comments there and I do not see a clear analysis. If this is a Java bug then why does only Jenkins seem to hit it? Is something creating an excessive number of threads—on the controller? on the agent? Why? There is mention both of `AnonymousClassWarnings` and of Pipeline log handling (`DelayBufferedOutputStream.reschedule`) in the issue, as well as mentions of the `git` plugin. #505 also mentions `JarCacheSupport.resolve`. Does not seem safe to merge until the actual problem is understood.

jglick:
Note that `DelayBufferedOutputStream` is using `jenkins.util.Timer` which is capped at 10 threads. It seems reasonable for `AnonymousClassWarnings` to also be restricted to one thread, because it is supposed to be pretty fast. If the agent is getting bogged down in calls to `doCheck` then something is going wrong but it is unclear from JENKINS-68954 what that would be—there are measurements of throughput but no thread dumps.

basil:
BTW my backport of the thread fix should finally be released in OpenJDK 11.0.16 (July 19 2022 GA).

bulanovk:
@basil is this fix will be available for JDK 17 ?

bulanovk:
What about merge and release 559 or 558 to fix hang up issue?

bulanovk:
Do we have a chance to get hang up issue fixed? We have 2 PR proposed to fix this behavior

jglick:
Regardless of the timing of the backport of a low-level JVM fix to Java 11, it remains unclear to me what the root cause of JENKINS-65873 was, and thus whether this is actually safe to merge and what else needs to be fixed. I believe @Vlatombe is not currently available to comment.

In the meantime, anyone should be able to test [this version of `agent.jar`](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/remoting/3037.v35344f192eb_b_/remoting-3037.v35344f192eb_b_.jar), specifically whether it indeed solves JENKINS-68954, whether it regresses JENKINS-65873 on pre-fix JREs, and whether it behaves well on post-fix JREs.

bulanovk:
JENKINS-68954 tested  with [agent ](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/remoting/3037.v35344f192eb_b_/remoting-3037.v35344f192eb_b_.jar), response time <= 600ms, no hang up was observed

jglick:
I finally got a chance to try https://issues.jenkins.io/browse/JENKINS-65873?focusedCommentId=413574&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-413574 and indeed it produced a truly ridiculous number of threads. My first thought was to just do nothing in the `channel != null` case, since this whole system is just there to warn developers about a bad idiom—dispensable; and in fact from the agent side it is pretty much useless (though without access to `JenkinsJVM` it is hard to determine this). Anyway, it turns out the code to only warn about a given class once was written stupidly (by me, in #259) and could short-circuit much earlier. With that patch, no glut of threads is created (even with an artificially slowed `doCheck`) and so there is no need to backport exotic JVM fixes.

jglick:
Specifically tested using a slight variant

```diff
diff --git pom.xml pom.xml
index ec9547f617..3a53a5f516 100644
--- pom.xml
+++ pom.xml
@@ -90,7 +90,7 @@ THE SOFTWARE.
     <changelog.url>https://www.jenkins.io/changelog</changelog.url>
 
     <!-- Bundled Remoting version -->
-    <remoting.version>3028.va_a_436db_35078</remoting.version>
+    <remoting.version>999999-SNAPSHOT</remoting.version>
     <!-- Minimum Remoting version, which is tested for API compatibility -->
     <remoting.minimum.supported.version>4.2.1</remoting.minimum.supported.version>
 
diff --git test/src/test/java/hudson/model/ProjectTest.java test/src/test/java/hudson/model/ProjectTest.java
index a5cc4bbdc0..d0c9ca54f9 100644
--- test/src/test/java/hudson/model/ProjectTest.java
+++ test/src/test/java/hudson/model/ProjectTest.java
@@ -68,6 +68,8 @@ import hudson.slaves.Cloud;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.EnvironmentVariablesNodeProperty;
 import hudson.slaves.NodeProvisioner;
+import hudson.slaves.RetentionStrategy;
+import hudson.slaves.SlaveComputer;
 import hudson.tasks.ArtifactArchiver;
 import hudson.tasks.BatchFile;
 import hudson.tasks.BuildTrigger;
@@ -102,6 +104,7 @@ import org.jvnet.hudson.reactor.ReactorException;
 import org.jvnet.hudson.test.FakeChangeLogSCM;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.SimpleCommandLauncher;
 import org.jvnet.hudson.test.TestBuilder;
 import org.jvnet.hudson.test.TestExtension;
 
@@ -154,6 +157,28 @@ public class ProjectTest {
         createAction = false;
     }
 
+    @Test
+    public void testRemoting() throws Exception {
+        FreeStyleProject p = j.createFreeStyleProject("project");
+        int sz = j.jenkins.getNodes().size();
+        SimpleCommandLauncher launcher = new SimpleCommandLauncher(
+                String.format("\"%s/bin/java\" -Djava.awt.headless=true -Xmx1g -Xms1g -jar \"%s\"",
+                        System.getProperty("java.home"),
+                        new File(j.jenkins.getJnlpJars("agent.jar").getURL().toURI()).getAbsolutePath()));
+        Slave agent = new DumbSlave("agent" + sz, "description", j.createTmpDir().getPath(), "1", Node.Mode.NORMAL, "", launcher, RetentionStrategy.NOOP, Collections.emptyList());
+        j.jenkins.addNode(agent);
+        j.waitOnline(agent);
+        SlaveComputer computer = (SlaveComputer) agent.toComputer();
+        System.err.println(computer.getLog());
+        p.setAssignedNode(agent);
+        p.getBuildersList().add(new Shell("bash -c 'for x in {000000..999999}; do echo -n $x; for y in {00..99}; do echo -n \" $y\"; done; echo; done'"));
+        try {
+            j.buildAndAssertSuccess(p);
+        } finally {
+            System.err.println(computer.getLog());
+        }
+    }
+
     @Test
     public void testGetEnvironment() throws Exception {
         FreeStyleProject p = j.createFreeStyleProject("project");
```

(The test times out rather than passing, but this is just on my laptop running on battery so I am not too concerned.)

basil:
Nice! I'll try to test this out tomorrow.

> If this is a Java bug then why does only Jenkins seem to hit it? Is something creating an excessive number of threads—on the controller? on the agent? Why?

I guess we now know the answers to these questions. Congratulations! You have figured out how to effectively stress test the Linux kernel's threads implementation. :grinning:

jglick:
Revised test `agent.jar`: https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/remoting/3038.v0937fb_2b_12f5/remoting-3038.v0937fb_2b_12f5.jar

basil (approved):
Reproduced the problem with #523 reverted and a 100-millisecond sleep in `doCheck` as described in [JENKINS-65873 (comment)](https://issues.jenkins.io/browse/JENKINS-65873?focusedCommentId=413574&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-413574) and a limit of 15,000 threads (but not 22,000 threads). After this PR I could no longer reproduce the issue with a limit of 15,000 threads and could even run the test successfully with a limit of 2,000 threads.

basil:
https://github.com/jenkinsci/remoting/releases/tag/3044.vb_940a_a_e4f72e

jglick:
⇒ https://github.com/jenkinsci/jenkins/pull/6936

src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
```diff
@@ -51,16 +47,14 @@ public class AnonymousClassWarnings {
...
 
-    private final static ExecutorService THREAD_POOL =  new AtmostOneThreadExecutor(new NamingThreadFactory(new DaemonThreadFactory(), AnonymousClassWarnings.class.getSimpleName()));
-
     /**
      * Checks a class which is being either serialized or deserialized.
      * A warning will only be printed once per class per JVM session.
      */
     public static void check(@NonNull Class<?> clazz) {
         synchronized (checked) {
-            if (checked.containsKey(clazz)) {
```
johnou, line 62: Can we keep the containsKey check and move it above the sync block (double checked locking) to potentially lower contention since this seems like a hot path?

johnou (commented):
Small nit.

src/main/java/org/jenkinsci/remoting/util/AnonymousClassWarnings.java
johnou: Ah nevermind the map in question is WeakHashMap and not thread safe.

PR #573, opened 2022-08-07
Restore `SSLContextRule` tests

basil:
When updating this repository to use Java 11 in #495, I took a shortcut and skipped the `SSLContextRule` tests rather than fixing them. This PR undoes that change, restoring the tests and getting them to work on Java 11.

They had started failing with "Certificate chain is not valid" errors. Looking into this further I found that the first argument to `X509v3CertificateBuilder`, which is supposed to be the signer/issuer, was actually a subject. This made no sense except for self-signed certificates, but apparently worked on older versions of Java. To fix this I changed the test logic to plumb through the root certificate in any certificate chains. With this object plumbed through, we could call its `getSubjectX500Principal()` method as recommended in the Bouncy Castle documentation to get the correct first argument to pass when building the X509 certificate. And with that the tests pass again on Java 11.


MarkEWaite (approved):
Thanks!

timja (approved):

PR #592, opened 2022-10-10
Closing WebSocket channel

twasyl:
Signed-off-by: Thierry Wasylczenko <twasyl@users.noreply.github.com>

Hello 👋 ,

Currently the channel used for the WebSocket connection was not terminated. We believe this could be the root cause. So this PR aims to close the channel to make sure all resources are freed and no side effects would arise.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -625,13 +625,23 @@ public class Engine extends Thread {
                     @Override
                     public void onClose(Session session, CloseReason closeReason) {
                         LOGGER.fine(() -> "onClose: " + closeReason);
-                        transport.terminate(new ChannelClosedException(ch.get(), null));
+                        final Channel channel = ch.get();
+                        final IOException ex = new ChannelClosedException(channel, null);
+                        if (channel != null) {
+                            channel.terminate(ex);
+                        }
+                        transport.terminate(ex);
                     }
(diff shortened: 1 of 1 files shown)


jglick (commented):
Seems reasonable. Has this been tested in context to make sure it at least does not break anything obvious in the happy path?

basil (commented):
https://github.com/jenkinsci/remoting/blob/09b895d8da14a10e33bae6b2fb59772ca9274d3f/src/main/java/hudson/remoting/Engine.java#L628 calls https://github.com/jenkinsci/remoting/blob/09b895d8da14a10e33bae6b2fb59772ca9274d3f/src/main/java/hudson/remoting/AbstractByteBufferCommandTransport.java#L314 which calls https://github.com/jenkinsci/remoting/blob/09b895d8da14a10e33bae6b2fb59772ca9274d3f/src/main/java/hudson/remoting/Channel.java#L620 which calls https://github.com/jenkinsci/remoting/blob/09b895d8da14a10e33bae6b2fb59772ca9274d3f/src/main/java/hudson/remoting/Channel.java#L1042-L1110 so I do not see a reason why we should call `Channel#terminate` a second time.

basil:
As in https://github.com/jenkinsci/remoting/pull/592#pullrequestreview-1150232524 I do not see a reason why this change should be made.

PR #593, opened 2022-10-13
Support junit 5; Migrate non-forked junit 3 test cases

Vlatombe:
Migrate existing junit 3 tests to junit 5 with standard `ParameterizedTest` facility

This uses `@MethodSource` providing the various ChannelRunner instances used for testing.

Migrated existing assert/assume calls impacted tests to the new jupiter api.

Overall this covers more combinations than the previous framework as some of them were actually not covered previously for some reason, even though they were extending `RmiTestBase`.

Keeping existing junit 3 test cases from forked packages as the proper approach here would rather be to pull new versions.


jglick:
https://github.com/jenkinsci/remoting/pull/593/checks?check_run_id=8873100454 flake or real?

src/main/java/hudson/remoting/Checksum.java
```diff
@@ -19,7 +20,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
-final class Checksum {
+final class Checksum implements Serializable {
```
jglick, line 23: ?

src/test/java/hudson/remoting/ChannelRunners.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.stream.Stream;
+
+public final class ChannelRunners {
+    public static final String PROVIDER_METHOD = "hudson.remoting.ChannelRunners#provider";
+
+    private ChannelRunners() {}
+
+    @SuppressWarnings("unused") // used by JUnit
+    static Stream<ChannelRunner> provider() {
```
jglick, line 11:
```suggestion
    public static Stream<ChannelRunner> provider() {
```
? Seems weird to be accessing a non`public` method reflectively.

src/main/java/hudson/remoting/Checksum.java
Vlatombe: Comes from one of the combination of `hudson.remoting.PrefetchingTest`, which I guess was not tested before.

Vlatombe:
> [#593 (checks)](https://github.com/jenkinsci/remoting/pull/593/checks?check_run_id=8873100454) flake or real?

~50% flake locally if I recall properly.

src/test/java/hudson/remoting/DeadRemoteOutputStreamTest.java
```diff
@@ -55,7 +64,7 @@ public Void call() throws Exception {
             StringWriter sw = new StringWriter();
             e.printStackTrace(new PrintWriter(sw));
             String whole = sw.toString();
-            assertTrue(whole, whole.contains(MESSAGE) && whole.contains("hudson.rem0ting.TestCallable"));
+            assertTrue(whole.contains(MESSAGE) && whole.contains("hudson.rem0ting.TestCallable"), whole);
```
jglick, line 67: Or switch to Hamcrest assertions.

src/test/java/hudson/remoting/PipeTest.java
```diff
@@ -68,15 +77,19 @@ public void testRemoteWrite() throws Exception {
...
+        channelRunner.withChannel(channel -> {
+            final Pipe p = Pipe.createRemoteToLocal();
+            final Future<Void> f = channel.callAsync(new InfiniteWriter(p));
+            try (InputStream in = p.getIn()) {
+                assertEquals(in.read(), 0);
+            }
 
-        final ExecutionException e = assertThrows(ExecutionException.class, f::get);
-        assertThat(e.getCause(), instanceOf(IOException.class));
+            final ExecutionException e = assertThrows(ExecutionException.class, () -> f.get(2, TimeUnit.SECONDS));
```
jglick, line 90: Correcting flake I guess?

src/test/java/hudson/remoting/PipeWriterTest.java
```diff
@@ -19,12 +25,11 @@ public class PipeWriterTest extends RmiTestBase implements Serializable, PipeWri
      * Proxy that can be used from the other side to verify the state.
      */
     PipeWriterTestChecker checker;
-
-    @Override
-    protected void setUp() throws Exception {
-        super.setUp();
-        // Checker operates using the user-space RMI
```
jglick, line 26: lost comment

jglick (approved):
OK I think. JUnit 5 is pretty opaque to me.

Does this patch make Surefire timeouts apply automatically?

src/test/java/hudson/remoting/PipeTest.java
Vlatombe: Yes

jglick:
Seems `ChannelTest` can still hang?

jglick:
Do you plan to look into https://github.com/jenkinsci/remoting/pull/593#issuecomment-1278857617, or shall I just merge this as is?

Vlatombe:
Added a timeout, I don't plan to investigate more time into the hanging test.

jglick:
https://github.com/jenkinsci/remoting/pull/593/commits/ca83285cc87ec1870a9bcb25232fd2a06cac2a50 did not work it seems:

```
00:06:57.224  [INFO] Running hudson.remoting.ChannelTest
00:17:11.572  Sending interrupt signal to process
```

jglick:
Maybe it worked so far as it went, but each test method gets run up to five times, and there are up to six parameter values, so spending a minute on each run in the worst case can be a lot.

Hmm, but it seems that by default timeouts are not preëmptive—if you exceed the timeout (but the method completes), the test fails, but that is no good for ensuring CI completes in time.

jglick:
Some really weird stuff like https://github.com/jenkinsci/remoting/blob/ca83285cc87ec1870a9bcb25232fd2a06cac2a50/src/test/java/org/jenkinsci/remoting/protocol/impl/ConnectionHeadersFilterLayerTest.java#L68-L71

PR #602, opened 2022-11-14
Remove unused logo

NotMyFault:
12 years later, and we can remove the unused hudson logo.


basil (approved):
This is the vector version of `src/main/resources/hudson/remoting/jnlp/title.png`, which was only removed this past May in #532. But I agree that if the `.png` was removed that this should also be removed for consistency.

timja (approved):

PR #541, opened 2022-05-24
fix: add exponential backoff

PertsevRoman:
I faced with DNS/timeout issues during agent initialization
WebSocket mode
```
io.jenkins.remoting.shaded.javax.websocket.DeploymentException: Connection failed.
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.container.jdk.client.JdkClientContainer$1.call(JdkClientContainer.java:187)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.container.jdk.client.JdkClientContainer$1.call(JdkClientContainer.java:107)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.container.jdk.client.JdkClientContainer.openClientSocket(JdkClientContainer.java:192)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.client.ClientManager$3$1.run(ClientManager.java:647)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.client.ClientManager$3.run(ClientManager.java:696)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.client.ClientManager$SameThreadExecutorService.execute(ClientManager.java:849)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.client.ClientManager.connectToServer(ClientManager.java:493)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.client.ClientManager.connectToServer(ClientManager.java:337)
	at hudson.remoting.Engine.runWebSocket(Engine.java:656)
	at hudson.remoting.Engine.run(Engine.java:495)
Caused by: java.nio.channels.UnresolvedAddressException
	at sun.nio.ch.Net.checkAddress(Net.java:104)
	at sun.nio.ch.UnixAsynchronousSocketChannelImpl.implConnect(UnixAsynchronousSocketChannelImpl.java:302)
	at sun.nio.ch.AsynchronousSocketChannelImpl.connect(AsynchronousSocketChannelImpl.java:210)
	at io.jenkins.remoting.shaded.org.glassfish.tyrus.container.jdk.client.TransportFilter.handleConnect(TransportFilter.java:184)
[...]

PR #608, opened 2022-12-06
[JENKINS-72163] Add tests for agent endpoint resolution retries

Vlatombe:
Follow-up https://github.com/jenkinsci/remoting/pull/449

When using `hudson.remoting.Engine.nonFatalJnlpAgentEndpointResolutionExceptions`, the failure to resolve JNLP endpoint is considered non-fatal, so there should be retries handled by remoting.

This can help with cases where a flaky network causes initial JNLP endpoint lookup to fail.


sboardwell:
Hi @Vlatombe, I have two additions if possible.

- add an option to avoid an infinite loop by setting the value of, e.g:
  - `nonFatalJnlpAgentEndpointResolutionExceptionsMaxRetries` max number of retries before failing
  - or `nonFatalJnlpAgentEndpointResolutionExceptionsMaxSecondsElapsed` for the max time elapsed before failing.
  - it might also be useful to add an 'x' second wait in between retries (possibly making it configurable?).

- secondly, would it be possible to make these properties configurable via environment variables?
  - currently you need to set `_JAVA_OPTIONS` in a pod template
  - being able to set environment variables seems a cleaner solution

basil (commented):
The primary intended consumer of this existing property is Swarm, which wraps Remoting around a layer that does its own retries. So this is unnecessary for Swarm. Was there another consumer you had in mind that would benefit from this?

Vlatombe:
@sboardwell has a customer case where it would be beneficial for the initial jnlp endpoint lookup to apply retries.

#541 also seems to stem from a similar use case.

In any case, I think this needs further formalization of the expected behaviour as the various states (lookup jnlp endpoint, connecting to tcp port, connecting through websockets) all have various different retry behaviours.

basil:
Yeah, I added this property to Remoting more or less as a hack to allow Swarm (with Swarm's higher-layer retrying) to work with Remoting. But I think ideally Remoting would natively support retries and exponential backoff which would allow Swarm to remove its own retrying functionality and delegating to Remoting. Happy to work toward that goal one step at a time.

sboardwell:
If we are going to implement this, would it also be possible to give a more informative error message if it does fail?

For example, the following seems to suggest the (HTTP) endpoint is timing when, in actual fact, it was the connection to the underlying TCP port 5000x

```log
Nov 29, 2022 1:49:37 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Locating server among [http://ci-server.ci-namespace.svc.cluster.local/ci-server/]
Nov 29, 2022 1:50:07 PM hudson.remoting.jnlp.Main$CuiListener error
SEVERE: Failed to connect to http://ci-server.ci-namespace.svc.cluster.local/ci-server/tcpSlaveAgentListener/: connect timed out
java.io.IOException: Failed to connect to http://ci-server.ci-namespace.svc.cluster.local/ci-server/tcpSlaveAgentListener/: connect timed out
at org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver.resolve(JnlpAgentEndpointResolver.java:211)
at hudson.remoting.Engine.innerRun(Engine.java:744)
at hudson.remoting.Engine.run(Engine.java:543)
Caused by: java.net.SocketTimeoutException: connect timed out
at java.base/java.net.PlainSocketImpl.socketConnect(Native Method)
```

sboardwell:
@Vlatombe - can I have a crack at an initial non-breaking stop-gap solution to this.

We still have this problem and it looks to be a bigger discussion to solve "properly". My solution was mentioned in https://github.com/jenkinsci/remoting/pull/608#issuecomment-1346437026 (the first one). This would not change current behaviour but allow people to set the retries/interval value themselves if they so wish.

basil (approved):
New tests fail without #675 and pass with it.

PR #615, opened 2023-01-04
[JENKINS-59910] Fix linked list bugs in `ProtocolStack`

basil:
Speculative fix for [JENKINS-59910](https://issues.jenkins.io/browse/JENKINS-59910): without a way to reproduce the problem locally, I cannot do any analysis. But while reading the code I found the locking quite messy, so I think tightening it up cannot hurt. I plan to ask affected users to try this and report back if it helps (or not). `mvn clean verify -Dtest=hudson.model.SlaveTest,hudson.slaves.JNLPLauncherRealTest,jenkins.AgentProtocolTest,jenkins.agents.WebSocketAgentsTest,jenkins.model.JenkinsTest,jenkins.security.Security218Test,jenkins.slaves.OldRemotingAgentTest,jenkins.slaves.RemotingVersionInfoTest,jenkins.slaves.restarter.JnlpSlaveRestarterInstallerTest,jenkins.slaves.UnsupportedRemotingAgentEscapeHatchTest,jenkins.slaves.UnsupportedRemotingAgentTest` passed in core with this PR.

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
+++ b/src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
@@ -606,6 +606,7 @@ public class ProtocolStack<T> implements Closeable, ByteBufferPool {
         /**
          * Flag to track this {@link ProtocolLayer} as removed from the stack.
          */
+        @GuardedBy("ProtocolStack.stackLock")
         private boolean removed;
 
         /**
@@ -760,7 +761,12 @@ public class ProtocolStack<T> implements Closeable, ByteBufferPool {
          * Requests removal of this {@link ProtocolLayer} from the {@link ProtocolStack}
          */
         public void remove() {
-            removed = true;
+            stackLock.writeLock().lock();
+            try {
+                removed = true;
+            } finally {
+                stackLock.writeLock().unlock();
+            }
         }
 
         /**
@@ -838,6 +844,21 @@ public class ProtocolStack<T> implements Closeable, ByteBufferPool {
         }
 
         /**
+         * Gets the {@link Ptr} for the next layer down the stack, skipping any removed {@link Ptr}
+         * instances and leaving the update to a later call.
+         *
+         * @return the {@link Ptr} for the next layer down the stack
+         */
+        @Nullable
+        private Ptr getNextSendImpl() {
+            Ptr nextSend = this.nextSend;
+            while (nextSend != null && nextSend.removed && nextSend.nextSend != null) {
+                nextSend = nextSend.nextSend;
+            }
+            return nextSend;
+        }
+
+        /**
          * Gets the {@link Ptr} for the next layer down the stack (processing the send half of any intermediary
          * {@link #removed} flags if possible, if not possible it skips the removed {@link Ptr} instances anyway,
          * leaving the update to a later call)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
```diff
@@ -760,7 +760,12 @@ public ProtocolStack<?> stack() {
          * Requests removal of this {@link ProtocolLayer} from the {@link ProtocolStack}
          */
         public void remove() {
-            removed = true;
+            stackLock.writeLock().lock();
+            try {
+                removed = true;
```
jtnord, line 765: interestingly `removed` is not `volatile`, could that be another cause of the issue?  

If we need the lock would that imply there are multiple threads at play and that this does also need to become volatile?

jtnord (approved):
looks reasonable to me.

src/main/java/org/jenkinsci/remoting/protocol/ProtocolStack.java
basil: I do not see a need for a memory barrier here but it could not hurt.

basil (commented):
There are two changes here: the first is mutual exclusion during removal, and the second is re-checking state.

Removal does not occur right away but rather a node is first marked for removal and then removals are processed in bulk. The problematic `NullPointerException` was coming from an `AckFilterLayer` node with a null `nextSend` pointer, which is an illegal state. Running through the scenario locally I found that both `AgentProtocolClientFilterLayer` and `AckFilterLayer` had been marked for deletion but not yet processed. But the important thing is that there is no way for the `nextSend` pointer on a filter node to be null other than a deletion gone wrong: unlinking a node is the only time in which we set its `nextSend` pointer to null. The deletion code looked right overall in the single-threaded case, so I think it is likely that something is going wrong when marking things for deletion or processing deletion requests in the concurrent case. Since the critical sections are so short, it seemed desirable to apply mutual exclusion to all sensitive operations in the removal workflow: marking things for deletion and actually deleting them. That way they cannot interfere with each other. So I added a write lock to the method that marks a node for deletion. This should also imply a memory fence.

The code to process pending removals relies on a local variable that was computed prior to acquiring the write lock. This clearly has the potential for error: as this [example](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/ReentrantReadWriteLock.html) shows, when we drop the read lock before acquiring the write lock then state may have changed between the time the local variable was computed with the read lock and the time we enter the write lock's critical section. So the local variable needs to be recomputed with the write lock. This is already done elsewhere in this class where a read lock is upgraded to a write lock, so it is just an oversight that it was not also done here. The `tryLock` usage now matches [this example](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/Lock.html#tryLock()): rather doing some work with a read lock, dropping it, and acquiring a write lock, we go for the write lock right away and if we get it then we do all of our work with it. If we don't get it then we do only the safe portion of the work with the read lock.

basil:
Sadly the problem still occurred when running this change in a loop in jenkinsci/jenkins#7567, so while this PR may fix _a_ problem, it doesn't fix _the_ problem we are interested in. However, it is promising that we can occasionally hit _the_ problem in automated test loops. There must still be something wrong with node deletion, but I just can't see it yet.

basil:
After reviewing Cormen, Leiserson, Rivest, and Stein's _Introduction to Algorithms_ (3rd edition), Chapter 10 (Elementary Data Structures), Section 2 (Linked Lists), "Deleting from a linked list", I see another problem: when unlinking the node, we set `x.prev.next` to `x.next`, but we fail to set `x.next.prev` to `x.prev` as the algorithm requires.

basil:
The original code processed deletions in two phases: first, unlinking one half of the doubly linked list (in one critical section), then unlinking the other half (in a different critical section). In practice the timing usually works out such that we process the previous pointer and next pointer of one node prior to attempting to unlink another node, but not always, and that is where disaster can happen. Because if a node is only partially unlinked and we try to unlink another node, undefined behavior can result.

I have no idea why this two-phase approach was used, but the whole separation seems pointless to me: the objective is to unlink the node, so I see no reason to do it in two stages. Better to get it done with in one fell swoop according to the classical linked list deletion algorithm. The bottom line is that there is only a single completion flag (`Ptr#remove`), so there is no need for a two-stage removal. If you look at the callers of `FilterLayer#onSendRemoved` and `FilterLayer#onRecvRemoved`, you can see that their callers are `getNextRecv` and `getNextSend` which are keyed off the same completion flag (`Ptr#remove`). So this is all just unnecessary complexity (a common theme when reading this class).

---

As an aside, the naming within this class is extremely confusing. Reviewers should ignore the names in the code and refer to the following names which actually make sense:

- `Ptr` is not a pointer but a linked list node; think of it as `Node`.
- `Ptr#removed` is not an indication of removal but rather a completion flag (i.e., removal request); think of it as `Node#completed`.
- `Ptr#nextRecv` is a next pointer in a doubly linked list; think of it as `Node#next`.
- `Ptr#nextSend` is a previous pointer in a doubly linked list; think of it as `Node#prev`.

I thought about renaming all of these, but I did not want to clutter up the diff with subjective naming changes.

basil:
Even with the latest changes I still got the same `UnsupportedOperationException: Network layer is not supposed to call isSendOpen` after 57 repetitions of `jenkins.slaves.restarter.JnlpSlaveRestarterInstallerTest#tcpReconnection`. I lack the interest in working on this further. I would recommend that the next person to work on this start running `jenkins.slaves.restarter.JnlpSlaveRestarterInstallerTest#tcpReconnection` in a loop and adding more logging to figure out what is going on.

PR #616, opened 2023-01-05
Document protocol stack

basil:


NotMyFault (commented):
It would be nice to generate a vector graphic over a png.

timja:
> It would be nice to generate a vector graphic over a png.

Mermaid is supported natively in GitHub if that's of interest: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams

basil:
I see no reason to do that.

jtnord (approved):

jtnord:
Changing to something else could be done in a future enhancement if some one wishes to spend time on that.

docs/protocol-stack.dot
```diff
@@ -0,0 +1,24 @@
+digraph g {
+	rankdir=LR;
+	node [shape=record];
+	null1 [label="null"];
+	network [label="<f0> nextRecv|<f1> BIONetworkLayer|<f2> nextSend"];
```
jtnord, line 5: NIT: generally Jenkins should be using [`NIONetworkLayer`](https://github.com/jenkinsci/remoting/blob/master/src/main/java/org/jenkinsci/remoting/protocol/impl/NIONetworkLayer.java) as the default so it would have been nice to start from there.  (its just an example to demonstrate but would be nice if it matched what was mostly used)

basil: I think Oleg mentioned the same thing to me years ago, but as far as I can tell open source Jenkins is using `BIONetworkLayer` as the default everywhere I've looked. Not familiar enough with proprietary usage to comment about other consumers, but the use of `BIONetworkLayer` in open source Jenkins seems to go back many years.

jtnord: https://github.com/jenkinsci/jenkins/blob/fbc5133ec3d6446460781fdea4d72ec70371a0e1/core/src/main/java/jenkins/slaves/JnlpSlaveAgentProtocol4.java#L132C22-L133

https://github.com/jenkinsci/remoting/blob/0267f59585dcc1982a994c9082755d9e92dbd7bf/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java#L107-L109

and eventually https://github.com/jenkinsci/remoting/blob/0267f59585dcc1982a994c9082755d9e92dbd7bf/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java#L193-L203

where preferNio is `true` so it should be NIO...

basil: I had a conversation with Oleg about this back in 2018 where I observed:

```
JnlpProtocol4Handler(JnlpProtocolHandler<STATE>).<init>(JnlpClientDatabase, boolean) line: 63
JnlpProtocol4Handler.<init>(JnlpClientDatabase, ExecutorService, IOHub, SSLContext, boolean, boolean) line: 107
JnlpProtocolHandlerFactory.handlers() line: 166
Engine.innerRun(IOHub, SSLContext, ExecutorService) line: 485 Engine.run() line: 469
```

The interesting part of this stack is the `innerRun` method, which is where we set `preferNio` to false:

```java
    private void innerRun(IOHub hub, SSLContext context,
ExecutorService service) {
        // Create the protocols that will be attempted to connect to the master.
        List<JnlpProtocolHandler> protocols = new JnlpProtocolHandlerFactory(service)
                .withIOHub(hub)
                .withSSLContext(context)
                .withPreferNonBlockingIO(false) // we only have one connection, prefer blocking I/O
                .handlers();
```

As you can see, here we explicitly set `preferNio` to false when initializing `JnlpProtocol4Handler`.

Anyway the point of this PR was not to dive deep into this long-standing behavior but just to document the basic protocol stack.

PR #628, opened 2023-02-24
Expand non-fatal JNLP agent endpoint resolution to do retries

sboardwell:
### What?

This PR adds two new system properties which can be used to further configure the retry attempts when using the `nonFatalJnlpAgentEndpointResolutionExceptions` property.

- `nonFatalJnlpAgentEndpointResolutionExceptionsMaxRetries`
    the number of retries before failing (defaults to 0 for backwards compatibility with the swarm plugin)
- `nonFatalJnlpAgentEndpointResolutionExceptionsIntervalInMillis`
    the interval in milliseconds before attempt to reconnect (defaults to 5000)

### Why?

The `nonFatalJnlpAgentEndpointResolutionExceptions` is currently used in the swarm plugin. The swarm plugin manages its own retry strategy.

In some envrionments with a flaky network, the JNLP endpoint resolution might simple fail due to a network issue. With standard incoming agents, the agent connection would fail at the first attempt and the build would fail.

These additional properties allow the configuration of retries in order to handle a flaky network.

### References

Follow-up #449, and initial fix before #608 hits.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -113,6 +113,12 @@ public class Engine extends Thread {
      */
     public static final String WEBSOCKET_COOKIE_HEADER = "Connection-Cookie";
 
+    static boolean nonFatalJnlpAgentResolutionExceptions = Boolean.getBoolean(Engine.class.getName() + ".nonFatalJnlpAgentEndpointResolutionExceptions");
+
+    static int nonFatalJnlpAgentResolutionExceptionsMaxRetries = Integer.getInteger(Engine.class.getName() + ".nonFatalJnlpAgentResolutionExceptionsMaxRetries", 2);
+
+    static int nonFatalJnlpAgentResolutionExceptionsIntervalInMillis = Integer.getInteger(Engine.class.getName() + ".nonFatalJnlpAgentResolutionExceptionsIntervalInMillis", 5000);
+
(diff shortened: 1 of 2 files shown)


sboardwell:
@basil / @Vlatombe - can you have a look at this please? This would be a non-invasive solution until a fully fledged solution is discussed and finally implemented in #608.

sboardwell:
@basil - one thing here would be that the default retry would be 2 (so 3 attempts in all) with an interval of 5 seconds in between each. Is that okay for your implementation, or would you like a `-1` value for infinite retries?

basil:
So you are proposing a change in behavior and asking me if it would affect an existing consumer? The source code for that existing consumer is available at https://github.com/jenkinsci/swarm-plugin so it should be possible to judge for oneself.

Whatever your use case may be, it is not described clearly in this PR. That would be enough to preclude the review of this PR even if this PR was not a self-described temporary workaround.

sboardwell:
I fully deserved that 😅.

Thank you @basil - I will make the the appropriate changes and come back to you.

sboardwell:
Hi @basil, is this okay?

basil:
Not sure we want to commit to this API.

sboardwell:
This would just be a stop gap until a long-term solution is found. It is non-breaking so I would hope that it would be okay.

PR #662, opened 2023-08-20
JENKINS-71856 - Replace TeeOutputStream with Apache commons implementation

solvingj:
Per the [JIRA issue](https://issues.jenkins.io/browse/JENKINS-71856) , this internal implementation of TeeOutputStream seems to have an obvious issue of double writing to the second `OutputStream`. It's apparently gone un-fixed because this implementation has only been used in one place which is likely to be rarely seen by human eyes. Other uses of `TeeInputStream` around the Jenkins codebase use an alternate (and apparently more correct) implementation of `TeeInputStream` from Apache commons which does not suffer from the double write.

This PR replaces the only internal usage of the internal TeeOutputStream with the Apache Commons implementation, and marks the internal class as deprecated with a link.

Of note, this change will likely fix the issue reported in https://issues.jenkins.io/browse/JENKINS-38520 which was closed as could not reproduce. I did not confirm it because I never experienced that issue personally and the reproduction and confirmation of the fix would be very time consuming. Also, it seems extremely likely to be connected on the surface.

### Testing done
The following code demonstrates the bug and simple fix by switching imports.  I ran the groovy code locally to confirm the fix, and also @MarkEWaite stated in the JIRA he confirmed the same behavior in Groovy console on latest Jenkins.

I am not planning to add anything the test suite here because there were no existing tests for `TeeOutputStream` and the test I would add would be a pure sanity test of a class from a third-party library (and apache commons at that) which isn't appropriate for multiple reasons.

```groovy
import hudson.remoting.TeeOutputStream  
//import org.apache.commons.io.output.TeeOutputStream 

File file = new File("test.txt")
File file2 = new File("test2.txt")
OutputStream os = new FileOutputStream(file)
OutputStream os2 = new FileOutputStream(file2)
TeeOutputStream tos = new TeeOutputStream(os, os2)
PrintWriter writer = tos.newPrintWriter()
writer.write("testing ")
writer.flush()
assert file.getText() == "testing "
assert file2.getText() == "testing testing "
```

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/remoting/TeeOutputStream.java
+++ b/src/main/java/hudson/remoting/TeeOutputStream.java
@@ -24,7 +24,6 @@ import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 
-
 /**
  * @deprecated Use instead {@link org.apache.commons.io.output.TeeOutputStream}
  * 
--- a/src/main/java/hudson/remoting/TeeOutputStream.java
+++ b/src/main/java/hudson/remoting/TeeOutputStream.java
@@ -24,13 +24,17 @@ import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 
+
 /**
+ * @deprecated Use instead {@link org.apache.commons.io.output.TeeOutputStream}
+ * 
  * Classic splitter of OutputStream. Named after the unix 'tee'
  * command. It allows a stream to be branched off so there
  * are now two streams.
  *
  * @version $Id: TeeOutputStream.java 610010 2008-01-08 14:50:59Z niallp $
  */
+@Deprecated
 @Restricted(NoExternalUse.class)
 public class TeeOutputStream extends FilterOutputStream {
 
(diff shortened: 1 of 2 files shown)


MarkEWaite:
Since the prior commons.io dependency was only a test dependency, I don't think that we can replace TeeOutputStream with a call to Apache commons.io.

The TeeOutputStream source code in remoting looks like a 2008 copy of the Apache commons.io code.  I think we'll be better served by applying the changes from most recent commons.io TeeOutputStream rather than adding commons.io as a compile time dependency of the remoting library.

solvingj:
I can appreciate wanting to keep maven dependencies low, and respect the decision to keep a vendored source instead of a maven dependency for a single file.

Upon further review, I found that there are 7 total files in the repo which are all old copies of sources from apache commons.  `TeeOutputStream` + 3 other files in the source + 3 corresponding unit tests:
https://github.com/search?q=repo%3Ajenkinsci%2Fremoting+%28ASF%29&type=code

There's also `Base64.java` but that's already deprecated and converted to a wrapper around `java.utils.Base64`.

I don't know what the threshold/policy is for adding a maven dependency instead of just copying and pasting sources from other libraries.  I don't know who decides that threshold.  I also don't know if there is some special reason this library (`remoting`), is trying harder-than-normal to avoid runtime maven dependencies, or to avoid deprecations and eventually deleting old dead sources like this

I do know that vendoring sources like this has disadvantages.  This bug is an example. There are many others.

With my limited knowledge, my recommendation to the project is to add the maven dependency for apache commons, deprecate the remaining vendored sources,

PR #666, opened 2023-08-22
Upgrade Commons Net from 3.6 to 3.9.0

basil:
In #312 with had inlined a copy of `SubnetUtils` from Commons Net 3.6 and proceeded to never upgrade it. This PR upgrades that class and its test to the latest code from version 3.9.0

### Testing done

Ran the unit test locally.


MarkEWaite (approved):
Thanks!

PR #677, opened 2023-10-13
Unify two entrypoints

basil:
### Problem

While reading this code recently, I found it quite confusing that there were two entrypoints: `java -jar remoting.jar` and `java -cp remoting.jar hudson.remoting.jnlp.Main`. The entrypoint code is a bit messy on both the interface level and the implementation level. There are two unique entrypoints, with quite similar APIs that overlap much of the time but also differ in some minor ways. This is quite confusing; for example, why should the argument be called `-noreconnect` in one entrypoint and `-noReconnect` in another, or why should the secret be specified via `-secret` in one entrypoint and with a positional argument in the other? Similarly, reading the implementation is quite confusing, as in some cases (but not all) the first implementation vectors into the second. So there is not only duplicate code, but also the logic is confusing to follow because sometimes only one implementation is being used, but sometimes two implementations are being used.

### Solution

Without breaking compatibility, this PR unifies the two entrypoints into a single implementation, with `Main` now delegating to `Launcher`, which now supports both its original semantics and that of `Main`. While this entails quite a few aliases, the idea is that once all usages are going through the `Launcher` entrypoint, we can also take the opportunity to standardize on the desired semantics (e.g., `-noReconnect` instead of `-noreconnect`, `-credentials` instead of `-jnlpCredentials`, etc). The idea is that once this PR is widely adopted, I will go through and update usages to standardize on a common set of calling conventions; for example, named arguments rather than positional arguments, camel case whenever possible, etc. More than just cleaning up the API, this also cleans up the implementations by merging them into one class that is easier to read and understand because the execution is all in one file and can be followed linearly rather than trying to understand what happens when one main method vectors into the other.

### Testing done

Apart from `mvn clean verify`, I've done some light testing against my local controller with both `java -jar remoting.jar` and `java -cp remoting.jar hudson.remoting.jnlp.Main`. I also tested Swarm with these changes and tested that the `-noCertificateCheck` option could be used to connect to a TLS-based Jenkins controller with a self-signed certificate on my local machine. I also ran the core test suite and PCT with these changes.


src/main/java/hudson/remoting/Engine.java
```diff
@@ -1034,16 +1036,18 @@ private static FileInputStream getFileInputStream(final File file) throws Privil
         });
     }
 
-    private SSLSocketFactory getSSLSocketFactory()
+    @CheckForNull
+    @Restricted(NoExternalUse.class)
+    static SSLSocketFactory getSSLSocketFactory(List<X509Certificate> x509Certificates)
```
basil, line 1041: Widening the visibility of this method and making static so that it can be reused in another place where this same logic is needed. Also added a `@CheckForNull` annotation while I was here, since this method can sometimes return null.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -127,17 +121,25 @@ public void setTextMode(boolean b) {
...
+    /**
+     * @deprecated use {@link #secret}, {@link #name}, {@link #urls}, {@link #webSocket}, {@link #tunnel},
+     * {@link #workDir}, {@link #internalDir}, and/or {@link #failIfWorkDirIsMissing} directly.
+     */
     @Option(name="-jnlpUrl",usage="instead of talking to the controller via stdin/stdout, " +
             "emulate a JNLP client by making a TCP connection to the controller. " +
-            "Connection parameters are obtained by parsing the JNLP file.")
+            "Connection parameters are obtained by parsing the JNLP file.", forbids = {"-direct", "-name", "-tunnel", "-url", "-webSocket"})
+    @Deprecated
     public URL agentJnlpURL = null;
```
basil, line 132: The most complicated mode of this plugin is when `-jnlpUrl` is passed in, which is not done by the Docker inbound agent image used for the Kubernetes plugin but which is recommended when creating an inbound agent in the UI. In this mode, you have to pass some arguments in the command-line invocation, but some other arguments can be provided by the server from the pseudo-JNLP endpoint. This is quite complex and it seems simpler to prefer that all clients simply invoke the tool with the parameters they want rather than relying on the pseudo-JNLP endpoint as a primitive bootstrapping mechanism.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -127,17 +121,25 @@ public void setTextMode(boolean b) {
...
 
-    @Option(name="-jnlpCredentials",metaVar="USER:PASSWORD",usage="HTTP BASIC AUTH header to pass in for making HTTP requests.")
+    @Option(name="-credentials",metaVar="USER:PASSWORD",aliases="-jnlpCredentials",usage="HTTP BASIC AUTH header to pass in for making HTTP requests.")
```
basil, line 134: `-credentials` in the other entrypoint. Since JNLP is an outdated concept, I think it's better to prefer `-credentials` as the main name for this argument.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -163,8 +165,11 @@ public void addClasspath(String pathList) throws Exception {
...
     public File tcpPortFile=null;
 
-
-    @Option(name="-auth",metaVar="user:pass",usage="If your Jenkins is security-enabled, specify a valid user name and password.")
+    /**
+     * @deprecated use {@link #agentJnlpCredentials} or {@link #proxyCredentials}
+     */
+    @Deprecated
+    @Option(name="-auth",metaVar="user:pass",usage="(deprecated) unused; use -credentials or -proxyCredentials")
     public String auth = null;
```
basil, line 173: The implementation of this was dead code, which is now deleted. The only ways of passing credentials that actually did anything were `-credentials` and/or `-proxyCredentials`.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -211,20 +221,16 @@ public void setConnectTo(String target) {
...
+    @Deprecated
     public void setNoCertificateCheck(boolean ignored) throws NoSuchAlgorithmException, KeyManagementException {
         System.out.println("Skipping HTTPS certificate checks altogether. Note that this is not secure at all.");
 
         this.noCertificateCheck = true;
-        SSLContext context = SSLContext.getInstance("TLS");
-        context.init(null, new TrustManager[]{new NoCheckTrustManager()}, new java.security.SecureRandom());
-        HttpsURLConnection.setDefaultSSLSocketFactory(context.getSocketFactory());
-        // bypass host name check, too.
-        HttpsURLConnection.setDefaultHostnameVerifier(new NoCheckHostnameVerifier());
```
basil, line 224: This entrypoint set the default SSL socket factory and the default hostname verifier JVM-wide, not just for the HTTP connections made by Remoting. The other entrypoint only set these values for the HTTP connections made by Remoting. Whenever dealing with a matter of security where each entrypoint had different behavior, we always chose the more secure option. Thus, we prefer the behavior of the other entrypoint which does not change JVM-wide behavior.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -283,12 +356,20 @@ public void setNoCertificateCheck(boolean ignored) throws NoSuchAlgorithmExcepti
     @Option(name="-version",usage="Shows the version of the remoting jar and then exits")
     public boolean showVersion = false;
 
+    /**
+     * The original calling convention takes two positional arguments: secret key and agent name.
+     * @deprecated use {@link #secret} and {@link #name}
+     */
+    @Argument
+    @Deprecated
+    public List<String> args = new ArrayList<>();
```
basil, line 365: The other entrypoint took two positional arguments, one for the secret and one for the agent name. Confusingly enough, the secret can also be provided by the server if using `-jnlpUrl` and `-credentials`. In any case, we now have named arguments for these for consistency with everything else, so deprecating the positional arguments.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -283,12 +356,20 @@ public void setNoCertificateCheck(boolean ignored) throws NoSuchAlgorithmExcepti
...
 
-    public static void main(String... args) throws Exception {
+    public static void main(String... args) throws IOException, InterruptedException {
         Launcher launcher = new Launcher();
         CmdLineParser parser = new CmdLineParser(launcher);
         try {
             parser.parseArgument(args);
+            normalizeArguments(launcher);
```
basil, line 372: Invariant: after calling `CmdlineParser#parseArguments`, call `normalizeArguments` to convert the old-positional arguments (if any) to the new-style named arguments. This is done here and one other place.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -326,95 +407,38 @@ public void run() throws Exception {
...
 
-        if(auth!=null) {
-            final int idx = auth.indexOf(':');
-            if(idx<0)   throw new CmdLineException(null, "No ':' in the -auth option", null);
-            Authenticator.setDefault(new Authenticator() {
-                @Override public PasswordAuthentication getPasswordAuthentication() {
-                    return new PasswordAuthentication(auth.substring(0,idx), auth.substring(idx+1).toCharArray());
-                }
-            });
-        }
```
basil, line 337: The only use of `-auth` (and only in this entrypoint: changing the default authenticator. This was dead code, as the values of `-credentials` and/or `-proxyCredentials` are used to preemptively send an authentication header.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -326,95 +407,38 @@ public void run() throws Exception {
...
-                for (String c: candidateCertificates) {
-                    jnlpArgs.add("-cert");
-                    jnlpArgs.add(c);
-                }
-            }
-            if (noCertificateCheck) {
-                // Generally it is not required since the default settings have been changed anyway.
-                // But we set it up just in case there are overrides somewhere in the logic
-                jnlpArgs.add("-disableHttpsCertValidation");
-            }
```
basil, line 383: Fret not as this code has not been deleted but rather moved a few lines below.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -326,95 +407,38 @@ public void run() throws Exception {
...
+                bootstrapInboundAgent();
             }
+            runAsInboundAgent();
         } else
         if(tcpPortFile!=null) {
             runAsTcpServer();
         } else {
             runWithStdinStdout();
         }
-        System.exit(0);
```
basil, line 398: This is actually needed when running with stdin/stdout, but not in every other case. Actually, Swarm _cannot_ have the process exit since it needs to programmatically control the loop. So I have moved this `System.exit` to the stdin/stdout case, where it lives alongside its other `System.exit` siblings.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -456,25 +480,146 @@ private SSLSocketFactory getSSLSocketFactory()
...
-            SSLContext ctx = SSLContext.getInstance("TLS");
-            ctx.init(null, trustManagerFactory.getTrustManagers(), null);
-            // now we have our custom socket factory
-            sslSocketFactory = ctx.getSocketFactory();
-        }
-        return sslSocketFactory;
+        }
+    }
+
+    private void bootstrapInboundAgent() throws CmdLineException, IOException, InterruptedException {
```
basil, line 493: By far the most complicated part of this code, this at least is better than the status quo by containing the complicated logic in one method rather than encoding it in the very structure of the classes and entrypoints. Basically, this makes a request to the server's pseudo-JNLP endpoint and based on the result overrides some of the arguments. In the long term I think this mode should cease to exist, which is why I have deprecated it above. But this method should preserve compatibility. Another way this code is better than the status quo is that it reuses more code between this mode and the non `-jnlpUrl` mode.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -456,25 +480,146 @@ private SSLSocketFactory getSSLSocketFactory()
...
+                throw new CmdLineException(null, "Cannot provide secret via both named and positional arguments");
+            }
+            if (launcher.name == null) {
+                launcher.name = launcher.args.get(1);
+            } else {
+                throw new CmdLineException(null, "Cannot provide name via both named and positional arguments");
+            }
+            launcher.args.clear();
+        }
+    }
```
basil, line 590: This is here to retain backwards compatibility with the positional argument calling convention. It converts the positional arguments, if present, to their named argument replacements.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -456,25 +480,146 @@ private SSLSocketFactory getSSLSocketFactory()
...
+        if (launcher.urls.isEmpty() && launcher.directConnection == null) {
+            throw new CmdLineException(null, "At least one URL is required for inbound agents");
+        }
+        if (launcher.webSocket) {
+            assert !launcher.urls.isEmpty(); // depends = "-url"
+            if (launcher.urls.size() > 1) {
+                throw new CmdLineException(null, "Only a single URL is supported for WebSocket agents");
+            }
+        }
+    }
```
basil, line 609: Invariant: whenever connecting as an inbound agent, invoke this method to validate the arguments. This happens once in the normal flow and twice in the `-jnlpUrl` flow.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -456,25 +480,146 @@ private SSLSocketFactory getSSLSocketFactory()
...
+        engine.startEngine();
+        try {
+            engine.join();
+            LOGGER.fine("Engine has died");
+        } finally {
+            // if we are programmatically driven by other code, allow them to interrupt our blocking main thread to kill
+            // the on-going connection to Jenkins
+            engine.interrupt();
+        }
     }
```
basil, line 623: Not original code, just moved from the other entrypoint.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -491,19 +636,7 @@ public List<String> parseJnlpArguments() throws ParserConfigurationException, SA
...
-                    if  (agentJnlpCredentials != null) {
-                        String userPassword = agentJnlpCredentials;
-                        String encoding = Base64.getEncoder().encodeToString(userPassword.getBytes(StandardCharsets.UTF_8));
-                        http.setRequestProperty("Authorization", "Basic " + encoding);
-                    }
-                    if (System.getProperty("proxyCredentials", proxyCredentials) != null) {
-                        String encoding = Base64.getEncoder().encodeToString(System.getProperty("proxyCredentials", proxyCredentials).getBytes(StandardCharsets.UTF_8));
-                        http.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
-                    }
-                }
```
basil, line 506: Deduplicating this logic, which also lived in the method that is now being called. Also note that this was the real logic that handled authentication, not the dead code above that was used for the `-auth` argument.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -550,10 +683,6 @@ public List<String> parseJnlpArguments() throws ParserConfigurationException, SA
                 List<String> jnlpArgs = new ArrayList<>();
                 for( int i=0; i<argElements.getLength(); i++ )
                         jnlpArgs.add(argElements.item(i).getTextContent());
-                if (agentJnlpCredentials != null) {
-                    jnlpArgs.add("-credentials");
-                    jnlpArgs.add(agentJnlpCredentials);
-                }
```
basil, line 556: This code is actually being kept, just moved next to its siblings for legibility rather than in this random place.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -819,6 +949,105 @@ private static void closeWithLogOnly(Closeable stream, String name) {
...
+            System.exit(-1);
+        }
+
+        @Override
+        public void onDisconnect() {}
+
+        @Override
+        public void onReconnect() {}
+    }
+
```
basil, line 1050: Not original; moved from the other entrypoint.

src/main/java/hudson/remoting/Util.java
```diff
@@ -100,63 +91,6 @@ static String indent(String s) {
         return "    " + s.trim().replace("\n", "\n    ");
     }
 
-    /**
```
basil, line 103: Fret not that these public methods are being deleted, as this whole class is `@Restricted(NoExternalUse.class)`. Anyway, these were all duplicate code. Keeping the more widely used copy.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -499,8 +501,9 @@ else if(entry.endsWith("*"))
      * Credentials can be passed e.g. to support running Jenkins behind a (reverse) proxy requiring authorization
      * FIXME: similar to hudson.remoting.Util.openURLConnection which is still used in hudson.remoting.Launcher
      */
+    @Restricted(NoExternalUse.class)
     @SuppressFBWarnings(value = "URLCONNECTION_SSRF_FD", justification = "Used by the agent for retrieving connection info from the server.")
-    static URLConnection openURLConnection(URL url, String credentials, String proxyCredentials,
+    public static URLConnection openURLConnection(URL url, String credentials, String proxyCredentials,
```
basil, line 506: The more widely used copy of the utility code deleted from the `Util` class. Visibility is being increased so that this code can be reused.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -550,8 +553,6 @@ static URLConnection openURLConnection(URL url, String credentials, String proxy
 
             } else if (sslSocketFactory != null) {
                 httpsConnection.setSSLSocketFactory(sslSocketFactory);
-                //FIXME: Is it really required in this path? Seems like a bug
-                httpsConnection.setHostnameVerifier(new NoCheckHostnameVerifier());
```
basil, line 554: As before, whenever it comes to matters of security, we always err on the side of the more restrictive option when unifying the two implementations. Previously, this bug existed in the second entrypoint but not the first (i.e., the second entrypoint was more loose). So if invoking the first entrypoint (which itself invokes the second entrypoint), you got the more restrictive behavior, and if invoking the second entrypoint directly you got the loose behavior. Now that the entrypoints are being unified, I had to choose which behavior to follow. Rather than risk introducing a security issue, I choose the more restrictive behavior. By the way, this can be turned off with `-noCertificateCheck` (I tested this) and these semantics exactly match Swarm, where there have been no complaints, so I think this is a reasonable change in behavior in a very minor edge case.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -211,20 +221,16 @@ public void setConnectTo(String target) {
      *
      * @param ignored
      *      This is ignored.
+     * @deprecated {use {@link #noCertificateCheck}
```
jglick, line 224: typo I think

```suggestion
     * @deprecated use {@link #noCertificateCheck}
```

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -269,6 +275,73 @@ public void setNoCertificateCheck(boolean ignored) throws NoSuchAlgorithmExcepti
...
+            metaVar = "HOST:PORT",
+            aliases = "-directConnection",
+            depends = "-instanceIdentity",
+            forbids = {"-jnlpUrl", "-url", "-tunnel"},
+            usage = "Connect directly to this TCP agent port, skipping the HTTP(S) connection parameter download. For example, \"myjenkins:50000\".")
+    public String directConnection;
+
+    /**
+     * The controller's instance identity.
+     * @see <a href="https://wiki.jenkins.io/display/JENKINS/Instance+Identity">Instance Identity</a>
```
jglick, line 326: May as well switch to a non-deprecated page while we are touching this:
```suggestion
     * @see <a href="https://plugins.jenkins.io/instance-identity/">Instance Identity</a>
```

docs/inbound-agent.md
```diff
@@ -78,7 +78,7 @@ This mechanism requires a download of the `agent.jar`, as described for "Downloa
 
 Once all the prerequisite files and data have been obtained, the agent can be launched with a command like this
 ```
-java -cp agent.jar hudson.remoting.jnlp.Main \
+java -jar agent.jar \
   -workDir <work directory> \
   -direct <HOST:PORT> \
   -protocols JNLP4-connect \
```
jglick, line 84: I think https://github.com/jenkinsci/remoting/blob/2232d9d9beb25ac293aa6bb2eb67ab097c87d173/docs/inbound-agent.md#L86 should now read

```
-secret <secretString> -name <agentName>
```

or similar to avoid using the deprecated positional arguments?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -127,17 +121,25 @@ public void setTextMode(boolean b) {
...
-            "Connection parameters are obtained by parsing the JNLP file.")
+            "Connection parameters are obtained by parsing the JNLP file.", forbids = {"-direct", "-name", "-tunnel", "-url", "-webSocket"})
+    @Deprecated
     public URL agentJnlpURL = null;
 
-    @Option(name="-jnlpCredentials",metaVar="USER:PASSWORD",usage="HTTP BASIC AUTH header to pass in for making HTTP requests.")
+    @Option(name="-credentials",metaVar="USER:PASSWORD",aliases="-jnlpCredentials",usage="HTTP BASIC AUTH header to pass in for making HTTP requests.")
     public String agentJnlpCredentials = null;
 
     @Option(name="-secret", metaVar="HEX_SECRET", usage="Agent connection secret to use instead of -jnlpCredentials.")
```
jglick, line 137:
```suggestion
    @Option(name="-secret", metaVar="HEX_SECRET", usage="Agent connection secret to use instead of -credentials.")
```
since `-jnlpCredentials` is now deprecated; or simply
```suggestion
    @Option(name="-secret", metaVar="HEX_SECRET", usage="Agent connection secret.")
```
since using `-secret` is the normal usage and `-credentials` should perhaps just be deprecated (perhaps there are use cases in exotic reverse proxy configurations, but even then you probably _also_ need `-secret` for the TCP connection).

src/main/java/hudson/remoting/Launcher.java
jglick: > In the long term I think this mode should cease to exist

Agreed; see https://github.com/jenkinsci/docker-swarm-plugin/pull/125 https://github.com/KostyaSha/yet-another-docker-plugin/issues/288#issuecomment-687212127 etc.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -491,19 +636,7 @@ public List<String> parseJnlpArguments() throws ParserConfigurationException, SA
...
-                con = Util.openURLConnection(agentJnlpURL);
-                if (con instanceof HttpURLConnection) {
-                    HttpURLConnection http = (HttpURLConnection) con;
-                    if  (agentJnlpCredentials != null) {
-                        String userPassword = agentJnlpCredentials;
-                        String encoding = Base64.getEncoder().encodeToString(userPassword.getBytes(StandardCharsets.UTF_8));
-                        http.setRequestProperty("Authorization", "Basic " + encoding);
-                    }
-                    if (System.getProperty("proxyCredentials", proxyCredentials) != null) {
-                        String encoding = Base64.getEncoder().encodeToString(System.getProperty("proxyCredentials", proxyCredentials).getBytes(StandardCharsets.UTF_8));
```
jglick, line 503: Note loss of interpretation of system property. Fine I think; `Engine` still looks up the system property apparently.

jglick (approved):
Looks right. The more obscure options have no test coverage unfortunately.

docs/inbound-agent.md
basil: Fixed in commit 7875c0d0

src/main/java/hudson/remoting/Launcher.java
basil: Fixed in commit e4c97287.

basil:
I did some more testing with WebSocket, a self-signed certificate, and a local Squid forwarding proxy. All worked fine except for proxy authentication, but no matter which combination of entrypoints and arguments I tried, I couldn't get that to work in the original trunk code either before this PR. I did come up with a fix for proxy authentication, but I will leave that to a separate PR.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -326,95 +407,38 @@ public void run() throws Exception {
...
-                jnlpArgs.add("-disableHttpsCertValidation");
-            }
-            try {
-                hudson.remoting.jnlp.Main._main(jnlpArgs.toArray(new String[0]));
-            } catch (CmdLineException e) {
-                System.err.println("JNLP file "+ agentJnlpURL +" has invalid arguments: "+jnlpArgs);
-                System.err.println("Most likely a configuration error in the controller");
-                System.err.println(e.getMessage());
-                System.exit(1);
+        if (agentJnlpURL != null || !urls.isEmpty()) {
```
jglick, line 419: I think this also needs to cover `directConnection`, where neither `-jnlpUrl` nor `-url` is used yet we are running an inbound agent. https://github.com/jenkinsci/kubernetes-plugin/pull/1451#issuecomment-1775147649

basil: Indeed, somehow I missed testing this mode during all the other manual testing I did. Thanks for fixing this so quickly.

PR #637, opened 2023-05-08
Managing test flakes

jglick:
Some failures observed in CI, but mainly just running

```bash
mvn test -DforkCount=.45C
```

(about five concurrent tests on my machine) over and over. Surprisingly hard to get a clean run; #142 masks poor tests and we do not yet have the infrastructure to track flakes (maybe Launchable could).

PR #679, opened 2023-10-13
Need to use `@Ignore` on JUnit 4, `@Disabled` on 5

jglick:
Amending #637 after I noticed https://github.com/jenkinsci/remoting/pull/637#discussion_r1187796771 _still_ flaking in CI.

### DIFF
--- a/src/main/java/hudson/remoting/ObjectInputStreamEx.java
+++ b/src/main/java/hudson/remoting/ObjectInputStreamEx.java
@@ -41,6 +41,7 @@ public class ObjectInputStreamEx extends ObjectInputStream {
     /**
      * @deprecated as of 2.53
      */
+    @Deprecated
     public ObjectInputStreamEx(InputStream in, ClassLoader cl) throws IOException {
         this(in,cl,ClassFilter.DEFAULT);
     }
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/remoting/ObjectInputStreamEx.java
```diff
@@ -41,6 +41,7 @@ public class ObjectInputStreamEx extends ObjectInputStream {
     /**
      * @deprecated as of 2.53
      */
+    @Deprecated
```
jglick, line 44: Just wanted to be able to `mvn test` without getting distracted by compiler warnings.

src/test/java/hudson/remoting/RegExpBenchmark.java
```diff
@@ -32,9 +32,9 @@
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.regex.Pattern;
-import org.junit.jupiter.api.Disabled;
+import org.junit.Ignore;
 
-@Disabled("This is not a test just a benchmark and is here for ease of running")
+@Ignore("This is not a test just a benchmark and is here for ease of running")
```
jglick, line 37: Consistently fails for me locally; not sure how it even passes on CI. Something about JRE vs. JDK?

basil (approved):

PR #685, opened 2023-10-16
Send `Node-Name` HTTP header in `JnlpAgentEndpointResolver`

jglick:
https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/org/jenkinsci/remoting/engine/JnlpConnectionState.java#L65 is passed as an HTTP header in WebSocket mode https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/hudson/remoting/Engine.java#L555 and as a protocol header in TCP mode https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/hudson/remoting/Engine.java#L740 This patch just passes it also as an HTTP header in the handshake phase of TCP mode, when the agent checks `https://jenkins/tcpSlaveAgentListener/` to determine a host and port to connect to. Jenkins core will ignore the header; CloudBees CI can use it to adjust the TCP connection information according to the identity of the agent.


src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -107,22 +108,15 @@ public class JnlpAgentEndpointResolver extends JnlpEndpointResolver {
     private static String PROTOCOL_NAMES_TO_TRY =
             System.getProperty(JnlpAgentEndpointResolver.class.getName() + ".protocolNamesToTry");
 
-    public JnlpAgentEndpointResolver(String... jenkinsUrls) {
-        this.jenkinsUrls = new ArrayList<>(Arrays.asList(jenkinsUrls));
-    }
-
-    public JnlpAgentEndpointResolver(@NonNull List<String> jenkinsUrls) {
-        this(jenkinsUrls, null, null, null, null, false);
-    }
```
jglick, line 116: These overloads appear unused, so I did not bother retaining compatibility for them.

Vlatombe (approved):

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -529,6 +523,7 @@ static URLConnection openURLConnection(URL url, String credentials, String proxy
             String encoding = Base64.getEncoder().encodeToString(proxyCredentials.getBytes(StandardCharsets.UTF_8));
             con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
         }
+        con.setRequestProperty(JnlpConnectionState.CLIENT_NAME_KEY, agentName);
```
basil, line 526: There are duplicate copies of this code at

```
src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
526:            con.setRequestProperty("Authorization", "Basic " + encoding);
530:            con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);

src/main/java/hudson/remoting/Util.java
131:            con.setRequestProperty("Authorization", "Basic " + encoding);
135:            con.setRequestProperty("Proxy-Authorization", "Basic " + encoding);

src/main/java/hudson/remoting/Launcher.java
500:                        http.setRequestProperty("Authorization", "Basic " + encoding);
504:                        http.setRequestProperty("Proxy-Authorization", "Basic " + encoding);
```

raiseing the question: why change this copy but not the others?

jglick: https://github.com/jenkinsci/remoting/blob/5491eb1a523ec1d0fa98a2c2e763a38f5acd8820/src/main/java/hudson/remoting/Util.java#L103-L158 is used only from https://github.com/jenkinsci/remoting/blob/5491eb1a523ec1d0fa98a2c2e763a38f5acd8820/src/main/java/hudson/remoting/Launcher.java#L480-L507 and this is only used for the `*.jnlp` retrieval from the `java -jar` entrypoint in which the agent name is already encoded in the URL.

jglick: Indeed there are three copies of the code to define `Authorization` and `Proxy-Authorization`, and two of the code to use `Proxy` (some of this was noticed in #243), as well as some unused overloads and other dead code, and it would be nice to clean this sort of thing up. That does not have any particular bearing on this PR since the `Node-Name` header is only useful in the `/tcpSlaveAgentListener/` request made from `JnlpAgentEndpointResolver`.

basil: > That does not have any particular bearing on this PR

It does have particular bearing on this PR, since this PR is making the existing problem worse by increasing the divergence between the three copies, thereby increasing the maintenance burden on the Jenkins community.

jglick: Well at any rate #687 cleans up code without any behavioral change.

basil (changes requested):
Making it clear that I am blocking this PR until it does not increase the divergence between multiple copies of this code.

jglick:
This PR does not actually increase the divergence between copies, since it adds an independent line of code after the common portions; if #687 is merged then I would resolve the merge conflict which would make that more apparent.

basil:
> This PR does not actually increase the divergence between copies, since it adds an independent line of code after the common portions

I'm having a hard time following this point, so maybe you could clarify. Are you saying that even in the context of the more comprehensive refactoring done in #677 (which condenses all `openURLConnection` methods into one), you'd _still_ want to send this header only in the request to `/tcpSlaveAgentListener/` and not the request to `/computer/${NODE_NAME}/jenkins-agent.jnlp` because in the former case the server has no way of getting the agent name without the header but in the latter case the server can get the header from the URL?

jglick:
> Are you saying that…

Yes, exactly. `EncryptedSlaveAgentJnlpFile.slaveName` is already determined by Stapler ancestry. I guess it does no harm to pass a redundant header if the new caller of `JnlpAgentEndpointResolver.openURLConnection` in `Launcher` has access to `name` (or it is an optional parameter).

basil (approved):
OK, I could also see an argument for always passing the header for greater consistency among the client's requests. But I don't feel strongly either way. Approving this PR.

jglick:
I do not feel strongly either way either. Will wait for #677 to be merged and resolve conflicts with that.

basil (approved):

PR #686, opened 2023-10-16
`JNLP4-connect-proxy` connection protocol

jglick:
Introduces an agent protocol used by CloudBees CI to intercept TCP connection requests and route them specially to the controller, by determining the agent name before proceeding with the connection. Ideally this could simply be added as a header to be interpreted by `TcpSlaveAgentListener.ConnectionHandler` like in #685, but unfortunately the TCP meta-protocol only allows the protocol _name_ to be sent, and not any other metadata: https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/org/jenkinsci/remoting/protocol/impl/AgentProtocolClientFilterLayer.java#L56 The other obvious choice in the stack https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4Handler.java#L175-L182 would be the acknowledgement layer, but again this was built to only accept a single fixed string https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/org/jenkinsci/remoting/protocol/impl/AckFilterLayer.java#L94 and not offer any other extensibility while retaining compatibility: https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/org/jenkinsci/remoting/protocol/impl/AckFilterLayer.java#L142-L145 Routing at a lower layer would be much harder, so this was the most straightforward technique I could find.

An agent connecting to Jenkins core will simply skip this protocol since the server does not offer it: https://github.com/jenkinsci/remoting/blob/b329c48ba7fca39a3f59e593ba88490d1ea669df/src/main/java/hudson/remoting/Engine.java#L815-L818

If you wish to implement the server side of the protocol from inside Jenkins, it will look something like this:

```java
@Extension
public final class MyHandler extends AgentProtocol {
    @Override
    public String getName() {
        return JnlpProtocol4ProxyHandler.NAME;
    }
    @Override
    public String getDisplayName() {
        return "My handler";
    }
    @Override
    public void handle(Socket socket) throws IOException, InterruptedException {
        var agentIO = socket.getChannel();
        var br = new BufferedReader(Channels.newReader(agentIO, StandardCharsets.UTF_8));
        var headers = new HashMap<String, String>();
        while (true) {
            var line = br.readLine();
            if (line.isBlank()) {
                break;
            }
            var colon = line.indexOf(':');
            headers.put(line.substring(0, colon).trim(), line.substring(colon + 1).trim());
        }
        var nodeName = headers.get(JnlpConnectionState.CLIENT_NAME_KEY);
        // Do nothing special (handle with default protocol):
        var protocol = new DataInputStream(SocketChannelStream.in(socket)).readUTF();
        var delegate = ExtensionList.lookupSingleton(JnlpSlaveAgentProtocol4.class);
        if (!protocol.equals("Protocol:" + delegate.getName())) {
            throw new IOException("Unexpected protocol header: " + protocol);
        }
        delegate.handle(socket);
    }
}
```

### DIFF
--- a/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4ProxyHandler.java
+++ b/src/main/java/org/jenkinsci/remoting/engine/JnlpProtocol4ProxyHandler.java
@@ -1,0 +1,91 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2023 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


basil (commented):
Is this code exercised anywhere in (open-source) Jenkins? It seems like it would be desirable to have the basic example shown in the PR description checked in somewhere and running regularly to ensure that this code does not regress. Otherwise, that is putting a maintenance burden on the Jenkins community.

basil:
> I will check whether it is feasible

If it is feasible with a reasonable amount of effort, I think that should be done as part of the present effort (rather than deferred) in order to avoid introducing maintenance burden.

basil (approved):

PR #690, opened 2023-10-23
Support `-cert` and `-noCertificateCheck` in WebSocket mode

basil:
I couldn't see a good reason to support these in one transport but not the other, given that it is just a matter of plumbing through the relevant `SSLContext` to the HTTP layer.

### Testing done

Set up a WebSocket agent against my controller with a self-signed HTTPS certificate. Verified the connection failed before this PR and passed after this PR when adding `-noCertificateCheck`. I also tested that I could connect with HTTPS and `-noCertificateCheck` _without_ WebSocket. I did not test `-cert` because I don't actually have a valid certificate for my local testing environment. But this code is common code, so if it's plumbed through correctly in one case it should be plumbed through correctly in the other as well.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -46,6 +46,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.security.SecureRandom;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -1038,11 +1057,14 @@ private static FileInputStream getFileInputStream(final File file) throws Privil
...
+    static SSLContext getSSLContext(List<X509Certificate> x509Certificates, boolean noCertificateCheck)
             throws PrivilegedActionException, KeyStoreException, NoSuchProviderException, CertificateException,
             NoSuchAlgorithmException, IOException, KeyManagementException {
-        SSLSocketFactory sslSocketFactory = null;
-        if (x509Certificates != null && !x509Certificates.isEmpty()) {
+        SSLContext sslContext = null;
+        if (noCertificateCheck) {
+            sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(null, new TrustManager[]{new NoCheckTrustManager()}, new SecureRandom());
+        } else if (x509Certificates != null && !x509Certificates.isEmpty()) {
```
basil, line 1066: Moved from `JnlpAgentEndpointResolver#openURLConnection` to facilitate code reuse.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -297,8 +297,6 @@ public void setNoCertificateCheck(boolean ignored) throws NoSuchAlgorithmExcepti
                 "-tunnel",
                 "-credentials",
                 "-proxyCredentials",
-                "-cert",
-                "-noCertificateCheck",
```
basil, line 301: Removing these restrictions.

timja (approved):

PR #691, opened 2023-10-23
Proxy support for WebSocket

basil:
This component supports the `http.proxyHost` system property (or `proxy_host` environment variable) for configuring a proxy server, but it is not honored in WebSocket mode. By plumbing this through to Tyrus we can get Tyrus to respect the proxy settings the same way non-WebSocket mode does. Similarly we can plumb through the value of `-proxyCredentials` to Tyrus if the user needs to pass in a `Proxy-Authorization` header (like we do in non-WebSocket mode). We do this only for HTTP to match the non-WebSocket mode.

### Testing done

Set up a Squid proxy that required authentication and verified that I could connect through it using:

```
java -Dhttp.proxyHost=${HOSTNAME} -Dhttp.proxyPort=3128 -jar target/remoting-999999-SNAPSHOT.jar -url ${JENKINS_URL} -secret ${SECRET} -name test -webSocket -proxyCredentials ${USERNAME}:${PASSWORD}
```

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -37,6 +37,7 @@ import java.net.Socket;
 import java.net.URI;
 import java.net.URL;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.security.AccessController;
 import java.security.KeyManagementException;
@@ -52,6 +53,7 @@ import java.security.cert.X509Certificate;
 import java.security.interfaces.RSAPublicKey;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -682,7 +687,26 @@ public void closeRead() throws IOException {
...
                 hudsonUrl = candidateUrls.get(0);
                 String wsUrl = hudsonUrl.toString().replaceFirst("^http", "ws");
-                ContainerProvider.getWebSocketContainer().connectToServer(new AgentEndpoint(),
+                WebSocketContainer container = ContainerProvider.getWebSocketContainer();
+                if (container instanceof ClientManager) {
+                    ClientManager client = (ClientManager) container;
+
+                    String proxyHost = System.getProperty("http.proxyHost", System.getenv("proxy_host"));
+                    String proxyPort = System.getProperty("http.proxyPort");
+                    if (proxyHost != null && "http".equals(hudsonUrl.getProtocol()) && NoProxyEvaluator.shouldProxy(hudsonUrl.getHost())) {
```
basil, line 696: Compare with `JnlpAgentEndpointResolver#openURLConnection`

PR #693, opened 2023-10-24
Move initialization to common method for Swarm

basil:
Found a minor issue when trying to refactor Swarm to take advantage of these recent changes: Swarm directly calls `Launcher#parseJnlpArguments` to get the secret from the server (and can't easily be changed to stop doing this), and that currently means that the initialization logic from `Launcher#run` doesn't get run, so things like `-noCertificateCheck` are working only by accident in that their initialization is done lazily rather than eagerly. This PR fixes that problem and makes the code more easy to understand by putting all initialization logic in one place and ensuring that all public entrypoints invoke it exactly once, and eagerly at the beginning of program launch.

We also took the various initialization paradigms and made them consistent: eager rather than lazy, using the same set of fields everywhere, the same set of arguments everywhere in the same order, etc.

### Testing done

Ran both Remoting and Swarm tests with these changes.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -67,6 +67,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
@@ -186,6 +187,9 @@ public class Engine extends Thread {
     private String tunnel;
 
     private boolean disableHttpsCertValidation = false;
+
+    @CheckForNull
+    private HostnameVerifier hostnameVerifier;
 
     private boolean noReconnect = false;
 
@@ -424,6 +428,11 @@ public class Engine extends Thread {
      */
     public void setDisableHttpsCertValidation(boolean disableHttpsCertValidation) {
         this.disableHttpsCertValidation = disableHttpsCertValidation;
+        if (disableHttpsCertValidation) {
+            this.hostnameVerifier = new NoCheckHostnameVerifier();
+        } else {
+            this.hostnameVerifier = null;
+        }
     }
 
     /**
@@ -713,8 +722,8 @@ public class Engine extends Thread {
                     SSLContext sslContext = getSSLContext(candidateCertificates, disableHttpsCertValidation);
                     if (sslContext != null) {
                         SslEngineConfigurator sslEngineConfigurator = new SslEngineConfigurator(sslContext);
-                        if (disableHttpsCertValidation) {
-                            sslEngineConfigurator.setHostnameVerifier(new NoCheckHostnameVerifier());
+                        if (hostnameVerifier != null) {
+                            sslEngineConfigurator.setHostnameVerifier(hostnameVerifier);
                         }
                         client.getProperties().put(ClientProperties.SSL_ENGINE_CONFIGURATOR, sslEngineConfigurator);
                     }
@@ -928,8 +937,8 @@ public class Engine extends Thread {
             } catch (Exception e) {
                 events.error(e);
             }
-            resolver = new JnlpAgentEndpointResolver(jenkinsUrls, credentials, proxyCredentials, tunnel,
-                    sslSocketFactory, disableHttpsCertValidation, agentName);
+            resolver = new JnlpAgentEndpointResolver(jenkinsUrls, agentName, credentials, proxyCredentials, tunnel,
+                    sslSocketFactory, disableHttpsCertValidation);
         } else {
             resolver = new JnlpAgentEndpointConfigurator(directConnection, instanceIdentity, protocols, proxyCredentials);
         }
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -187,6 +188,9 @@ public Thread newThread(@NonNull final Runnable r) {
 
     private boolean disableHttpsCertValidation = false;
 
+    @CheckForNull
+    private HostnameVerifier hostnameVerifier;
```
basil, line 192: Following a general paradigm where everytime a boolean field exists to disable validation, a `HostNameVerifier` field exists beneath it. Following this convention consistently in all classes that do this allows for certain common utilities like `JnlpAgentEndpointResolver#openURLConnection` to be simplified.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -928,8 +937,8 @@ private JnlpEndpointResolver createEndpointResolver(List<String> jenkinsUrls, St
             } catch (Exception e) {
                 events.error(e);
             }
-            resolver = new JnlpAgentEndpointResolver(jenkinsUrls, credentials, proxyCredentials, tunnel,
-                    sslSocketFactory, disableHttpsCertValidation, agentName);
+            resolver = new JnlpAgentEndpointResolver(jenkinsUrls, agentName, credentials, proxyCredentials, tunnel,
```
basil, line 940: Following a general paradigm where user-specified flags come before computed objects and mandatory arguments come before optional arguments. Using a consistent order helps improve readability as you go from one file to the next, which otherwise used different ordering which was confusing.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -404,26 +431,18 @@ public void run() throws CmdLineException, IOException, InterruptedException {
...
-                bootstrapInboundAgent();
-            }
-            runAsInboundAgent();
-        } else
-        if(tcpPortFile!=null) {
-            runAsTcpServer();
-        } else {
-            runWithStdinStdout();
+        if (noCertificateCheck) {
+            hostnameVerifier = new NoCheckHostnameVerifier();
```
basil, line 442: Initializing this eagerly rather than lazily for consistency with other code. Of course, this necessitates that all public methods in this file now call initialization, which is why I factored out initialization into its own method.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -624,6 +643,7 @@ private void runAsInboundAgent() throws CmdLineException, IOException, Interrupt
      */
     @SuppressFBWarnings(value = {"CIPHER_INTEGRITY", "STATIC_IV"}, justification = "Integrity not needed here. IV used for decryption only, loaded from encryptor.")
     public List<String> parseJnlpArguments() throws ParserConfigurationException, SAXException, IOException, InterruptedException {
+        initialize();  // For Swarm, or anyone else who invokes this public method directly rather than going through main() or run()
```
basil, line 646: With lazy initialization of the `noCertificateCheck` functionality, this was working, but only by accident. Now, we explicitly do all initialization in one place, and from every entrypoint.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -633,7 +653,7 @@ public List<String> parseJnlpArguments() throws ParserConfigurationException, SA
         while (true) {
             URLConnection con = null;
             try {
-                con = JnlpAgentEndpointResolver.openURLConnection(agentJnlpURL, agentJnlpCredentials, proxyCredentials, sslSocketFactory, noCertificateCheck, null);
+                con = JnlpAgentEndpointResolver.openURLConnection(agentJnlpURL, null, agentJnlpCredentials, proxyCredentials, sslSocketFactory, hostnameVerifier);
```
basil, line 656: As elsewhere, putting mandatory user-specified arguments before optional ones, and putting user-specified arguments before computed things like SSL socket factories and hostname verifiers.

src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java
```diff
@@ -85,17 +86,19 @@ public class JnlpAgentEndpointResolver extends JnlpEndpointResolver {
...
     private String proxyCredentials;
 
     private String tunnel;
 
+    private SSLSocketFactory sslSocketFactory;
+
     private boolean disableHttpsCertValidation;
 
-    private final String agentName;
+    private HostnameVerifier hostnameVerifier;
```
basil, line 101: As elsewhere, putting mandatory user-specified arguments before optional ones, and putting user-specified arguments before computed things like SSL socket factories and hostname verifiers.

jglick:
`internal` seems inappropriate if this is needed for the Swarm plugin and thus merits a release; `enhancement` perhaps?

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -390,6 +396,29 @@ public void run() throws CmdLineException, IOException, InterruptedException {
             return;
         }
 
+        if (connectionTarget != null) {
+            initialize();
+            runAsTcpClient();
+        } else if (agentJnlpURL != null || !urls.isEmpty() || directConnection != null) {
```
jglick, line 402: Fortunately this now has test coverage! (Though https://github.com/jenkinsci/bom/pull/2417 is of no help since the relevant tests require special setup to run.)

jglick (approved):
Looks right.

PR #705, opened 2023-12-07
Print deprecation warning when using `-jnlpUrl`

basil:
Now that https://github.com/jenkinsci/jenkins/pull/8639 and https://github.com/jenkinsci/jenkins-test-harness/pull/670 have been shipping for a bit, time to deprecate this legacy usage mode.

### Testing done

Verified that I could connect to the controller with the new calling convention without any warnings and that the warning was printed when using the old calling convention.


NotMyFault (approved):

timja (approved):
I suspect won't be seen by many, inbound agents logs probably aren't seen very often but worth a go.

basil:
> I suspect won't be seen by many, inbound agents logs probably aren't seen very often

https://github.com/jenkinsci/jenkins/pull/8773 prints a warning on the server side as well to increase the likelihood that this message is seen.

PR #709, opened 2023-12-26
JENKINS-65873: reduce threads to download jars and fix promise value …

wangyong001:
Actually two issues here:
1. issue one: unable to create new native thread:The remoting class loader execute findClass much quicker than download thread, if there are too many threads run concurrently to download jars, it has the possibility to cause "unable to created new native thread" mentioned here: [JENKINS-65873](https://issues.jenkins.io/browse/JENKINS-65873).  Also if the JVM parameter of Xms is set too small, it will easy to have this issue.
2. if the inprogressDownload is null, it will not equal the promise passed in, then the promise won't have chance to set any value (normal url or exception), thus when caller call promise.get, it will hangs there forever, which means Jenkins agent will hangs there

### Testing done
It is hard to do automation testing since it is a multiple thread issue. We have around 150k Jenkins builds every day and there about 6~7 out of 10k builds with such kind of issue (unable to create new native thread).
After use this fix, the issue is gone.

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #713, opened 2024-01-05
Use `ConcurrentHashMap#computeIfAbsent` in `JarCacheSupport`

basil:
While reading this code recently, I found some complicated logic to deal with a risk of multiple downloads of the same JAR getting scheduled, leading to special code to handle this race condition whose correctness is being challenged in https://github.com/jenkinsci/remoting/pull/709. This complicated logic is not necessary if one instead uses `ConcurrentHashMap#computeIfAbsent`, whose [Javadoc states](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent(K,java.util.function.Function)):

> The entire method invocation is performed atomically, so the function is applied at most once per key.

Due to these exactly once semantics, we don't have to worry about a race condition, so we can simplify the logic considerably, deleting the logic that is being challenged in the other PR rather than attempting to fix it. We also don't need the retry logic to deal with the executor rejecting the queue submission because (as an existing comment already notes) `AtmostOneThreadExecutor` never rejects submissions anyway (and any feasible replacement we might choose in the future would likely block rather than rejecting submissions).

### Testing done

Ran `mvn clean verify` as well as running a job on an inbound agent with the Git plugin to load a lot of classes. Verified that the JAR cache was correctly populated after being purged.

### DIFF
--- a/src/main/java/hudson/remoting/JarCacheSupport.java
+++ b/src/main/java/hudson/remoting/JarCacheSupport.java
@@ -9,7 +9,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
-import org.jenkinsci.remoting.util.ExecutorServiceUtils;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  * Default partial implementation of {@link JarCache}.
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

timja (approved):
looks much simpler

basil:
> and any feasible replacement we might choose in the future would likely block rather than rejecting submissions

Foreshadowing the use of `LinkedBlockingQueue` in https://github.com/jenkinsci/remoting/pull/717.

PR #716, opened 2024-01-08
Harden `JarCacheSupport`

basil:
An implicit invariant in `JarCacheSupport` is that `DownloadRunnable` is only ever processing the promise that is currently recorded in the `inprogress` map, an invariant which can be proved by reading the code and studying the cases that arise. This PR hardens the code by enforcing this invariant programmatically and throwing an `IllegalStateException` if the invariant is violated, which would only be the case if a future PR introduced a programming error.

### Testing done

I ran a job on an inbound agent to check out a repository with JGit and verified that the new assertion was never thrown (as expected).


MarkEWaite (approved):
Thanks!  Looks good to me.

PR #741, opened 2024-05-14
Pass through the remote `ClassLoader`’s name

jtnord:
### Testing done

Created a version of Jenkins with a local build of this, attached a debugger to an `ssh-slaves` agent and put a breakpoint on some code that runs in the agent (from the git-client plugin) to observe classloading.

See https://github.com/jenkinsci/jenkins/pull/9277 

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #742, opened 2024-05-15
set since to released version

jtnord:
sets the since version https://github.com/jenkinsci/remoting/pull/741/files#r1601332693 to the actual release version.

There is no need for this as the version in the code is higher than the previous release without this method and lower than the version with this method.   But it's wrong and it makes me feel dirty :) 

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/remoting/RemoteClassLoader.java
+++ b/src/main/java/hudson/remoting/RemoteClassLoader.java
@@ -898,7 +898,7 @@ final class RemoteClassLoader extends URLClassLoader {
 
         /**
          * Name of the classLoader
-         * @since 3242
+         * @since 3244.vf7f977e04755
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/remoting/RemoteClassLoader.java
```diff
@@ -898,7 +898,7 @@ public interface IClassLoader {
 
         /**
          * Name of the classLoader
-         * @since 3242
+         * @since 3244.vf7f977e04755
```
jtnord, line 901: https://github.com/jenkinsci/remoting/releases/tag/3244.vf7f977e04755

NotMyFault (approved):

PR #750, opened 2024-06-18
Organize imports

basil:
I found this old branch lying around and figured I should get it merged before it got more stale. This repository is quite old and has a very large number of inconsistent import conventions; sometimes multiple conventions are in use within the same file making the code a bit difficult to read. This PR normalizes imports according to the convention we have been using recently in core and Stapler.

### Testing done

`mvn clean verify`


MarkEWaite (approved):
Thanks!

PR #771, opened 2024-10-24
Fix websocket connection retries logic

Vlatombe:
We have observed cases where the websocket connection would fail temporarily with a handshake error that was not retried and causing the agent to fail.

Also, in CloudBees CI HA, there are some cases when an agent is disconnected on purpose and we expect a reconnect attempt as soon as possible. The default 10 seconds delay appeared too long in that case.

To reconcile this use case with the typical failure scenario, I have implemented exponential backoff for retries (immediate, 1 sec, 3 seconds, 7 seconds, 10 seconds).

### Testing done
* WS: Initial connection
* WS: Applying a rolling restart to the controller. The previous fatal handshake error is now retriable and prints a single log line

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -200,8 +200,6 @@ public class Engine extends Thread {
 
     private Duration noReconnectAfter;
 
-    private Instant firstAttempt;
-
     /**
      * Determines whether the socket will have {@link Socket#setKeepAlive(boolean)} set or not.
      *
@@ -785,12 +783,37 @@ public class Engine extends Thread {
                         client.getProperties().put(ClientProperties.SSL_ENGINE_CONFIGURATOR, sslEngineConfigurator);
                     }
                 }
-                container.connectToServer(
-                        new AgentEndpoint(),
-                        ClientEndpointConfig.Builder.create()
-                                .configurator(headerHandler)
-                                .build(),
-                        URI.create(wsUrl + "wsagents/"));
+                if (!succeedsWithRetries(() -> {
+                    // Unlike JnlpAgentEndpointResolver, we do not use $jenkins/tcpSlaveAgentListener/, as that will be
+                    // a 404 if the TCP port is disabled.
+                    URL ping = new URL(hudsonUrl, "login");
+                    try {
+                        HttpURLConnection conn = (HttpURLConnection) ping.openConnection();
+                        int status = conn.getResponseCode();
+                        conn.disconnect();
+                        if (status == 200) {
+                            return true;
+                        } else {
+                            events.status(ping + " is not ready: " + status);
+                        }
+                    } catch (IOException x) {
+                        events.status(ping + " is not ready", x);
+                    }
+                    return false;
+                })) {
+                    return;
+                }
+                if (!succeedsWithRetries(() -> {
+                    container.connectToServer(
+                            new AgentEndpoint(),
+                            ClientEndpointConfig.Builder.create()
+                                    .configurator(headerHandler)
+                                    .build(),
+                            URI.create(wsUrl + "wsagents/"));
+                    return true;
+                })) {
+                    return;
+                }
                 while (ch.get() == null) {
                     Thread.sleep(100);
                 }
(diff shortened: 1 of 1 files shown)


Vlatombe:
I'm also considering a bigger refactoring (to be filed in a separate PR) to reduce duplication across inbound tcp and websocket connection flows and make them more similar.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -801,38 +807,115 @@ public void closeRead() throws IOException {
...
+     */
+    private boolean succeedsWithRetries(SupplierThrowingException<Boolean> condition) throws InterruptedException {
+        var exponentialRetry = new ExponentialRetry(noReconnectAfter);
+        while (exponentialRetry != null) {
+            try {
+                if (condition.get()) {
+                    return true;
+                }
+            } catch (Exception x) {
+                events.status("Failed to connect: " + x.getMessage());
```
jglick, line 832: We are losing the stack trace here; is that potentially important for diagnosis?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -801,38 +807,115 @@ public void closeRead() throws IOException {
...
+    private static class ExponentialRetry {
+        final int factor;
+        final Instant beginning;
+        final Duration delay;
+        final Duration timeout;
+        final Duration incrementDelay;
+        final Duration maxDelay;
+
+        ExponentialRetry(Duration timeout) {
+            this(Duration.ofSeconds(0), timeout, 2, Duration.ofSeconds(1), Duration.ofSeconds(10));
```
jglick, line 870: compare user-configurable version in #676

src/main/java/hudson/remoting/Engine.java
```diff
@@ -801,38 +807,115 @@ public void closeRead() throws IOException {
...
+                events.status("Waiting " + delaySeconds + " seconds before retry");
+                TimeUnit.SECONDS.sleep(delaySeconds);
+            }
+            return next;
+        }
+    }
+
+    @FunctionalInterface
+    private interface SupplierThrowingException<T> {
+        T get() throws Exception;
```
jglick, line 916: Why not just `java.util.concurrent.Callable`?

src/main/java/hudson/remoting/Engine.java
```diff
@@ -801,38 +807,115 @@ public void closeRead() throws IOException {
...
+
+        ExponentialRetry next(EngineListenerSplitter events) throws InterruptedException {
+            var next = new ExponentialRetry(this);
+            if (next.timeoutExceeded()) {
+                events.status("Bailing out after " + DurationFormatter.format(next.timeout));
+                return null;
+            } else {
+                var delaySeconds = next.delay.toSeconds();
+                events.status("Waiting " + delaySeconds + " seconds before retry");
+                TimeUnit.SECONDS.sleep(delaySeconds);
```
jglick, line 908: can probably convert `next.delay` directly

jglick (approved):

src/main/java/hudson/remoting/Engine.java
Vlatombe: In my experience these stacktraces are noisy. I could eventually log them with standard JUL logger (FINE) in parallel, but I don't really see the point of keeping them in status.

src/main/java/hudson/remoting/Engine.java
Vlatombe: True, I often forget about built-in types

src/main/java/hudson/remoting/Engine.java
Vlatombe: jitter could be useful, but exposing these as user-level settings seem overkill to me.

jglick: Agreed, just noting for reference.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -903,19 +903,13 @@ ExponentialRetry next(EngineListenerSplitter events) throws InterruptedException
                 events.status("Bailing out after " + DurationFormatter.format(next.timeout));
                 return null;
             } else {
-                var delaySeconds = next.delay.toSeconds();
-                events.status("Waiting " + delaySeconds + " seconds before retry");
-                TimeUnit.SECONDS.sleep(delaySeconds);
+                events.status("Waiting " + DurationFormatter.format(next.delay) + " before retry");
+                TimeUnit.SECONDS.sleep(next.delay.toSeconds());
```
jglick, line 907: Sorry for being unclear, what I meant was actually
```suggestion
                Thread.sleep(next.delay);
```
but now I see that is Java 19+. In the meantime, for subsecond precision
```suggestion
                Thread.sleep(next.delay.toMillis());
```

jglick (approved):

kitos9112:
Hi,

I should file a new issue for this but we think this commit has broken our inbound-connected Jenkins agents. 

We believe that the method below does not play well with authenticated Jenkins instances in which the `/login` endpoint would reply with 401 responses automatically. We use SSO Kerberos and LDAP plugins for authentication. Is there a way to bypass this or perhaps modify the ping URL?

```
    private Boolean pingSuccessful() throws MalformedURLException {
        // Unlike JnlpAgentEndpointResolver, we do not use $jenkins/tcpSlaveAgentListener/, as that will be
        // a 404 if the TCP port is disabled.
        URL ping = new URL(hudsonUrl, "login");
        try {
            HttpURLConnection conn = (HttpURLConnection) ping.openConnection();
            int status = conn.getResponseCode();
            conn.disconnect();
            if (status == 200) {
                return true;
            } else {
                events.status(ping + " is not ready: " + status);
            }
        } catch (IOException x) {
            events.status(ping + " is not ready", x);
        }
        return false;
    }
```

jglick:
> authenticated Jenkins instances in which the `/login` endpoint would reply with 401 responses

It should reply with a 200. Not sure what plugins or reverse proxies you are using that would produce a 401.

(In general I would suggest `/whoAmI/api/json?tree=authenticated` over `/login`, or for new controllers, `/health`.)

kitos9112:
THanks @jglick - We use an Nginx reverse proxy with the documented configuration available on your documents. When it comes to authentication plugins, as stated above, we use the SSO Kerberos and LDAP plugins. In fact, the 401 response from Jenkins contains the HTTP header `www-Authenticate` set to Negotiate and the basic real for our internal IDM.

I do not think it is a general practice to perform a healthcheck against a login endpoint. Unless a 40{1,3} gets considered "healthy". Using `/health` or even exposed it as a CLI arg would definitely be the way forward, hardcoding it makes it extremely difficult to get non-standard Jenkins solutions working.

kitos9112:
I filed this bug report on your Jira appliance

https://issues.jenkins.io/browse/JENKINS-75881

MarkEWaite:
> I'm unsure if we'd get much traction there. In the meantime, would you be willing at making some changes in the agent itself as we've been discussing so far?

I think it would be a mistake to make a change in remoting when a change can happen in the plugin with the bug.  If the plugin maintainer does not respond promptly, you can propose to adopt the plugin @kitos9112 and make the fix yourself.

kitos9112:
> _“I think it would be a mistake to change remoting when the plugin with the bug could change.”_

@MarkEWaite 

The thing is — from the protocol’s point of view, the plugin isn’t buggy at all. It is working as designed following Internet standards, let me explain.

1. The agent (or a browser) sends **`GET /login`** with no credentials.  
2. Jenkins answers **`401 Unauthorized`** and a **`WWW‑Authenticate: Negotiate`** header.  
   * That is exactly what **RFC 4559** ([SPNEGO over HTTP](https://datatracker.ietf.org/doc/html/rfc4559)) and **RFC 9110 §15.5.2** mandate when the request has no Kerberos ticket. :contentReference[oaicite:0]{index=0}  
3. A Kerberos‑aware browser then repeats the request with a SPNEGO token, but a headless inbound agent never will—so it will _always_ see the 401.

Asking the Kerberos plugin to return 200 would break the spec and weaken security. It’s not a bug; it’s the protocol working as designed.

A health probe simply needs “tell me you’re alive”.
Jenkins already provides unauthenticated endpoints for that, as your colleagues discussed here:

| Endpoint | Since | Returns |
|----------|-------|---------|
| `/health` | 2.506 | `200`/`204` with minimal JSON |
| `/whoAmI/api/json` | ancient? | tiny JSON, always `200` |

Those paths never trigger the Kerberos challenge, so they work out of the box in secured environments. 

```
$ curl -v https://jenkins.internal.net/health/
< HTTP/2 200
< server: nginx/1.20.1
< date: Mon, 14 Jul 2025 19:46:52 GMT
< content-type: application/json;charset=utf-8
< content-length: 15
< vary: Accept-Encoding
< x-content-type-options: nosniff
                                                                                                                                                                                                                          
{"status":true}%
```

As @jglick suggested, the way forward should be:

1. Remoting first tries `/health` (or an operator‑configurable URL) and treats **200/204** as healthy.  
2. If that’s missing (old controller), fall back to
 `/whoAmI/api/json`.  
3. Drop `/login` from the probe list; it was never designed for this use.

timja:
Won't you have the same problem with `/health` or `/whoAmI/api/json?tree=authenticated`?

FWIW changing from `/login` to `/health` with a fallback should be fine although ideally not released for a few weeks so that at least the current LTS (released 23/07) will hit the default endpoint.

kitos9112:
@timja 

Nope. Those endpoints are not protected by the Kerberos SSO plugin at all. This are simple curl requests getting a 200 response

```
❯ curl https://jenkins.internal.net/health/                                                                                                                                                                                                                                    
{"status":true}%
❯ curl https://jenkins.internal.net/whoAmI/api/json/                                                                                                                                                                                                                     
{"_class":"hudson.security.WhoAmI","anonymous":true,"authenticated":true,"authorities":["anonymous"],"name":"anonymous"}%
```

We're running the rolling releases in our lower environments (CI) so it'd not be a big issue... even using a newer agent JAR release.

PR #777, opened 2024-12-03
Simplify Util#makeResource

daniel-beck:
This simplifies the code of `Util#makeResource` by eliminating the creation of a directory hierarchy.

The previous FindBugs suppression explanation was incorrect. Technically this method did have a path traversal issue, but it's not `public` code and the callers don't invoke it in an unsafe manner. The new code is simpler and easier to reason about.

### Testing done

None

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/remoting/Util.java
+++ b/src/main/java/hudson/remoting/Util.java
@@ -55,13 +55,10 @@ public class Util {
     }
 
     @NonNull
-    @SuppressFBWarnings(
-            value = "PATH_TRAVERSAL_IN",
-            justification = "This path exists within a temp directory so the potential traversal is limited.")
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "Sanitized by #getBaseName")
     static File makeResource(String name, byte[] image) throws IOException {
         Path tmpDir = Files.createTempDirectory("resource-");
-        File resource = new File(tmpDir.toFile(), name);
-        Files.createDirectories(PathUtils.fileToPath(resource.getParentFile()));
+        File resource = new File(tmpDir.toFile(), getBaseName(name.replace('\\', '/')));
         Files.createFile(PathUtils.fileToPath(resource));
 
         try (FileOutputStream fos = new FileOutputStream(resource)) {


timja (approved):

src/main/java/hudson/remoting/Util.java
```diff
@@ -55,13 +55,10 @@ public static void copy(InputStream in, OutputStream out) throws IOException {
...
     @NonNull
-    @SuppressFBWarnings(
-            value = "PATH_TRAVERSAL_IN",
-            justification = "This path exists within a temp directory so the potential traversal is limited.")
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "Sanitized by #getBaseName")
     static File makeResource(String name, byte[] image) throws IOException {
         Path tmpDir = Files.createTempDirectory("resource-");
-        File resource = new File(tmpDir.toFile(), name);
-        Files.createDirectories(PathUtils.fileToPath(resource.getParentFile()));
+        File resource = new File(tmpDir.toFile(), getBaseName(name.replace('\\', '/')));
```
jtnord, line 61: I see this quite a lot presumably to attempt to normalize windows paths.
but Linux paths can (even though it is really really not nornal) have a `\` as part of the path.
`EXPORT TMP=/tmp/this_is_a_directory_with_a_\\_ok` 

Happens when the home directory uses a username that comes from samba and using multiple domains and you need to distinguish `DOMAIN1\user` and `DOMAIN2\user` and the full domainname e.g. `DOMAIN\user` 

```suggestion
        File resource = new File(tmpDir.toFile(), getBaseName(name.replace(File.separatorChar, '/')));
```

I don't think you can have a `/` inside a directory on windows

daniel-beck: So this breaks if someone passes in a filename whose last character is `\`. Unsure whether such callers exist.

Otherwise this will just use a substring of the filename. Since there's one directory per file, there's no risk of collisions, making diagnostics a bit more difficult.

Given [existing use of this idiom in the codebase](https://github.com/jenkinsci/remoting/blob/b79d2d09f803bd0171c2ce18f72d3156cdc9138b/src/main/java/hudson/remoting/RemoteClassLoader.java#L696-L721) I'm tempted to leave this as is. WDYT?

PR #788, opened 2025-03-13
Remove unnecessary SpotBugs suppressions

basil:
- Migrates all High and Medium SpotBugs suppressions from XML to Java annotations, where they are easier to manage
- Leaves all Low suppressions in XML
- Removes all unnecessary suppressions

The use of Low rather than Medium dates back to commit 39bc5430d7966e7ab184cb99044ed7510a2adb6b by Kohsuke:

> Couldn't figure out how to add specific checker to the default setting, so I'm activating a whole bunch by setting `threshold=low`, then excluding most of them except the serialization problem

Since these Low threshold violations are frequent and annoying, leaving their suppressions in XML so as not to litter the codebase with unnecessary annotations. Perhaps in the future some better solution can be found to Kohsuke's original problem.

### Testing done

`mvn clean verify`

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/remoting/ClassFilter.java
+++ b/src/main/java/hudson/remoting/ClassFilter.java
@@ -128,6 +128,7 @@ public abstract class ClassFilter {
         }
     };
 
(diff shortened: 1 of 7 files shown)


MarkEWaite (approved):
Thanks!

PR #792, opened 2025-03-18
Reduce calls to deprecated Java Platform methods

basil:
These methods are deprecated in the Java Platform, so move to the recommended non-deprecated equivalents.

### Testing done

`mvn clean verify`


basil:
Passed core and BOM test suites, so let's roll

PR #793, opened 2025-03-24
`-lockWorkDir`

jglick:
([CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-57023))

Tested on both Linux and Windows 10.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -44,8 +44,12 @@ import java.net.Socket;
 import java.net.URI;
 import java.net.URL;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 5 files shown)


jglick:
Probably so, and I think `-lockWorkDir` ought to be included in https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/java/jenkins/slaves/RemotingWorkDirSettings.java#L133 for the suggest command line for a permanent agent. (If you are seeing that page, it is presumably not a cloud agent.) And included in some integration tests in core using a forked `java -jar agent.jar`.

Keeping in draft for now pending feedback from a CloudBees CI customer who reported observing multiple processes for a given agent.

PR #805, opened 2025-06-24
Avoid `System.exit` from `Launcher.run`

jglick:
Avoiding `System.exit` when used as an inbound agent (`main(String[])`). The `main` overloads called by `runWithStdinStdout` (or deprecated `runAsTcpServer`) used for outbound agents https://github.com/jenkinsci/remoting/blob/1d80543ebaf27c64e3641d6a67993a13ca9cd7f6/src/main/java/hudson/remoting/Launcher.java#L991-L1065 still `exit`.

Also ensuring that some resources are closed when exiting the (again, inbound) agent loop which were previously leaking threads.

In combination, these fixes make it possible to run multiple (inbound) agents (`Engine`s & `Channel`s) from one JVM, with independent configurations and even class loading after `remoting.jar` itself. Tested in the context of CloudBees CI in high availability mode, where it is unsafe to offer multiple executors from a single agent. Without the fixes, running multiple (inbound) agents from one JVM mostly works but there are some poor effects from disconnecting some agents while others are still trying to run.

### DIFF
--- a/src/main/java/hudson/remoting/Engine.java
+++ b/src/main/java/hudson/remoting/Engine.java
@@ -128,21 +128,7 @@ public class Engine extends Thread {
     /**
      * Thread pool that sets {@link #CURRENT}.
      */
-    private final ExecutorService executor = Executors.newCachedThreadPool(new ThreadFactory() {
-        private final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
-
-        @Override
-        public Thread newThread(@NonNull final Runnable r) {
-            Thread thread = defaultFactory.newThread(() -> {
-                CURRENT.set(Engine.this);
-                r.run();
-            });
-            thread.setDaemon(true);
-            thread.setUncaughtExceptionHandler(
-                    (t, e) -> LOGGER.log(Level.SEVERE, e, () -> "Uncaught exception in thread " + t));
-            return thread;
-        }
-    });
+    private final ExecutorService executor;
 
     /**
      * @deprecated
@@ -276,6 +262,7 @@ public class Engine extends Thread {
             String directConnection,
             String instanceIdentity,
             Set<String> protocols) {
+        executor = makeExecutor(this, agentName);
         this.listener = listener;
         this.directConnection = directConnection;
         if (listener != null) {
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/remoting/Engine.java
```diff
@@ -750,68 +774,73 @@ public void closeRead() throws IOException {
...
+                    ch.get().join();
+                    events.status("Terminated");
+                    if (noReconnect) {
+                        return;
+                    }
+                    events.onDisconnect();
+                    reconnect();
+                } finally {
+                    if (container instanceof BaseContainer baseContainer) {
+                        baseContainer.shutdown();
```
jglick, line 841: Unfortunately this only shuts down the main thread pool. Tyrus seems to maintain another thread pool in `org.glassfish.tyrus.container.jdk.client.TransportFilter.channelGroup` which I cannot figure out how to close; even tracking down which fields hold on to this object is challenging in such a complex library. This pool seems to have been overlooked by https://github.com/eclipse-ee4j/tyrus/pull/927.

src/main/java/hudson/remoting/Engine.java
```diff
@@ -308,6 +297,25 @@ private static URL ensureTrailingSlash(URL u) {
         }
     }
 
+    private static ExecutorService makeExecutor(Engine engine, String agentName) {
+        return Executors.newCachedThreadPool(new ThreadFactory() {
+            private final ThreadFactory defaultFactory =
+                    new NamingThreadFactory(Executors.defaultThreadFactory(), agentName);
```
jglick, line 303: Extracted to a method so as to be able to pass `agentName` here.

src/main/java/hudson/remoting/Launcher.java
```diff
@@ -723,11 +724,16 @@ private static void validateInboundAgentArgs(Launcher launcher) throws CmdLineEx
...
-        Engine engine = createEngine();
+        SettableFuture<Void> completion = SettableFuture.create();
+        Engine engine = createEngine(completion);
         engine.startEngine();
         try {
-            engine.join();
+            completion.get();
             LOGGER.fine("Engine has died");
+        } catch (InterruptedException x) {
+            throw x;
```
Vlatombe, line 734: Doesn't seem to be needed?
```suggestion
```

Vlatombe (approved):

PR #811, opened 2025-07-21
[JENKINS-67829] compute the checksum per JVM

### LINKED ISSUE #1459: [JENKINS-67829] JarLoaderImpl is per connection in a controller -causing repeated checking of jars checksums
<p>When using ephemeral agents it was noticed even after Jenkins had been running for a day that there where still threads calculating the digest for jars with stack traces like the following in the controller</p>

<pre>
"Computer.threadPoolForRemoting [#17077] for JNLP4-connect connection from x.x.x.x/x.x.x.x:49778 id=1547" Id=164040 Group=main RUNNABLE
	at sun.security.provider.DigestBase.implCompressMultiBlock0(DigestBase.java:147)
	at sun.security.provider.DigestBase.implCompressMultiBlock(DigestBase.java:142)
	at sun.security.provider.DigestBase.engineUpdate(DigestBase.java:129)
	at java.security.MessageDigest$Delegate.engineUpdate(MessageDigest.java:601)
	at java.security.MessageDigest.update(MessageDigest.java:328)
	at java.security.DigestOutputStream.write(DigestOutputStream.java:147)
	at hudson.remoting.Util.copy(Util.java:58)
	at hudson.remoting.Checksum.forURL(Checksum.java:83)
	at hudson.remoting.JarLoaderImpl.calcChecksum(JarLoaderImpl.java:92)
	at hudson.remoting.JarLoaderImpl.calcChecksum(JarLoaderImpl.java:64)
	at hudson.remoting.RemoteClassLoader$ClassLoaderProxy.fetch4(RemoteClassLoader.java:1014)
	at hudson.remoting.RemoteClassLoader$ClassLoaderProxy.fetch3(RemoteClassLoader.java:1055)
</pre>

<p>it is highly unlikely that any jar that Jenkins knew about had not been already requested by at least one agent before.</p>

<p>calcChecksum has a cache <a href="https://github.com/jenkinsci/remoting/blob/efb2c4fbe2e3a848f39f96cb99408ea2350a3aa2/src/main/java/hudson/remoting/JarLoaderImpl.java#L88-L97" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/jenkinsci/remoting/blob/efb2c4fbe2e3a848f39f96cb99408ea2350a3aa2/src/main/java/hudson/remoting/JarLoaderImpl.java#L88-L97</a></p>

<p>however every remoting channel has its own loader - so in effect the cache is per agent not per Jenkins.  As class loading can only work one way (an agent can load classes from a controller) the cache could be shared.  (Jenkins can not change the jars at runtime without a restart, and all agents can see exactly the same jars as all other agents)</p>

<p>if you have lots of agents connecting then this imposes a scaling limit on Jenkins as it involves both IO to load the file and CPU to compute the digest.</p>

---
<details><summary><i>Originally reported by  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=teilo">teilo</a>, imported from: <a class="no-jira-link-rewrite" href="https://issues.jenkins.io/browse/JENKINS-67829" target="_blank">JarLoaderImpl is per connection in a controller -causing repeated checking of jars checksums</a></i></summary>
<i><ul>
<li><b>assignee</b>:  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=teilo">teilo</a>
<li><b>status</b>: In Progress
<li><b>priority</b>: Major
<li><b>component(s)</b>: remoting
<li><b>label(s)</b>: performance
<li><b>resolution</b>: Unresolved
<li><b>votes</b>: 0
<li><b>watchers</b>: 3
<li><b>imported</b>: 2025-11-25
</ul></i>
<details><summary>Raw content of original issue</summary>

<pre>
<p>When using ephemeral agents it was noticed even after Jenkins had been running for a day that there where still threads calculating the digest for jars with stack traces like the following in the controller</p>

<div class="code panel" style="border-width: 1px;"><div class="codeContent panelContent">
<pre class="code-none">"Computer.threadPoolForRemoting [#17077] for JNLP4-connect connection from x.x.x.x/x.x.x.x:49778 id=1547" Id=164040 Group=main RUNNABLE
	at sun.security.provider.DigestBase.implCompressMultiBlock0(DigestBase.java:147)
	at sun.security.provider.DigestBase.implCompressMultiBlock(DigestBase.java:142)
	at sun.security.provider.DigestBase.engineUpdate(DigestBase.java:129)
	at java.security.MessageDigest$Delegate.engineUpdate(MessageDigest.java:601)
	at java.security.MessageDigest.update(MessageDigest.java:328)
	at java.security.DigestOutputStream.write(DigestOutputStream.java:147)
	at hudson.remoting.Util.copy(Util.java:58)
	at hudson.remoting.Checksum.forURL(Checksum.java:83)
	at hudson.remoting.JarLoaderImpl.calcChecksum(JarLoaderImpl.java:92)
	at hudson.remoting.JarLoaderImpl.calcChecksum(JarLoaderImpl.java:64)
	at hudson.remoting.RemoteClassLoader$ClassLoaderProxy.fetch4(RemoteClassLoader.java:1014)
	at hudson.remoting.RemoteClassLoader$ClassLoaderProxy.fetch3(RemoteClassLoader.java:1055)
</pre>
</div></div>

<p>it is highly unlikely that any jar that Jenkins knew about had not been already requested by at least one agent before.</p>

<p>calcChecksum has a cache <a href="https://github.com/jenkinsci/remoting/blob/efb2c4fbe2e3a848f39f96cb99408ea2350a3aa2/src/main/java/hudson/remoting/JarLoaderImpl.java#L88-L97" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/jenkinsci/remoting/blob/efb2c4fbe2e3a848f39f96cb99408ea2350a3aa2/src/main/java/hudson/remoting/JarLoaderImpl.java#L88-L97</a></p>

<p>however every remoting channel has its own loader - so in effect the cache is per agent not per Jenkins.  As class loading can only work one way (an agent can load classes from a controller) the cache could be shared.  (Jenkins can not change the jars at runtime without a restart, and all agents can see exactly the same jars as all other agents)</p>

<p>if you have lots of agents connecting then this imposes a scaling limit on Jenkins as it involves both IO to load the file and CPU to compute the digest.</p></pre>
</details>
</details>

jtnord:
fixes #1459 (JENKINS-67829)

As we do not support dynamic class unloading the mappting between a `URL` and a `Checksum` can never change for the lifetime of the JVM.  That is in Jenkins we can only install new plugins, upgrades **always** require a restart.

To that end the `Checksum` cache in `JarLoaderImpl` always is the checksum of the file on the local side so the only way it can change is if the JVM is restarted.
Therefor we do not need to calculate it for every Channel, rather we can create it once for all Channels on a given JVM.

### Testing done

* functionally run an inbound (java -jar agent.jar) against a Jenkins that has this patch from an agent that does not
* functionally run an inbound (java -jar agent.jar) against a Jenkins that has this patch from an agent that also has this patch

Both of the above had some extra logging in [`hudson.remoting.Checksum.forUrl(URL)`](https://github.com/jenkinsci/remoting/blob/32e37a923332cee84c9ac8e8896ca21096cff6fd/src/main/java/hudson/remoting/Checksum.java#L81-L92) to show when a checksum was created.  Observed that without the each new agent / reconnection of the agent the method is called and with the path the method is called once per URL for the first connection since startup.

* ATH https://github.com/jenkinsci/acceptance-test-harness/pull/2103

ran an artificial test that consisted of one pipeline spawining N k8s agents in a loop (Y) both with and without some steps (sh,echo,sleep) (for N,Y = 5,10 & 15,20) both with this patch and without and with and without a restart.
The test system limited the CPU of the controller to 1 vCPU and no limit on disk access (local NVME drive).
Timings where marginally faster for the pipeline with this change, but CPU controller spikes on the controller lasted significanly less time.
Running the stress (N,Y = 15,20) after a restart (so the cache would be clean) showed a slight increase in the time to complete the pipeline (from 4min 49s before this patch to 5min 6s after this patch).  Inspecting the CPU stats for the controller showed that it was using less CPU throughout the test so the controller had CPU available for other tasks.
In all cases the testing was not fully scientific.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -25,49 +26,46 @@ class JarLoaderImpl implements JarLoader, SerializableOnlyOverRemoting {
...
 
-    @SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "TODO needs triage")
-    private final ConcurrentMap<Checksum, URL> knownJars = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<Checksum, URI> KNOWN_JARS = new ConcurrentHashMap<>();
 
-    @SuppressFBWarnings(
-            value = {"DMI_COLLECTION_OF_URLS", "SE_BAD_FIELD"},
-            justification = "TODO needs triage")
-    private final ConcurrentMap<URL, Checksum> checksums = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<URI, CompletableFuture<Checksum>> CHECKSUMS = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
jtnord: we do not support class unloading so this is acceptable (and the whole reason behind this PR)

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -90,19 +88,47 @@ public void notifyJarPresence(long[] sums) {
...
+        if (existing != null) {
+            // another thread already has a slot — wait for their result without locking
+            slot = existing;
+        } else {
+            // we won the race, compute the checksum then update the future
+            try {
+                Checksum c = Checksum.forURL(jar.toURL());
+                KNOWN_JARS.put(c, jar);
+                slot.complete(c);
+            } catch (IOException | Error | RuntimeException e) {
```
jtnord, line 108: `Error` and `RuntimeException` handle the non checked exceptions. IOException handles the only checked exception that is thrown.  If the code throws anything else this should be a compile error.

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -90,19 +88,47 @@ public void notifyJarPresence(long[] sums) {
...
 
     /**
-     * Obtains the checksum for the jar at the specified URL.
+     * Obtains the checksum for the jar at the specified URI.
      */
-    public Checksum calcChecksum(URL jar) throws IOException {
-        Checksum v = checksums.get(jar); // cache hit
-        if (v != null) {
-            return v;
+    public Checksum calcChecksum(URI jar) throws IOException {
```
jtnord, line 93: whilst this is public is it not part of the contract of `JarLoader` and so is not exported across the channel.

timja (approved):
Looks sensible

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -90,19 +88,47 @@ public void notifyJarPresence(long[] sums) {
...
+        // Use Future-based memoizer so that internal map locks (which can be shared for different keys)
+        // are never held during I/O / digest computation.
+
+        CompletableFuture<Checksum> slot = new CompletableFuture<>();
+        CompletableFuture<Checksum> existing = CHECKSUMS.putIfAbsent(jar, slot);
```
gbhat618, line 98: this is where the race is prevented

jtnord: correct.  and we are not using `computeIfAbsent` to compute the checksum as that uses IO and can lock the map for an unnecessarily long time.

gbhat618: right, noticed caching of the future instead of computing; neat.

gbhat618 (approved):

src/main/java/hudson/remoting/JarLoaderImpl.java
```diff
@@ -25,49 +26,46 @@ class JarLoaderImpl implements JarLoader, SerializableOnlyOverRemoting {
...
 
-    @SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "TODO needs triage")
-    private final ConcurrentMap<Checksum, URL> knownJars = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<Checksum, URI> KNOWN_JARS = new ConcurrentHashMap<>();
 
-    @SuppressFBWarnings(
-            value = {"DMI_COLLECTION_OF_URLS", "SE_BAD_FIELD"},
-            justification = "TODO needs triage")
-    private final ConcurrentMap<URL, Checksum> checksums = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<URI, CompletableFuture<Checksum>> CHECKSUMS = new ConcurrentHashMap<>();
```
gbhat618, line 31: just out of curiosity: why changing from URL to URI ?

jtnord: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-maps-and-sets-of-urls-can-be-performance-hogs-dmi-collection-of-urls

which in our case is moot as all the URLs *should* be `file:` so no DNS should occur, but still just switching and using a micro benchmark (JMH) showed that `URI` was significantly faster than `URL`

jglick:
FTR delivered in 2.571 via https://github.com/jenkinsci/jenkins/pull/26983.

PR #1544, opened 2026-08-31
[JENKINS-45023] Refuse all Request types once a channel close is requested

om7057:
This addresses the two long-disabled pending-close tests in `ChannelTest` (JENKINS-45023 / JENKINS-45294).

### Problem

`Channel.close()` sets `closeRequested` (and `closeRequestCause`) up front, then acquires the `Channel` monitor for the rest of the teardown. During that window the channel is "closing" but `outClosed` is still `null`.

The guard that refuses new requests during this window, `checkIfCanBeExecutedOnChannel()`, was only overridden on `UserRequest` and `RemoteInvocationHandler.UserRPCRequest` (identical copy-pasted bodies). The base `Request.checkIfCanBeExecutedOnChannel()` only checks `getSenderCloseCause()` (i.e. `outClosed`).

So a base `RPCRequest` (RPC on an internal proxy), a `ProxyInputStream.Chunk`, or any other non-user `Request` sent in that window passes the check and then blocks on the `Channel` monitor held by `close()` / `terminate()`, instead of failing fast with `ChannelClosedException`.

### Change

- Move the `isClosingOrClosed()` check into the base `Request.checkIfCanBeExecutedOnChannel()`; remove the two now-redundant overrides.
- Rework the `ChannelCloseLock` test helper to reach the pending-close state deterministically: a latch fires once the background task holds the `Channel` monitor, then it polls for `closeRequested`, instead of `Thread.sleep(1000)` plus thread-pool scheduling. That sleep-based setup is why both tests were `@Disabled("TODO flake timed out after 15 seconds")`.
- Re-enable `testShouldNotAcceptUserRequestsWhenIsBeingClosed` and `testShouldNotAcceptUserRPCRequestsWhenIsBeingClosed`. The latter now also exercises an internal (non user-space) proxy, which is the path this change fixes.

`UserRPCRequest` is kept as a (now trivial) subclass to avoid a wire-compatibility change.

### Testing done

- Both re-enabled tests: green across many consecutive local runs, on all channel runners.
- Revert check: with the `Request.java` change reverted, `testShouldNotAcceptUserRPCRequestsWhenIsBeingClosed` times out after 15s on every channel runner (6/6), which is the original disabled symptom.
- Full suite: `mvn test` green, 2119 run, 0 failures, 0 errors (12 more than before: the re-enabled parameterized tests).
- `mvn spotless:check` clean.

### Submitter checklist

- [x] There is automated testing or an explanation as to why this change has no tests.
- [x] New public classes, fields, and methods are annotated with `@Restricted` or have `@since TODO` Javadoc, as appropriate.
- [x] New deprecations are annotated with `@Deprecated(since = "TODO")` or `@Deprecated(forRemoval = true, since = "TODO")` if applicable.
- [x] For dependency updates: there are links to external changelogs and, if possible, full differentials.
- [x] For new APIs and extension points: there is a link to at least one consumer.

### DIFF
--- a/src/main/java/hudson/remoting/RemoteInvocationHandler.java
+++ b/src/main/java/hudson/remoting/RemoteInvocationHandler.java
@@ -1066,22 +1066,6 @@ final class RemoteInvocationHandler implements InvocationHandler, Serializable {
         UserRPCRequest(int oid, Method m, Object[] arguments, ClassLoader cl, boolean recordCreatedAt) {
             super(oid, m, arguments, cl, recordCreatedAt);
         }
-
-        // Same implementation as UserRequest
-        @Override
-        public void checkIfCanBeExecutedOnChannel(@NonNull Channel channel) throws IOException {
-            // Default check for all requests
-            super.checkIfCanBeExecutedOnChannel(channel);
-
-            // We also do not want to run UserRequests when the channel is being closed
-            if (channel.isClosingOrClosed()) {
-                throw new ChannelClosedException(
-                        channel,
-                        "The request cannot be executed on channel " + channel + ". "
-                                + "The channel is closing down or has closed down",
-                        channel.getCloseRequestCause());
-            }
-        }
     }
 
     private static final Object[] EMPTY_ARRAY = new Object[0];
--- a/src/main/java/hudson/remoting/Request.java
+++ b/src/main/java/hudson/remoting/Request.java
@@ -130,6 +130,17 @@ public abstract class Request<RSP extends Serializable, EXC extends Throwable> e
             // Sender is closed, we won't be able to send anything
             throw new ChannelClosedException(channel, senderCloseCause);
         }
+
+        // JENKINS-45023: also refuse once a close has been requested but has not completed yet.
+        // Otherwise the request would proceed to acquire the Channel lock, which is held for the
+        // duration of close()/terminate(), and block there instead of failing fast.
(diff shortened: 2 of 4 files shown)


timja (approved):
Thanks!

om7057:
> Thanks!

Really appreciate the quick review and merge!
The Copilot comments were unnecessary in this case, right?

timja:
> > Thanks!
> 
> Really appreciate the quick review and merge! The Copilot comments were unnecessary in this case, right?

I think so, may want to make `closeRequested` volatile though

om7057:
Opened #1545. `closeRequested` and `closeRequestCause` are written outside the `this` monitor but read from arbitrary threads via `isClosingOrClosed()` / `getCloseRequestCause()`, so both are now `volatile`, matching `inClosed`/`outClosed`.

om7057:
> > > Thanks!
> > 
> > 
> > Really appreciate the quick review and merge! The Copilot comments were unnecessary in this case, right?
> 
> I think so, may want to make `closeRequested` volatile though

That works; I've opened: #1545 
`closeRequested` and `closeRequestCause` are written outside the monitor but read from other threads through `isClosingOrClosed()` and `getCloseRequestCause()`. Both are now marked `volatile`, consistent with the existing `inClosed` and `outClosed` fields.

Please go through it once.
Thank you!

